PR #45, opened 2026-05-04
refactor(skills): migrate config/user.md to .apache-steward-overrides/user.md

andreahlert:
## Summary

Migrates the 4 security skills from the legacy `config/user.md` path (pre-snapshot architecture) to `.apache-steward-overrides/user.md`, aligning with the snapshot+overrides adoption model introduced in #38.

The `config/` directory was never created in the framework repo under the new architecture, and the `.gitignore` explicitly states that per-user files live in the adopter repo. This PR closes the gap between the architectural intention and the skill implementation.

## Changes

- **4 security skills** now read `.apache-steward-overrides/user.md` instead of `config/user.md`
- **setup-steward/adopt.md** gains Step 9b, scaffolding a `user.md` template during first-time adoption
- **Docs and tool adapters** updated to reference the new path
- Fallback-interactive behaviour is preserved

## Test plan

- `prek run --all-files` passes (check-placeholders + markdownlint)
- No hardcoded project references introduced

### DIFF
--- a/.claude/skills/security-cve-allocate/SKILL.md
+++ b/.claude/skills/security-cve-allocate/SKILL.md
@@ -179,7 +179,7 @@ Before touching the tracker, verify:
    tell the user up front to install `uv` (one command:
    `curl -LsSf https://astral.sh/uv/install.sh | sh`).
 3. **Resolve the user's PMC status.** First try to read it from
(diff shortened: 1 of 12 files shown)


potiuk (approved):
Nice!

PR #61, opened 2026-05-05
feat: surface MISSION mode taxonomy in code, skills, and adopter-facing docs

andreahlert:
## Summary

[`MISSION.md`](MISSION.md) frames the framework around four toggleable **modes** of agent-assisted repository maintainership (A: triage, B: mentoring, C: agent-authored fixes with human review, D: narrowly-scoped auto-merge). The codebase did not yet reflect that taxonomy. This PR surfaces it in three layers, smallest blast radius first:

1. **[`docs/modes.md`](docs/modes.md)** — honest snapshot mapping every shipped skill to a mode (or to *outside the modes* for the setup family) with a `stable / experimental / proposed / off` status legend. Mode B is **proposed** (no skill yet), Mode D is **off** (deliberately not implemented per [`MISSION.md` § Mode D](MISSION.md#technical-scope)). Includes a mode lifecycle and explicit cross-references to MISSION, README, and AGENTS.
2. **`mode:` frontmatter on SKILL.md** — every skill that fits a mode gets `mode: A` or `mode: C`. Setup-family skills are infrastructure and intentionally omit the field. `tools/skill-validator` is extended to recognise `mode` as an optional frontmatter key and validate the value against `{A, B, C}` (Mode D is excluded because shipping a `mode: D` skill would be a sequencing violation, not a typo). New unit tests cover both branches.
3. **README + .asf.yaml** — the `Skill families` table gains a `Modes` column linking to `docs/modes.md`; the GitHub repo description widens from "security vulnerabilities" to "agent-assisted maintainership", since security is one mode of one family rather than the project's identity.

## Why this ordering

The three commits are independently reviewable but ship together because they are mutually load-bearing: the validator change is meaningless without `mode:` in SKILL.md, the README column is meaningless without `docs/modes.md` to link to. Splitting them would force readers to context-switch across PRs to evaluate any single piece. They land as one PR with three commits so each layer remains greppable in `git log`.

## Test plan

- [x] `uv run --project tools/skill-validator pytest -k 'not test_real_repo_passes'` — 33 passed (the `test_real_repo_passes` integration test fails on `main` with 191 pre-existing link-validation violations and is out of scope for this PR)
- [x] `prek run --files <changed-files>` clean on every commit
- [x] `skill-validate` reports zero new mode-related violations against the real repo
- [ ] Maintainer review of the proposed/off statuses in [`docs/modes.md`](docs/modes.md) — these are MISSION interpretations and should match the PMC's read of the sequencing commitments
[...]

PR #63, opened 2026-05-05
docs: Mode B (mentoring) spec — tone guide + adopter contract, no skill code yet

andreahlert:
## Stacked on top of #61

This PR's diff includes the 5 commits from #61 (`feat/modes-taxonomy`) plus 1 new commit (\`b65ccf6\`) that adds the Mode B spec docs. 

## Summary

[\`MISSION.md\` § Mode B](MISSION.md#technical-scope) names conversational mentoring as the highest-value mode and the one off-the-shelf agent tooling skips. Per MISSION sequencing, the spec lands ahead of any skill code so the project's tone choices and hand-off rules are reviewable independently from runtime behaviour.

This PR adds:

- [\`docs/mentoring/spec.md\`](docs/mentoring/spec.md) — scope (in/out), triggers (opt-in only, no auto-fire), teaching register with do/don't examples, hand-off protocol, adopter knobs, open questions for reviewer input.
- [\`docs/mentoring/README.md\`](docs/mentoring/README.md) — family overview pinned at status \`proposed\`, cross-references.
- [\`projects/_template/mentoring-config.md\`](projects/_template/mentoring-config.md) — adopter scaffold for the six required keys the future skill will read.

Updates [\`docs/modes.md\`](docs/modes.md) Mode B section to point at the new spec docs instead of describing them as "follow-up work".

## Why spec-before-code

Three reasons surfaced in #61's docs/modes.md:

1. **Tone is the load-bearing decision.** A mentoring agent that talks down to contributors does net-negative work; a mentoring agent that posts the wrong link in a patient voice still does net-negative work. Reviewing tone alongside skill code conflates "is the runtime behaviour right" with "is the voice right" and loses both signals.
2. **Hand-off rules are a project policy, not a skill detail.** When the agent bows out, who gets `@`-mentioned, and on what surface, is the kind of question that needs PMC sign-off independent of which model or which CLI runs the skill.
3. **Adopter contract before runtime.** The six keys in [\`mentoring-config.md\`](projects/_template/mentoring-config.md) define what an adopter is on the hook for. Naming those before any skill code means an adopter can decide *to adopt* on the contract alone.

## Open questions for reviewers

Listed in [\`spec.md\` § Open questions](docs/mentoring/spec.md#open-questions). The three I'd most like maintainer eyes on:

- Mailing-list mentoring vs. GitHub-only — current draft is GitHub-only.
- Same AI-attribution footer as triage, or distinct — current draft is same wording, different step token.
- Pre-authorized comment classes vs. every-draft-reviewed — current draft is every draft reviewed (Mode-D-shaped sequencing constraint).

## Test plan

- [x] \`prek run --files <changed-files>\` clean
- [x] \`uv run --project tools/skill-validator skill-validate\` reports zero new violations against the mentoring docs (pre-existing 191 link violations on \`main\` unchanged, out of scope)
- [ ] PMC review of the **tone guide** (do/don't examples in [\`spec.md\` § Voice](docs/mentoring/spec.md#voice) and [§ Forbidden](docs/mentoring/spec.md#forbidden))
- [ ] PMC review of the **hand-off triggers** in [\`spec.md\` § Hand-off protocol](docs/mentoring/spec.md#hand-off-protocol) — specifically the \`max_agent_turns: 2\` default
- [ ] PMC sign-off on the **six adopter-config keys** before they harden into a skill contract

## Out of scope

- The skill itself. Lands as a separate PR (\`pr-management-mentor\`, working name) flagged \`mode: B\` + \`experimental\` once this spec is approved.
- Runtime-evaluation harness for measuring mentoring effectiveness — separate workstream tied to MISSION's contributor-sentiment evaluation commitment.
- Auto-fire / pre-authorized comment classes — explicitly Mode-D-shaped per the sequencing constraint.

### DIFF
--- a/docs/mentoring/README.md
+++ b/docs/mentoring/README.md
@@ -1,0 +1,58 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [Mentoring skill family](#mentoring-skill-family)
+  - [Status](#status)
+  - [Adopter contract (proposed)](#adopter-contract-proposed)
+  - [Cross-references](#cross-references)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<!-- SPDX-License-Identifier: Apache-2.0
+     https://www.apache.org/licenses/LICENSE-2.0 -->
+
+# Mentoring skill family
+
+**Status: proposed.** No skill ships yet. The framework lands the
+spec — tone guide, hand-off protocol, adopter contract — ahead
+of any skill code so the project's tone choices are reviewable
+independently of runtime behaviour. See
+[`MISSION.md` § Mode B](../../MISSION.md#technical-scope) for
+the *why*.
+
+Why a framework skill family? Mentoring is named in
+[`MISSION.md`](../../MISSION.md) as the highest-value mode and
+the one off-the-shelf agent tooling skips. Lifting it into the
+framework lets every adopter pick up the tone work and the
+hand-off rules without re-deriving them, and lets the
+framework's contributor-sentiment evaluation cover all adopters
+at once.
+
+## Status
+
+Mode B is **proposed**. No SKILL.md exists yet. This directory
+currently contains:
+
+- [**`spec.md`**](spec.md) — what the future skill should do:
+  scope, triggers, tone, hand-off, adopter knobs.
+
+A prototype skill (`pr-management-mentor`, working name) lands
+in a follow-up PR after this spec is reviewed. The skill ships
+flagged `mode: B` + `experimental` per the
+[mode lifecycle](../modes.md#mode-lifecycle).
+
+## Adopter contract (proposed)
+
+The future skill resolves project-specific content from the
+adopter's `<project-config>/mentoring-config.md` — see the
+template at
+[`projects/_template/mentoring-config.md`](../../projects/_template/mentoring-config.md).
+
+## Cross-references
+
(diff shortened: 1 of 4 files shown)


docs/mentoring/README.md
```diff
@@ -0,0 +1,58 @@
...
+  - [Cross-references](#cross-references)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<!-- SPDX-License-Identifier: Apache-2.0
+     https://www.apache.org/licenses/LICENSE-2.0 -->
+
+# Mentoring skill family
+
+**Status: proposed.** No skill ships yet. The framework lands the
```
potiuk, line 17: Hmmm. I think those are kind of "cross skill" thingies. (same - conversational one next). It's sort of -> "cross-cutting concern". The idea behind those two is:

a) mentoring means "Real human involved"
b) conversational - the sheer fact we are using agentic CLI for those make all our skills "conversational"

So I think we should slightly re-think this one :)

andreahlert: "conversational" being cross-cutting, agreed. The mode reads abstract here because this PR is spec-only. #64 opens the SKILL.md with the concrete definition: "draft a teaching-register comment on a single GitHub issue or PR thread, aimed at a contributor missing repo context". Audience is the contributor, register is teaching, and that is what isolates Mode B from A and C. "Conversational" carries no weight in that framing.

Happy to drop "conversational" from the spec and the mode label, keep "mentoring", and rewrite the spec opener to lead with audience plus register instead of the adjective. Does that address it, or are you questioning the mode itself rather than the name?

potiuk: Hmm. That's interesting approach :) . I will take a close look later today :D

potiuk: But... I think the #64 explained a bit better what you mean and ... it is very much following the mission :)  ... so let me revise it.

potiuk:
OK. Let's get it in. I like it and let's leave conversational in as well for now.

Issue #80, opened 2026-05-06
Adopt 'caveman'-style skill-authoring write-skill to cut per-turn token cost


potiuk:
Skill `description` and `when_to_use` frontmatter fields are loaded into the Claude Code system prompt **every turn** for routing, even when the skill body itself is not invoked. Verbose phrasing in those fields is therefore a permanent steady-state token tax across every session.

### Worked example — [apache/airflow#66444](https://github.com/apache/airflow/pull/66444)

Trimming `maintainer-review`, `pr-triage`, and `pr-stats` frontmatter saved roughly **500 tokens/turn**:

```
maintainer-review: ~420 → ~170 tokens
pr-triage:         ~300 → ~125 tokens
pr-stats:          ~170 → ~105 tokens
```

The `description` / `when_to_use` fields only need what the *router* uses: what the skill does, how it picks targets where that matters, trigger phrases. Anything else (file-path cross-refs, mutation enumerations, history footnotes, "successor to X" notes) belongs in the SKILL.md body — which loads only on invocation.

### Proposed
- Add a "caveman" (or similarly named — terse-by-design) **write-skill** that all skill authoring and editing flows pipe through.
- The skill enforces lean phrasing in frontmatter: drop the categories above, keep what the router needs.
- Update existing skill-authoring guidance and any agent prompts to reference the new write-skill so it gets used by default.
- Sweep existing skills once, applying the new style.

**Why:** every token in skill frontmatter is paid on every turn across every session; lean phrasing is a multiplicative win.

potiuk:
We have other ways to limit tokens.

PR #87, opened 2026-05-07
docs(pr-management-stats): caveman-style rewrite (pilot)

potiuk:
> [!IMPORTANT]
> **Draft pilot.** Marked draft so the side-by-side compare gets reviewed before the umbrella issue spawns subissues for the rest. The full caveman conversion across 80 framework markdown files is gated on this pilot reading cleanly in practice (both for human review and for agent-runtime invocation).

## Summary

Pilot of caveman-style rewrite per the [caveman skill](https://github.com/JuliusBrussee/caveman) rules: drop articles / filler / pleasantries / hedging; fragments OK; arrows for causality; abbreviate common terms; keep technical names, code blocks, and link references exact.

## Sizes (body only — frontmatter + tables + code + links unchanged)

| Metric | Before | After | Δ |
|---|---|---|---|
| Words | 2719 | 2060 | **-24%** |
| Characters | 19469 | 16295 | -16% |
| Lines | 304 | 422 | +39% (fragmentation per style) |

The 75% target the caveman skill claims is for chat-mode prose. SKILL.md prose is dense in tables, code, and links that don't compress; the win lives in the prose-paragraph sections (Golden rules, step descriptions, "What this skill does NOT do").

## What stayed verbatim

- All YAML frontmatter (`name`, `description`, `when_to_use`) — compression there could hurt the matching layer's trigger recall.
- All technical identifiers (skill names, file paths, GraphQL field names, label strings).
- All code blocks.
- All link targets (relative paths to detail files, anchor links).
- All placeholder tokens (`<upstream>`, `<viewer>`, `<repo>`, `<project-config>`).
- All the load-bearing values (cutoff defaults, GraphQL pagination size, pressure-score weights).

## Why this skill as the pilot

- Read-only — no safety regression risk if the agent's reading of the compressed prose drifts.
- Recently rewritten — clean baseline (no historical cruft).
- Mid-size — big enough to demonstrate compression, small enough for one human pass.
- Structurally clear — golden-rule list, numbered steps, `Skill does NOT do` enumerated list — all good caveman targets.

## Test plan

- [x] `prek run --files <SKILL.md>` clean (markdownlint, typos, check-placeholders all pass).
- [x] **Manual eyeball pass** — compare side-by-side against `main` and confirm no technical-substance regression.
- [ ] **Live invocation test** — next time the maintainer runs `pr-management-stats` against `apache/airflow`, confirm the agent still produces the same dashboard shape (same panels, same numbers).
- [ ] **Trigger recall** — confirm phrases like "run PR stats" / "how is the queue doing" still match this skill (the `description` and `when_to_use` were kept readable for this reason).

## Next step (if pilot reads cleanly)

Open umbrella issue `Convert all skills + tools to caveman style` with one subissue per file (~80 files). Each subissue follows the same recipe: caveman the body, leave frontmatter / code / links alone, target ~20-30% prose-word reduction.

## Out of scope

- This PR is the pilot for the umbrella plan only — no other files touched.
- The caveman skill in `JuliusBrussee/caveman` is a *communication mode*; we're applying it as a *documentation style*. The two uses overlap but aren't identical — if practice shows the doc-style application doesn't carry, the umbrella issue gets re-scoped or dropped.

### DIFF
--- a/.claude/skills/pr-management-stats/SKILL.md
+++ b/.claude/skills/pr-management-stats/SKILL.md
@@ -1,16 +1,16 @@
 ---
 name: pr-management-stats
 description: |
-  Read-only maintainer dashboard for the open-PR backlog of <upstream>.
-  Surfaces a health rating, prioritised action recommendations, weekly closure
-  velocity trends, area pressure ranking, and a triage-funnel breakdown — with
-  the underlying area-grouped tables as a collapsible details section.
+  Read-only maintainer dashboard for open-PR backlog of <upstream>.
+  Surfaces health rating, prioritised action recommendations, weekly
+  closure velocity trends, area pressure ranking, triage-funnel
+  breakdown — area-grouped tables collapsed at bottom.
 when_to_use: |
-  When the user asks "how is the PR queue doing", "run PR stats", "what should
-  I do today", "show me the trends", "where is queue pressure sitting", or any
-  variation on "give me the maintainer view of the backlog". Good as a daily
-  health check, before or after a triage sweep, or as an input to a planning
-  session.
+  When user asks "how is the PR queue doing", "run PR stats", "what
+  should I do today", "show me the trends", "where is queue pressure
+  sitting", or any variation on "give me maintainer view of backlog".
+  Daily health check, before/after triage sweep, planning-session
+  input.
 license: Apache-2.0
 ---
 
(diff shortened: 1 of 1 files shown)


potiuk:
Hey Everyone here: WDYT ? (@andreahlert @johnslavik @choo121600 ?  especially). Since the mission of the project is to be affordable, I think this makes a lot of sense even if prose gets a little bit more difficult to read by human.

Let me know what you think -> if this will be acceptable, I plan to create an issue with tasks and we could attempt to convert all of the SKILLS/prose this way?

Particularly when testing wiht @shahar1 - the tokens on smaller subscription got very quickly eaten.

potiuk:
BTW. I created a number of issues already in https://github.com/orgs/apache/projects/627 and we could add some more ideas and try to organize it in some swimlanes/ areas/ mostly around the A/B/C/D modes from mission but also in other ways - I am about to kick-start project this afternoon in an email and throw some ideas in for discussion - but this one might help to onboard people who do not have that many tokens  at their disposal

andreahlert:
> Hey Everyone here: WDYT ? (@andreahlert @johnslavik @choo121600 ? especially). Since the mission of the project is to be affordable, I think this makes a lot of sense even if prose gets a little bit more difficult to read by human.
> 
> Let me know what you think -> if this will be acceptable, I plan to create an issue with tasks and we could attempt to convert all of the SKILLS/prose this way?
> 
> Particularly when testing wiht @shahar1 - the tokens on smaller subscription got very quickly eaten.

hey! Thanks for that. i took some time to fully test with Claude,read #80 end to end and ran the numbers. scripts and raw data: [[gist](https://gist.github.com/andreahlert/f76cd049b7418aa39b34f73a4c5df8d7)]

aligned with #80 's core thesis. frontmatter (`description` + `when_to_use`) is loaded into the system prompt every turn for routing, which makes any verbosity there a multiplicative steady-state tax. the worked example in #80 (`maintainer-review`, `pr-triage`, `pr-stats` saving ~500 tokens/turn combined) is real and the mechanism is sound. trimming frontmatter across 80 skills is a high-ROI move and i support it.

my disagreement is with how this PR scopes that goal.

#80 calls for a write-skill that enforces lean phrasing in `description` and `when_to_use`. this PR does that, and also rewrites the entire SKILL.md body in caveman style. those are two different optimizations with very different ROI:

frontmatter trim (the #80 goal): permanent, every-turn, multiplicative across sessions. real win.

body caveman: only loads on skill invocation, sits behind prompt caching at 10% on cache hits, and the body itself is a small fraction of the turn input. measured end to end:

- PR's `-24%` is words. claude tokenizer gives `-9.4%` for the file in isolation.
- realistic mid-session turn (system prompt, tools, CLAUDE.md, prior conversation, skill body): `-2.10%`.
- long pilot sessions (100k+ history): `-0.4%` to `-1%`.

so the umbrella, if it ships as "caveman the body of 80 skills", commits review budget and locks in style overhead (frontmatter loss of articles/connectives in prose, harder review surface) for `-0.5%` to `-2%` per turn. that's the wrong order of magnitude vs the frontmatter saving #80 actually points at.

separate concern: the PR body says frontmatter stayed verbatim. it didn't. the diff rewrites both `description` and `when_to_use`, which is the trigger and recall surface. that needs to either land with a recall test or be split into its own PR so the routing impact is auditable.

my position:

1. land the frontmatter trim across the 80 skills as #80 describes. that's the actual multiplicative win and i'll review every PR.
2. drop the body caveman rewrite from this PR. defer it until A/B/C/D modes are shipped, we have telemetry on which skill bodies are actually hot, and we can target compression where it pays.
3. complementary cheap moves that probably dominate body caveman on ROI right now: haiku 4.5 for skill routing (20-probe A/B, 90% accuracy, 15x cheaper than opus, 5x faster, gateable with a name validator) and boilerplate dedup of `Adopter overrides`, `Snapshot drift`, `Prerequisites` across SKILL.md (~5,900 tokens one shot, no maintenance debt).

on #80: yes. on the body half of this PR: not yet.

potiuk:
Whoa :) ... This is very cool analysis.. I've learned a TON from it... I would love to hear other's comments and tests maybe but that analysis looks pretty sound!

choo121600:
I also agree with #80. 

But before optimizing the prose style itself, I think the current description and when_to_use fields already have some fundamental issues in terms of routing quality and trigger precision. 

To me, that is the more important problem to solve first, and likely the change that would make the project meaningfully cheaper to operate in practice.


I think the discussion in #92 is closely related to this direction as well :)

potiuk:
Closing it for now... We have ideas in #92 that are much more polished to achieve better goals!

PR #100, opened 2026-05-10
fix(pre-commit): silent-pass CWD bug + link/anchor rot cleanup

choo121600:
## Summary
* Fixes a long-standing silent-pass bug in `skill-validate`.

  which changed the working directory before startup. As a result,
  `Path(".claude/skills")` resolved inside the validator subtree,
  zero skill files were scanned, and validation falsely reported `OK`.

  This PR switches to `--project` and adds repo-root discovery as an
  additional safeguard.

* Improves validator behavior:

  * ignore `<token>` placeholder URLs
  * ignore ellipsis/example URLs (`...`, `…`)
  * ignore markdown links inside inline/fenced code blocks
  * align repeated-heading anchors with GitHub slug rules

* Adds missing prek hooks for the validator itself:

  * `ruff-check`
  * `ruff-format`
  * `mypy`
  * `pytest`

* Repairs accumulated link rot discovered once validation started
  scanning real files again (`134 -> 0` across 14 skill files),
  including moved docs paths, renamed anchors, outdated references,
  and stale heading slugs.

## Why

Previous validations reported `0 violations`, but that result came from
the silent-pass bug above. Once the validator started scanning actual
skill files again, it surfaced stale link and anchor references that
had accumulated during earlier docs and `AGENTS.md` restructures.

### DIFF
--- a/.claude/skills/pr-management-code-review/criteria.md
+++ b/.claude/skills/pr-management-code-review/criteria.md
@@ -44,19 +44,16 @@ table; see [`review-flow.md#area-specific-overlay`](review-flow.md).
 
 ## Categories — link out to the source section
 
-The category headings below are the **abstract names** the skill
-uses when grouping findings. The concrete URL each one links to
-lives in the adopter's
+The list below is the **abstract canonical category list** the
+skill uses when grouping findings. The concrete review-doc URL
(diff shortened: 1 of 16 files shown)


potiuk (approved):
Nice!

PR #110, opened 2026-05-10
feat(security-cve-allocate): title-only allocation recipe

potiuk:
## Summary

- The Vulnogram allocation form accepts a bare title — product / `packageName`, CWE, affected versions, public summary, and reporter credits are all easier to set correctly *after* the tracker body has settled, during the existing Step 6 `security-issue-sync` handoff (and the `vulnogram-api-record-update` push that follows). Drop them from Step 3's paste-into-form recipe so the user copies one field (title) and clicks *Allocate*.
- Update the non-PMC relay message in the same vein: only the URL, the stripped title, and the "paste the allocated CVE back here" line. The scope/product/package-name block is removed — the relayed PMC member doesn't need it at allocation time, and the product lands during sync anyway.
- Refresh the Step 1 scope-label blocker rationale. The check stays (sync depends on the scope), but its old "much easier to set product at allocation time than after" justification contradicts the new model and would confuse a future reader. New rationale: surface it as a blocker so the user fixes it once, up front, instead of being interrupted mid-flow when Step 4's JSON regeneration or Step 6's sync handoff can't map the product.

## Test plan

- [ ] Invoke `security-cve-allocate` on a sample tracker; confirm Step 3's printed recipe contains only the title-paste step (no Product / CWE / Affected versions / Summary / Reporter credits list).
- [ ] Confirm the non-PMC relay message no longer contains the scope/product/package-name block.
- [ ] Confirm Step 1 still surfaces a missing scope label as a blocker, with the refreshed rationale.
- [ ] Walk a real allocation end-to-end and verify Step 6's `security-issue-sync` handoff lands product / packageName / CWE / summary / credits into the CVE JSON correctly (this is the load-bearing claim of the change).

### DIFF
--- a/.claude/skills/security-cve-allocate/SKILL.md
+++ b/.claude/skills/security-cve-allocate/SKILL.md
@@ -254,10 +254,14 @@ Blocker checks — if any fail, stop and surface the failure:
 - **Scope label set.** The CVE record's `product` / `packageName`
   fields depend on the scope (`airflow` → `apache-airflow`,
   `providers` → `apache-airflow-providers-<name>`,
-  `chart` → `apache-airflow-helm-chart`). If no scope label is
-  set, stop and ask the user to confirm the scope before allocating
-  — it is **much** easier to set the right product at allocation
-  time than to correct it after the Vulnogram record has a CVE ID.
+  `chart` → `apache-airflow-helm-chart`). Allocation itself is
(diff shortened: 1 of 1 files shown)


potiuk:
Something we saw with @shahar1 -> only title is neded for allocation of CVE - all the rest is confusing.

PR #112, opened 2026-05-11
fix(pr-management-triage): post-filter on conclusion to find action_required runs

potiuk:
## Summary

The `?status=action_required` query parameter on `GET /repos/{owner}/{repo}/actions/runs` does not return runs awaiting maintainer approval. GitHub returns those runs as `status: "completed"` with `conclusion: "action_required"` (the queueing phase is finished, but the run is blocked pending approval).

The triage skill used `?status=action_required` in 4 places — none of them returned anything, which let `pending_workflow_approval` PRs slip through classification as `passing` and would have caused the `mark-ready` guard to silently fail-open.

## Empirical evidence

Hit live during a triage sweep on `apache/airflow` today: the mandatory `action_required` REST index missed 13 pending workflow runs across 4 first-time-contributor PRs (caught by the rollup-pattern Real-CI guard, but only because all 4 had no real CI in the rollup). One of the two `mark-ready`-classified PRs (#65974) had 3 pending workflow runs at its head SHA — the broken pre-check would have label-added it as ready for review.

## Changes

Replace `?status=action_required` with `?per_page=N` + jq post-filter `select(.conclusion == "action_required")` in 4 places:

- `fetch-and-batch.md` — mandatory per-page REST index
- `actions.md` — `mark-ready` pre-mutation guard
- `actions.md` — `mark-ready-with-ping` pre-mutation guard
- `actions.md` — `approve-workflow` recipe (used to discover the run IDs to approve)

Add explanatory text in `fetch-and-batch.md` and inline comments in the recipes so a future reader doesn't reintroduce the broken filter.

The `gh api -X POST .../approve` mutation itself works against runs in this state (verified: 16 successful approvals during the same triage sweep) — only the discovery query was broken.

## Test plan

- [x] Visual diff review — all 4 buggy uses replaced; explanatory text only references the broken filter to warn against it
- [x] `prek run --files` passes (markdownlint, skill-validate, etc.)
- [x] Verified the corrected jq filter returns the expected runs against `apache/airflow` (16 runs across 5 PRs)

---

- [X] Yes — Claude Code (Opus 4.7)

Generated-by: Claude Code (Opus 4.7) following [the guidelines](https://github.com/apache/airflow/blob/main/contributing-docs/05_pull_requests.rst#gen-ai-assisted-contributions)

### DIFF
--- a/.claude/skills/pr-management-triage/actions.md
+++ b/.claude/skills/pr-management-triage/actions.md
@@ -189,10 +189,14 @@ the maintainer queue fills with PRs whose CI has not actually
 executed.
 
 ```bash
-# Pre-check: index action_required runs repo-wide, then look up head SHA
+# Pre-check: index action_required runs at the head SHA.
+# Note: runs awaiting approval are returned as `status: "completed"`
+# with `conclusion: "action_required"`. The query parameter
+# `?status=action_required` matches no runs and would silently
+# return an empty result — post-filter on `conclusion` instead.
 head_sha=$(gh api "repos/<owner>/<repo>/pulls/<N>" --jq '.head.sha')
-pending=$(gh api "repos/<owner>/<repo>/actions/runs?head_sha=${head_sha}&status=action_required&per_page=10" \
-  --jq '.workflow_runs | length')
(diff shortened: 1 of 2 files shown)


potiuk:
Found during testing with steward adopted in Airflow :)

Issue #118, opened 2026-05-11
Meta: audit and trim oversized skill frontmatter


choo121600:
## Context

Claude Code enforces a per-skill metadata budget of **1,536 characters** for
`description` + `when_to_use` combined. Skills that exceed this budget get
their frontmatter truncated at routing time, weakening discovery.

#103 trimmed `pr-management-code-review` from **1,708 → 944 chars** using
this principle:

> **Compress depth, preserve surface.**
> Frontmatter is the *routing layer* — keep only trigger phrases, short task
> descriptions, and routing cues. Rationale, implementation notes, long
> examples, criteria source paths, and sub-step descriptions belong in the
> body.

This meta-issue tracks applying the same audit to every other skill.

## Trimming rules

**Keep in frontmatter**
- Concise trigger phrases (every literal phrase a user might type)
- Routing cues (`use when …`, `skip when …`)
- High-signal task descriptions

**Move to body**
- Rationale / "why"
- Implementation notes
- Long examples
- "Distinct from X" explanations
- Criteria source paths
- Sub-step descriptions

**Non-negotiable:** every literal trigger phrase from the original
`when_to_use` must survive verbatim — including niche ones. Routing recall
must not regress.

## Per-PR checklist

- [ ] PR description shows before/after char counts (`description`, `when_to_use`, total, budget margin)
- [ ] PR description lists where moved content now lives
- [ ] Trigger phrases preserved verbatim (call this out explicitly)

## Skills to audit

- [ ] `pr-management-mentor` — 1,514 / 1,536 (margin: 22) #119
- [ ] `security-issue-triage` — 1,405 / 1,536 #120 
- [ ] `setup-isolated-setup-install` — 1,292 #121 
- [ ] `security-issue-fix` — 1,249 #122 
- [ ] `setup-steward` — 1,231 #123 
- [ ] `security-cve-allocate` — 1,197 #124 
- [ ] `pr-management-triage` — 1,183 #125 
- [ ] `security-issue-import-from-md` — 1,149 # 126
- [ ] `setup-shared-config-sync` — 1,138 #127
- [ ] `setup-override-upstream` — 1,096 #128
- [ ] `security-issue-import` — 1,073
- [ ] `security-issue-import-from-pr` — 1,036
- [ ] `write-skill` — 1,036
- [ ] `security-issue-invalidate` — 1,025
- [ ] `setup-isolated-setup-verify` — 1,024
- [ ] `setup-isolated-setup-update` — 996
[...]

PR #128, opened 2026-05-11
skills/setup-override-upstream: trim frontmatter to fit metadata budget

choo121600:
## Summary

Trims `setup-override-upstream` frontmatter (`description` + `when_to_use`) from **1,096 → 892** characters.

Same principle as #103 and the rest of #118's audit pass: frontmatter is the routing layer. The body covers the generalisability test, the design-decision example list, and the pre-commit step.

Tracking: #118

## Before / after

|              | before | after | Δ      |
|--------------|-------:|------:|-------:|
| description  | 736    | 532   | -204   |
| when_to_use  | 360    | 360   | 0      |
| **total**    | **1,096** | **892** | **-204** |
| budget margin | 440   | 644   | +204   |
| budget        | 1,536  | 1,536 |        |

`when_to_use` is untouched — every line is either a literal trigger phrase or a routing skip cue (*"typically after running the override locally for a while"*) and has to stay verbatim.

## What moved where

| Detail                                                                                                              | Where it lives now                            |
|---------------------------------------------------------------------------------------------------------------------|-----------------------------------------------|
| `(some overrides encode project-specific decisions and should stay local)` (rationale parenthetical)                | Body's generalisability discussion           |
| `(configurable parameter, optional step, changed default, etc.)` (abstraction-example list)                         | Body's design-decision section                |
| `runs the framework's pre-commit` (implementation detail)                                                            | Body Walk-through / standard project workflow |

The post-merge override-removal flow (*"After the PR merges and the adopter runs `/setup-steward upgrade`, the override file is no longer needed and the skill prompts for its removal"*) is **kept** in the frontmatter — it is the only routing signal that tells the agent this skill cleans up after itself.

## Trigger-phrase preservation

Every literal trigger phrase from the original `when_to_use` is preserved verbatim:

- `"upstream my override"`
- `"promote this override to the framework"`
- `"convert my local modification into a steward feature"`
- `"make this override a framework feature"`
- `"open a PR to apache-steward for this override"`

Routing skip cue preserved:

- *"typically after running the override locally for a while and deciding the change is worth contributing back"*

Routing recall does not regress.

### DIFF
--- a/.claude/skills/setup-override-upstream/SKILL.md
+++ b/.claude/skills/setup-override-upstream/SKILL.md
@@ -4,17 +4,13 @@ description: |
   Walk an adopter through promoting a local
   `.apache-steward-overrides/<skill>.md` file into a PR
   against `apache/airflow-steward`. Lists the adopter's
-  overrides, helps the user pick one, reads it alongside the
-  framework skill it modifies, helps decide whether the
-  change is generalisable (some overrides encode project-
-  specific decisions and should stay local), designs the
-  framework-level abstraction (configurable parameter,
-  optional step, changed default, etc.), implements it in
-  the user's local apache-steward clone, runs the
-  framework's pre-commit, and opens the PR. After the PR
-  merges and the adopter runs `/setup-steward upgrade`, the
-  override file in the adopter repo is no longer needed and
-  the skill prompts for its removal.
+  overrides, helps pick one, reads it alongside the framework
+  skill it modifies, decides whether the change is
+  generalisable, designs the framework-level abstraction,
+  implements it in the user's local apache-steward clone, and
+  opens the PR. After the PR merges and the adopter runs
+  `/setup-steward upgrade`, the override file is no longer
+  needed and the skill prompts for its removal.
 when_to_use: |
   Invoke when the user says "upstream my override", "promote
   this override to the framework", "convert my local


choo121600 (commented):
The trim is clean and the body accounting checks out.

A few observations:

**Budget note**: the pre-change total (1,096 chars) was already 440 chars under the 1,536 limit, so this PR was not required by the validator. It is a legitimate style improvement — the post-trim description is more concise — but worth flagging so the tracking issue (#118) records that this file was never a compliance failure.

**`runs the framework's pre-commit`**: removed from the description chain. Confirmed still present in the body at Walk-through Step 5 §3 (`prek run --all-files`). No information loss — it is an implementation detail, not a routing signal, and belongs in the body.

**Trigger phrases**: all five literals in `when_to_use` are unchanged. Routing recall does not regress.

**`/setup-steward upgrade` cross-reference**: kept in the frontmatter description (post-merge cleanup flow). Correct placement — that sentence is a routing signal telling the agent this skill self-cleans after upgrade.

Everything else the description previously contained (generalisability parenthetical, abstraction-example list) is accounted for in the body. Approving.

> 🤖 *This review was drafted by an AI-assisted tool. An apache-steward maintainer has read and confirmed it before posting; mistakes are still possible — point them out and the maintainer will sort it. See [CONTRIBUTING.md](https://github.com/apache/airflow-steward/blob/main/CONTRIBUTING.md) for what an apache-steward maintainer review covers.*

PR #158, opened 2026-05-15
An idea for testing fixture-based eval harness for skill steps

justinmclean:
## What this adds

A small, model-agnostic eval harness under `tools/skill-evals/` for
verifying that skill steps produce the expected structured output.

The runner prints the exact system prompt + user prompt for each fixture
case so you can paste it into any model and compare the response against
`expected.json` so as to be vendor-neutral.

This is an early idea; I'm sharing it for feedback before we go further.

- Should fixture cases live next to their skill rather than under `tools/skill-evals/evals/`?
- Is manual paste-and-compare enough, or do we want automated scoring via the API? If so, how do we make that vendor-neutral?
- Is there a simpler approach entirely?

Happy to rework or close in favour of a better idea.


choo121600 (commented):
Wow, this looks really great!!!

For (1), I think tools/skill-evals/evals/ looks like the better location. Putting it under the agent path does look cleaner structurally, but I worry there’s a risk of eval fixtures accidentally mixing into agent context at some point, so keeping them separated seems safer.

For (2), I think manual paste-and-compare is sufficient for now. If we eventually automate model execution, I’d prefer a CLI-style approach (e.g. claude -p) over direct API integration.

potiuk:
Really cool . I love how clean and nice it is.. I would probably use Yaml not json ;) (but this is just my eastethics looking at all this extra {} :) . 

> For (1), I think tools/skill-evals/evals/ looks like the better location. Putting it under the agent path does look cleaner structurally, but I worry there’s a risk of eval fixtures accidentally mixing into agent context at some point, so keeping them separated seems safer.

Agree.

> For (2), I think manual paste-and-compare is sufficient for now. If we eventually automate model execution, I’d prefer a CLI-style approach (e.g. claude -p) over direct API integration.

I think it's good to start. We could even add a SKILL to run all such evals locally (asking the agent to essentially copy & paste) things - and have us to do periodic re-run with our own agents (We could even have this skill written in the way that it would setup some reminders to do it from time to time) . I think that should do for now.

justinmclean:
I updated and expanded on this for the cover two skills.

runner.py now assembles the system prompt by extracting the relevant section directly from SKILL.md at run time (via a step-config.json pointer in each fixtures directory), rather than relying on a static copy. A change to the skill rules is immediately visible in the prompt — if it would cause the model to produce different output, the test fails.

External tool calls (GitHub CLI, Gmail MCP, canned-response scan) are never executed during evals. Their outputs are pre-rendered as structured text in each case's report.md and injected into the user turn as mock data. This keeps inputs fully deterministic and requires no network access or API credentials to run.

Removed the hardcoded SYSTEM_PROMPT fallback in runner.py. 

65/65 cases pass (32 import, 33 triage).

justinmclean:
BTW, there's no need for a manual copy and paste unless you want to. You can ask Claude Desktop to run the tests, and it will, no SKILL file is needed.

potiuk:
Wowowowoowow! That looks super cool :)

potiuk:
Some results of those agentic skils are terrifying :) .... But we can also use agents to review those changes now:

Screenshot 2026-05-16 at 13 17 42

potiuk (commented):
Thanks for putting this up early — the fixture corpus is impressively thorough, and the model-agnostic paste-and-compare framing is a sensible MVP. A few things to flag before going further; left some specifics inline.

## Direction / answers to your questions

**Fixtures next to the skill vs. under `tools/skill-evals/`?**
Keep under `tools/skill-evals/`. Co-locating 620 fixture files inside `.claude/skills/<name>/` would bloat the agent's view of "what is the skill" and the skill loader has to walk past them. A one-line pointer from each `SKILL.md` to the eval suite is enough cross-linking.

**Vendor-neutral automation?**
Paste-and-compare collapses at 206 cases. `eval_runner.py` already proves the wiring exists — generalise it behind a `call_model(system, user) -> str` interface keyed by `--provider {anthropic,openai,bedrock,vertex}`, each a ~30-line adapter. JSON extraction + the comparator stay vendor-neutral.

**Simpler approach?**
The harness itself is small (~250 lines) — not the place to simplify. Where you can: cut the duplication between `runner.py` and `eval_runner.py` by extracting `prepare_case(case_dir, fixtures_dir, _cache) -> (label, system, user, expected)`.

## Two blocking things not anchorable to a single line

1. **No `pyproject.toml`.** Neither `uv run --project tools/skill-evals skill-eval ...` (README) nor `python3 -m skill_evals.eval_runner ...` (eval_runner docstring) resolves without a `pyproject.toml` under `tools/skill-evals/` declaring the `skill_evals` src-layout package and the `skill-eval` console script. The PR ships none.

2. **README's "All five cases passed" is unverifiable.** `runner.py` doesn't compare, `eval_runner.py` isn't wired to CI in this PR. Either drop the claim or land a regenerable summary.

## Review ergonomics

620 files in one PR is brutal to review. The framework (runners + 1 suite) could land first; the rest stack as follow-ups — would let the runner bugs (see inline) get caught before the fixture corpus is locked in around them.

## Risk summary

| Risk | Severity |
|------|----------|
| `extract_skill_section` truncates Step 2 of `security-cve-allocate` today | **high** — silently passes a wrong test |
| Missing `pyproject.toml` — neither documented runner invocation works | **high** |
| `extract_json` `MULTILINE` regex strips internal fences too | medium |
| README oversells "structural assertions" and "all five passed" | low (docs) |

tools/skill-evals/src/skill_evals/runner.py
```diff
@@ -0,0 +1,251 @@
...
+        (i for i, line in enumerate(lines) if line.rstrip() == heading_stripped),
+        None,
+    )
+    if start is None:
+        raise ValueError(f"Heading {heading!r} not found in {skill_md_path}")
+
+    end = len(lines)
+    for i in range(start + 1, len(lines)):
+        depth = len(lines[i]) - len(lines[i].lstrip("#"))
+        if depth > 0 and depth <= heading_level and lines[i].lstrip("#").startswith(" "):
```
potiuk: This depth check has a confirmed bug. `lstrip("#")` strips *all* leading `#`s, so a python/shell comment like `# Leading passes twice…` satisfies `depth=1, lstrip("#").startswith(" ")` and is misread as a heading that terminates the section.

Concrete repro on this branch: `.claude/skills/security-cve-allocate/SKILL.md` line 309 (`# Leading passes twice — strip order reveals nested tags.`) falls inside Step 2 (heading at line 267). When `step-2-title-normalize/fixtures/step-config.json` runs through `load_step_config`, the extracted section is silently truncated at 309 — `case-4-over-strip-warning` is very likely passing against a half-prompt rather than the real Step 2 text.

Fix: track fenced-code-block state and use `re.match(r"^#{1,6} ", line)` for "this is a markdown heading" instead of the `lstrip` trick.

tools/skill-evals/src/skill_evals/runner.py
```diff
@@ -0,0 +1,251 @@
...
+    """
+    if (path / "report.md").exists():
+        return [(path, path.parent)]
+    # Direct fixtures dir — all cases share the same fixtures dir.
+    direct = sorted(p for p in path.iterdir() if p.is_dir() and (p / "report.md").exists())
+    if direct:
+        return [(p, path) for p in direct]
+    # Recursive search — e.g. skill dir spanning multiple steps.
+    results = []
+    for fixtures_dir in sorted(path.rglob("fixtures")):
```
potiuk, line 214: Minor: `rglob("fixtures")` matches every `fixtures` directory under `path`. Today the layout has one level, but a stray nested `fixtures/` (someone copy-pasting a sub-case) would be double-counted. Worth either restricting depth or de-duping by path.

tools/skill-evals/src/skill_evals/runner.py
```diff
@@ -0,0 +1,251 @@
...
+    # the same fixtures dir (common when running a whole skill at once).
+    _step_config_cache: dict[Path, tuple[str, str]] = {}
+
+    for case_dir, fixtures_dir in cases:
+        if fixtures_dir not in _step_config_cache:
+            _step_config_cache[fixtures_dir] = load_step_config(fixtures_dir)
+        system_prompt, user_prompt_template = _step_config_cache[fixtures_dir]
+
+        corpus, roster, report, expected = load_case(case_dir)
+        user_prompt = user_prompt_template.format(
```
potiuk: `.format()` on a user-supplied template is fragile — if anyone ever puts a literal `{` in a future `user-prompt-template.md` (e.g. a JSON schema fragment), this raises `KeyError`. Current templates are clean, but worth either documenting the contract in `user-prompt-template.md` files or switching to `string.Template` which uses `$name` and ignores stray braces.

tools/skill-evals/src/skill_evals/eval_runner.py
```diff
@@ -0,0 +1,195 @@
...
+from skill_evals.runner import find_cases, load_case, load_step_config, build_corpus_text, build_roster_text
+
+# ---------------------------------------------------------------------------
+# JSON extraction
+# ---------------------------------------------------------------------------
+
+def extract_json(text: str) -> dict:
+    """Extract the first JSON object from a model response."""
+    # Strip markdown code fences
+    text = re.sub(r"^```(?:json)?\s*", "", text.strip(), flags=re.MULTILINE)
```
potiuk: `re.MULTILINE` makes `^` and `$` match every line — every ``` triplet anywhere in the response gets stripped, not just the outer fence pair. The fallback `{...}` brace-counter then runs against a mangled string. Drop `MULTILINE` (strip only at string ends), or extract the first balanced ```...``` block deliberately.

tools/skill-evals/src/skill_evals/eval_runner.py
```diff
@@ -0,0 +1,195 @@
...
+    """
+    failures = []
+    for key, exp_val in expected.items():
+        if key not in actual:
+            failures.append(f"  missing key '{key}' (expected {exp_val!r})")
+            continue
+        act_val = actual[key]
+        if exp_val != act_val:
+            failures.append(f"  '{key}': expected {exp_val!r}, got {act_val!r}")
+    # Flag unexpected keys that are booleans and wrong (lenient on extras)
```
potiuk: The README explains that composition steps use structural flags (`has_security_model_quote`, `has_bare_issue_numbers`, `mention_handles`) to avoid brittle prose match. But `compare` just does `exp_val != act_val` per key — meaning the *model* has to compute and emit those derived booleans about its own prose, not the runner. That's (a) not what the README implies and (b) a weaker test than the runner deriving the booleans from the response. Either clarify the README or move the derivation into the comparator.

Also: the dead comment on this line (`Flag unexpected keys that are booleans and wrong (lenient on extras)`) describes behavior that isn't implemented — remove or implement.

tools/skill-evals/README.md
```diff
@@ -0,0 +1,86 @@
...
+- **security-issue-fix** — 30 cases across 10 steps (2, 4a, 4b, 4c, 4d, 4e, 4f, 4g, 5, 10)
+- **security-issue-invalidate** — 24 cases across 9 steps (2, 3, 4, 5a, 5b, 5d, 5e, 5f, 7)
+- **security-issue-import-from-md** — 11 cases across 4 steps (1, 2, 4, 6)
+- **security-issue-import-from-pr** — 13 cases across 4 steps (2, 3, 6, 8)
+
+## Run
+
+```bash
+# All cases for a skill
+uv run --project tools/skill-evals skill-eval \
```
potiuk, line 21: This invocation needs a `pyproject.toml` under `tools/skill-evals/` declaring the src-layout `skill_evals` package and the `skill-eval` console script. None is shipped in this PR — `uv run --project tools/skill-evals skill-eval ...` will fail with "package not found" today. Same applies to the `python3 -m skill_evals.eval_runner ...` invocation in `eval_runner.py`'s docstring.

potiuk (approved):

potiuk (commented):
## Reviewed at SHA `1142fb5` (3 commits since my first review: `1c74f9c`, `2076710`, `1142fb5`)

### Status of my prior comments

| # | Topic | Status |
|---|-------|--------|
| 1 | `extract_skill_section` heading bug | **Fixed.** New code tracks ```` / `~~~` fence state and uses `re.match(r"^(#{1,6}) ", …)` instead of the `lstrip("#")` trick. Verified concretely: the line-309 `# Leading…` in `security-cve-allocate/SKILL.md` is inside a fenced block, so Step 2 extraction is no longer truncated. |
| 2 | `rglob("fixtures")` double-count | **Fixed.** `seen_fixtures` set + `is_relative_to` check (runner.py:213-219). |
| 3 | `.format()` fragility | **Fixed.** Slot contract documented at runner.py:48-50, plus an actionable error wrapped around the `.format()` call (runner.py:263-268). Reasonable choice to keep `str.format` rather than `string.Template`. |
| 4 | `extract_json` `MULTILINE` regex | **Moot.** `eval_runner.py` removed — when automation lands as a follow-up, please re-check this. |
| 5 | `compare` vs README structural-assertions claim | **Moot** (also lifted by `eval_runner.py` removal). The remaining README paragraph now correctly describes a *fixture* convention rather than runner behavior. |
| 6 | Missing `pyproject.toml` | **Fixed.** Hatchling backend, `[project.scripts]` console-script entry, `requires-python >= 3.11` matches the PEP-604 type hints, `dependencies = []` matches stdlib-only imports. `uv run --project tools/skill-evals skill-eval …` should resolve cleanly. |

The "All five cases passed" claim is also gone from the README.

### Direction

Cutting `eval_runner.py` was the right call for an MVP — the harness is now self-contained, has no third-party deps, and the print-and-paste workflow is the right primitive to land first. Automation is a clean follow-up.

### Minor follow-ups (non-blocking)

- `pyproject.toml` declares `dev = ["pytest>=8.0", …]` but ships no `tests/`. `extract_skill_section`, `find_cases`, and the fence-state tracker are exactly the kind of small pure-function logic that benefits most from a handful of unit tests — would have caught the original truncation bug before it ever shipped. Worth a follow-up PR.
- README still describes the suite as "currently manual — it prints prompts for human review" (line 70). Accurate as of this PR; just flagging that the "CI considerations" paragraph below it (lines 71-74) reads as deferred-but-planned, which is appropriate.

Blockers cleared from my side.

potiuk (approved):

potiuk:
Let's get it in :)

choo121600:
COOOL🙌

PR #163, opened 2026-05-15
feat(release-management): propose family with 14-step lifecycle and non-ASF backend abstraction

andreahlert:
## Summary

Propose the **release-management skill family** docs-first, mirroring the Mentoring precedent (spec before code). Ten `release-*` skills compose the canonical 14-step ASF release lifecycle, from planning issue and version bump through `[ANNOUNCE]`, archive sweep, and per-release audit log. The family lands as docs only in this PR; the skills follow flagged `experimental` in subsequent PRs.

Non-ASF adopters are first-class. Three backend switches parametrise distribution (`svnpubsub` | `github-releases` | `s3` | `self-hosted`), approval (`dev-list-vote` | `github-discussion` | `pr-approval` | `maintainer-roster`), and announcement (`announce-list` | `github-release-notes` | `site-post` | `discord-channel`). The 14 steps stay identical across backends; only the command set the agent emits changes. ASF TLPs are pinned to `dev-list-vote` and `announce-list` per [release-policy.html](https://www.apache.org/legal/release-policy.html).

Two non-negotiable state-change boundaries cross every Drafting skill:

- Agent never holds, invokes, or proxies the Release Manager's private signing key. Steps 3, 4, 10 emit paste-ready recipes; the RM signs and runs `svn commit` as themselves.
- Agent never publishes. Steps 10 (`svn mv dist/dev → dist/release`) and 11 (`[ANNOUNCE]` send, site bump merge) are the moments of release; the human commit is the act of release.

Mirrors [`security-cve-allocate`](.claude/skills/security-cve-allocate/SKILL.md) (Vulnogram URL + paste-ready JSON, human submits) and satisfies [RFC-AI-0004 Principle 1](docs/rfcs/RFC-AI-0004.md).

### What lands

- [`docs/release-management/README.md`](docs/release-management/README.md), family overview, skill table, state-change boundaries
- [`docs/release-management/process.md`](docs/release-management/process.md), 14-step lifecycle with Mermaid flowchart, per-step description, label state diagram, Adopter backends table
- [`docs/release-management/spec.md`](docs/release-management/spec.md), per-skill scope, state-change boundary, hand-off protocol, adopter contract
- [`projects/_template/release-management-config.md`](projects/_template/release-management-config.md), family-wide adopter contract with the 3 backend switches
- [`projects/_template/release-build.md`](projects/_template/release-build.md), build invocation, digest set, binary-exclude rules
- [`projects/_template/pmc-roster.md`](projects/_template/pmc-roster.md), roster used by `release-vote-tally` to classify binding vs non-binding
- [`projects/_template/site-repo.md`](projects/_template/site-repo.md), site-bump PR target for `release-announce-draft`
- Wires the family into [`README.md`](README.md) (root) and [`docs/modes.md`](docs/modes.md) under Triage + Drafting subsections, each row marked `proposed`

Operationalises the [`MISSION.md` § Initial Goals](MISSION.md) commitment to *cut a first Apache release through the standard process within 3 months of resolution adoption*, with non-ASF adopters served from day one.

### Status

**Proposed.** No `release-*` skill code in this PR. Promotion of any skill from `experimental` to default-on, or from Drafting to a state-changing lane, requires evidence sourced from Release Managers and binding voters that the project's release process is healthier (fewer stalled RCs, shorter time-to-`[ANNOUNCE]`, fewer reverted promotions), not throughput numbers alone.

## Test plan

- [x] `prek run --all-files` green (markdownlint, typos, doctoc, end-of-file, trailing whitespace)
- [x] `lychee --offline` green across the 7 new + 2 modified files
- [x] doctoc TOCs reflect current H2/H3 structure on all 4 docs files + 4 scaffolds
- [x] Cross-references between [`process.md`](docs/release-management/process.md), [`spec.md`](docs/release-management/spec.md), [`README.md`](docs/release-management/README.md), and [`projects/_template/release-management-config.md`](projects/_template/release-management-config.md) resolve
- [x] [`docs/modes.md`](docs/modes.md) rows match the skill table in [`README.md`](docs/release-management/README.md)
- [x] Adopter contract required-keys list in [`spec.md`](docs/release-management/spec.md) matches the keys actually defined in [`projects/_template/release-management-config.md`](projects/_template/release-management-config.md)
- [x] Non-ASF backend switches in [`process.md` § Adopter backends](docs/release-management/process.md) align with the switches block in [`release-management-config.md` § Backends](projects/_template/release-management-config.md)
- [x] ASF state-change boundaries (key, publication, mail, labels) consistent between [`spec.md`](docs/release-management/spec.md) and the Drafting-row purposes in [`README.md`](docs/release-management/README.md)

### DIFF
--- a/docs/release-management/README.md
+++ b/docs/release-management/README.md
@@ -1,0 +1,201 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [Release-management skill family](#release-management-skill-family)
+  - [Status](#status)
+  - [Skills](#skills)
+  - [Deep documentation](#deep-documentation)
+  - [Adopter contract](#adopter-contract)
+  - [Mode mapping](#mode-mapping)
+  - [Cross-references](#cross-references)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<!-- SPDX-License-Identifier: Apache-2.0
+     https://www.apache.org/licenses/LICENSE-2.0 -->
+
+# Release-management skill family
+
+End-to-end automation for an ASF project's release lifecycle,
+from the planning issue and version bump through to `[ANNOUNCE]`
+on `announce@apache.org`, archive sweep, and the per-release
+audit log. Ten skills that compose into the canonical 14-step
+process documented in [`process.md`](process.md).
+
+Why a framework skill family? Every ASF project runs essentially
+the same release process: version bump, `KEYS` reconciliation, RC
+signed with the Release Manager's key, staged to
+`dist/apache.org/repos/dist/dev/<project>/`, voted on `dev@` per
+[`release-policy.html § release approval`](https://www.apache.org/legal/release-policy.html#release-approval),
+promoted to `dist/release/<project>/`, announced on
+`announce@apache.org` per
+[`release-policy.html § announcements`](https://www.apache.org/legal/release-policy.html#release-announcements),
+and archived past retention per
+[`release-distribution`](https://infra.apache.org/release-distribution.html).
+The procedural shape is foundation-wide; the project-specific
+content (release-train identity, build invocation, `KEYS` file
+path, vote-window length, retention rule, audit-log location)
+plugs in through [`<project-config>/`](../../projects/_template/)
+just like the security family.
+
+Non-ASF adopters are first-class adopters of this family, not a
+follow-up case. The 14-step lifecycle is described in ASF
+terminology because the framework's first pilot is an ASF PMC,
+but every step that touches an ASF-specific surface is implemented
+as a backend call the adopter selects in
+[`release-management-config.md`](../../projects/_template/release-management-config.md).
+Three dimensions parametrise the lifecycle, with no ASF assumption
+baked into the install path:
+
+- **Distribution backend** (`release_dist_backend`): `svnpubsub`
+  (ASF), `github-releases`, `s3`, `self-hosted`.
+- **Approval mechanism** (`release_approval_mechanism`):
+  `dev-list-vote` (ASF), `github-discussion`, `pr-approval`,
+  `maintainer-roster`.
+- **Announcement backend** (`release_announce_backend`):
+  `announce-list` (ASF), `github-release-notes`, `site-post`,
+  `discord-channel`.
+
+The 14 steps stay identical across backends; only the command set
+the agent emits changes. The state-change boundaries (Drafting vs
+Triage; agent never holds the signing key; agent never publishes)
+stay identical too. See
+[`process.md` § Adopter backends](process.md#adopter-backends)
+for the full backend table and per-step mapping.
+
+## Status
+
+**Proposed.** No `release-*` skill code exists in the framework
+today. This family lands as docs first (this README, the 14-step
+[`process.md`](process.md), the per-skill [`spec.md`](spec.md), and
+the adopter scaffold
+[`projects/_template/release-management-config.md`](../../projects/_template/release-management-config.md))
+so the lifecycle, the state-change boundaries, and the adopter
+contract are reviewable independently from runtime behaviour.
+The skills follow in subsequent PRs, each shipped flagged
+`experimental` and tracked in [`docs/modes.md`](../modes.md). This
+pattern matches [Mentoring](../mentoring/README.md).
+
+Promotion of any skill in this family from `experimental` to
+default-on, or from Drafting to a state-changing lane, requires
+evidence sourced from Release Managers and binding voters that
+the project's release process is healthier (fewer stalled
+RCs, shorter time-to-`[ANNOUNCE]`, fewer reverted promotions),
+not throughput numbers alone. The evidence window is set by
+adopter governance, not by this family.
+
+See [`MISSION.md` § Initial Goals](../../MISSION.md#initial-goals)
+for the commitment to *cut a first Apache release through the
+standard process within 3 months of resolution adoption*; this
+family operationalises it.
+
+## Skills
+
+The skill table below names each `release-*` skill, its mode, and
+the lifecycle step(s) it owns. Read [`spec.md`](spec.md) for the
+per-skill state-change boundary; read [`process.md`](process.md)
+for the step it executes against.
+
+| Skill | Mode | Steps owned | Purpose |
+|---|---|---|---|
+| `release-prepare` | Drafting | 1, 2, 14 | Open the planning issue, draft the version-bump + changelog + NOTICE/LICENSE PR, then draft the post-release `-SNAPSHOT` bump. |
+| `release-keys-sync` | Drafting | 3 | Draft the `KEYS` diff for a Release Manager cutting their first release for the project. Agent never holds the private key. |
+| `release-rc-cut` | Drafting | 4, 5 | Emit the paste-ready command sequence, signed tag, build, detached signatures, checksums, `svn` import to `dist/dev/<project>/`. Agent never signs and never imports. |
+| `release-verify-rc` | Triage / Pairing | 6 | Read-only pre-flight: signatures against the project's `KEYS`, checksums, license headers (Apache RAT), NOTICE/LICENSE presence, no prohibited binaries, version-string consistency. Voters can run it in their own dev loop before posting `+1`. |
+| `release-vote-draft` | Drafting | 7 | Draft the `[VOTE]` email body to `dev@<project>`. Agent never sends. |
+| `release-vote-tally` | Triage | 9 | Parse the vote thread, classify each reply (+1 / 0 / -1) binding vs non-binding against the PMC roster, propose `[RESULT] [VOTE]`. Conservative on ambiguous votes, refuses to count, flags `AMBIGUOUS, needs RM call`. |
+| `release-promote` | Drafting | 10 | Emit the paste-ready `svn mv dist/dev → dist/release` command set plus commit message. Agent never moves; the human commit is the act of release. |
+| `release-announce-draft` | Drafting | 11 | Draft the `[ANNOUNCE]` email body to `announce@apache.org` and the site-bump PR (download page, release notes, version banner). Agent never sends mail and never merges the site PR. |
+| `release-archive-sweep` | Triage | 12 | Scan `dist/release/<project>/`, identify releases past retention, propose the `svn mv` sequence to `archive.apache.org`. Agent never moves. |
+| `release-audit-report` | Triage (dashboard) | 13 | Read-only structured report per release, RM, voters with binding flags, artefacts with sigs and checksums, promotion revision, `[ANNOUNCE]` archive URL. Output appended to the project's audit log. |
+
+Two non-negotiable boundaries cross every Drafting skill above:
+
+- **The agent never holds, invokes, or proxies the Release
+  Manager's private signing key.** Steps 3, 4, 10 emit paste-ready
+  recipes; the RM runs every signing or `svn commit` operation as
+  themselves. This mirrors
+  [`security-cve-allocate`](../../.claude/skills/security-cve-allocate/SKILL.md)
+  (Vulnogram URL + paste-ready JSON, human submits) and satisfies
+  [RFC-AI-0004 Principle 1](../rfcs/RFC-AI-0004.md#principle-1--human-in-the-loop-on-every-state-change).
+- **The agent never publishes the release.** Steps 10
+  (`svn mv dist/dev → dist/release`) and 11 (`[ANNOUNCE]` send,
+  site bump merge) are the moments of release; the agent drafts
+  artefacts, the RM and the PMC execute and merge.
+
+## Deep documentation
+
+- [**`process.md`**](process.md), the 14-step lifecycle with
+  Mermaid flowchart + per-step description; the label-lifecycle
+  state diagram + label reference table. The authoritative
+  process reference.
+- [**`spec.md`**](spec.md), per-skill scope, state-change
+  boundary, hand-off protocol, adopter knobs. The contract the
+  future skill implementations must satisfy.
+
+Two documents that **do not** ship in this family but are
+referenced from it:
+
+- [`<project-config>/release-trains.md`](../../projects/_template/release-trains.md)
+ , release-train identity (already present in the adopter
+  scaffold for security use; release-management reuses it).
+- [`<project-config>/release-management-config.md`](../../projects/_template/release-management-config.md)
+ , the family's adopter contract (new in this PR).
+
+## Adopter contract
+
+The skills resolve project-specific content from the release-
+workflow files in
+[`<project-config>/`](../../projects/_template/), see the
+adopter scaffold's
+[`README.md`](../../projects/_template/README.md) for the
+file-by-file index. Required at minimum:
+
+- `project.md`, identity, repos, mailing lists, tools
+- `release-management-config.md`, vote window, vote-pass rule,
+  signing-key requirements, audit-log location, retention rule,
+  build command, KEYS file path
+- `release-trains.md`, release-train identity (shared with the
+  security family)
+- `release-build.md`, build invocation, digest set, binary-exclude
+  list (new in this PR; minimal scaffold)
+- `pmc-roster.md`, PMC member roster used by
+  `release-vote-tally`
+  to classify binding vs non-binding votes (new in this PR;
+  minimal scaffold)
+- `site-repo.md`, site-bump PR target for
+  `release-announce-draft`
+  (new in this PR; minimal scaffold)
+
+## Mode mapping
+
+Release-management is a **family**, not a mode. The lifecycle
+spans the existing Triage and Drafting modes; no new mode is
+introduced. See [`docs/modes.md`](../modes.md) for the
+family-by-mode breakdown, `release-*` skills appear under the
+**Triage** and **Drafting** subsections, each marked `proposed`.
+
+The family's read-only dashboard skill
+(`release-audit-report`)
+sits in Triage because it classifies and reports against existing
+state, not because it routes inbound work. The
+[`docs/modes.md` § Outside the modes](../modes.md#outside-the-modes)
+section is reserved for framework infrastructure (`setup-*`,
+`issue-reassess-stats`, isolation tooling) and the audit skill
+does not belong there, it is project-facing maintainership work.
+
+## Cross-references
+
+- [`process.md`](process.md), [`spec.md`](spec.md), within this
+  family.
+- [`MISSION.md` § Initial Goals](../../MISSION.md#initial-goals),
+  the standard-release commitment this family operationalises.
+- [`docs/modes.md`](../modes.md), the mode taxonomy each skill
+  in this family inhabits.
+- [`docs/security/README.md`](../security/README.md), the
+  precedent for a multi-skill ASF-process family with shared
+  state-change-boundary discipline.
+- [`docs/mentoring/README.md`](../mentoring/README.md), the
+  precedent for spec-before-code on a proposed family.
+- [ASF release policy](https://www.apache.org/legal/release-policy.html), [ASF release distribution](https://infra.apache.org/release-distribution.html), [ASF release signing](https://infra.apache.org/release-signing.html), [ASF licensing-howto](https://www.apache.org/legal/resolved.html), [Apache RAT](https://creadur.apache.org/rat/), the canonical foundation references the lifecycle is anchored to.
--- a/docs/release-management/process.md
+++ b/docs/release-management/process.md
@@ -1,0 +1,555 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [Release-management workflow: process and label lifecycle](#release-management-workflow-process-and-label-lifecycle)
+  - [Adopter backends](#adopter-backends)
+  - [Process reference: the 14 steps](#process-reference-the-14-steps)
+    - [Step 1: Release planning + version bump](#step-1-release-planning--version-bump)
+    - [Step 2: Changelog, NOTICE, LICENSE](#step-2-changelog-notice-license)
+    - [Step 3: `KEYS` reconciliation](#step-3-keys-reconciliation)
+    - [Step 4: Cut the release candidate](#step-4-cut-the-release-candidate)
+    - [Step 5: Stage to `dist/dev/`](#step-5-stage-to-distdev)
+    - [Step 6: Pre-flight RC verification](#step-6-pre-flight-rc-verification)
+    - [Step 7: `[VOTE]` thread on `dev@`](#step-7-vote-thread-on-dev)
+    - [Step 8: Voting window](#step-8-voting-window)
+    - [Step 9: Tally + `[RESULT] [VOTE]`](#step-9-tally--result-vote)
+    - [Step 10: Promote `dist/dev/` to `dist/release/`](#step-10-promote-distdev-to-distrelease)
+    - [Step 11: `[ANNOUNCE]` + site bump](#step-11-announce--site-bump)
+    - [Step 12: Archive sweep](#step-12-archive-sweep)
+    - [Step 13: Audit log](#step-13-audit-log)
+    - [Step 14: Post-release version bump](#step-14-post-release-version-bump)
+  - [Label lifecycle](#label-lifecycle)
+    - [State diagram](#state-diagram)
+    - [Label reference](#label-reference)
+  - [Cross-references](#cross-references)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<!-- SPDX-License-Identifier: Apache-2.0
+     https://www.apache.org/licenses/LICENSE-2.0 -->
+
+# Release-management workflow: process and label lifecycle
+
+The authoritative reference for the 14-step release lifecycle and
+the label-lifecycle state diagram the [release-management
+skills](../../.claude/skills/) execute against. The
+[family README](README.md) lists the skills; this document is the
+process they share. The [spec](spec.md) defines per-skill scope,
+state-change boundary, and adopter knobs.
+
+The lifecycle is described in **ASF terminology by default**
+(svnpubsub on `dist.apache.org`, `[VOTE]` on `dev@`, `[ANNOUNCE]`
+on `announce@apache.org`), because the framework's first pilots
+include an ASF PMC release. Every step that touches an ASF-specific
+surface is implemented as a *backend call* the adopter selects in
+[`release-management-config.md`](../../projects/_template/release-management-config.md),
+not a hard-coded operation. Non-ASF adopters resolve the same
+abstract step to their own backend; see the
+[Adopter backends](#adopter-backends) section for the dimensions
+and the per-step backend mapping.
+
+## Adopter backends
+
+Three dimensions parametrise the lifecycle. Each adopter picks one
+value per dimension in `release-management-config.md`. The 14
+steps stay identical; only the backend the agent emits commands
+against changes.
+
+| Dimension | Config key | ASF default | Non-ASF examples |
+|---|---|---|---|
+| Distribution backend | `release_dist_backend` | `svnpubsub` (`svn import` to `dist/dev/`, `svn mv` to `dist/release/`) | `github-releases` (`gh release upload` / `gh release edit --draft=false`), `s3` (`aws s3 cp` / `aws s3 mv`), `self-hosted` (project-supplied command template) |
+| Approval mechanism | `release_approval_mechanism` | `dev-list-vote` (`[VOTE]` thread on `dev@<project>.apache.org`, 72h window, 3 binding +1, more +1 than -1) | `github-discussion` (named Discussion thread on `<upstream>` repo), `pr-approval` (a "release-NN" PR with approvals from the configured roster), `maintainer-roster` (signed approvals from a named roster file) |
+| Announcement backend | `release_announce_backend` | `announce-list` (mail to `announce@apache.org`, cc `dev@`, `users@`) | `github-release-notes` (the release-page body is the announcement), `site-post` (a blog post in `site_repo`), `discord-channel` (a webhook into a named channel) |
+
+Each `release-*` skill consults the relevant key and emits
+backend-shaped paste-ready commands. The state-change boundaries
+([spec § Cross-cutting commitments](spec.md#cross-cutting-commitments))
+do not change: the agent still emits a recipe; the human still
+runs it. Swapping `svn` for `gh release upload` swaps the
+backend, not the boundary.
+
+The vote-tally roster (`release-vote-tally`) reads from the
+adopter's `pmc-roster.md` for ASF projects, or
+`<release_approver_roster_path>` (typically
+`<project-config>/release-approvers.md`) for non-ASF adopters; both
+files share the same schema (handle, binding-flag, optional GPG
+fingerprint).
+
+> [!IMPORTANT]
+> Release Management is **proposed** in the framework today. No
+> `release-*` skill code exists yet. This document, the family
+> [`README.md`](README.md), the [`spec.md`](spec.md), and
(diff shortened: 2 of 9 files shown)


choo121600 (commented):
Wow...! this is huge 😄
Overall this looks really good to me.

I’d especially love to hear feedback from people who have deep ASF release-management experience, since this touches a lot of important release-process and policy boundaries.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+release train it belongs to (see
+[`<project-config>/release-trains.md`](../../projects/_template/release-trains.md)),
+the cut-off commit, and the issues / PRs in scope. The skill drafts
+that planning issue from the configured release-train metadata, then
+drafts the version-bump PR (e.g. `pom.xml`, `pyproject.toml`,
+`Cargo.toml`, `setup.py`, package manifests). The PR remains in
+draft until the RM marks it ready; the agent never marks ready, never
+merges.
+
+For non-ASF adopters with no release-train concept the planning step
```
potiuk, line 177: I would not even mention "non-ASF" - I woudl rather say that we should explicitly mention here that in various projects this step can have differnt shapes. For example some projects (like Airflow) might release from release branches and version might be already set there, or version might be dynamically generated based on tag name etc. .. So even in ASF this step might be different - I think it might be worth mentioning here that it's even likely and that it often needs an override or maybe better - it needs a pointer to existing release documentation that can be used as strong reference.
Possibly we should add - in the configuration - a way for the project to tell where their "local" relase documentation is, rather than using the overrride mechanism, because this is something that we anyhow want to make "manually" followable, not only agentically.
This should not be really an "override" but more clarification of particular implementation steps, and we could even add a check here to see if such existing release docuementation is complete and has all the boxes checked.

I would even go as far as to say that we should "import" all the configuration from that exising release documentation at adoption time - and keep it in sync while doing the release.

The release process should also have expectation that agent stores the changes / isues/ new things during the release and proposes to RM to create a PR After release is done that will fix both - release documentation and SKILL overrides / configuration in their repo.

potiuk:
On it.. That one looks really cool !

docs/release-management/README.md
```diff
@@ -0,0 +1,201 @@
...
+| `release-promote` | Drafting | 10 | Emit the paste-ready `svn mv dist/dev → dist/release` command set plus commit message. Agent never moves; the human commit is the act of release. |
+| `release-announce-draft` | Drafting | 11 | Draft the `[ANNOUNCE]` email body to `announce@apache.org` and the site-bump PR (download page, release notes, version banner). Agent never sends mail and never merges the site PR. |
+| `release-archive-sweep` | Triage | 12 | Scan `dist/release/<project>/`, identify releases past retention, propose the `svn mv` sequence to `archive.apache.org`. Agent never moves. |
+| `release-audit-report` | Triage (dashboard) | 13 | Read-only structured report per release, RM, voters with binding flags, artefacts with sigs and checksums, promotion revision, `[ANNOUNCE]` archive URL. Output appended to the project's audit log. |
+
+Two non-negotiable boundaries cross every Drafting skill above:
+
+- **The agent never holds, invokes, or proxies the Release
```
potiuk, line 115: One comment here. PRACTICALLY speaking, the "isolation" framework already prevents those credentials to leak.

Currently it does "svn commit" -> isolated environment does not have credentials -> it detects that it needs to get out of sandbox, proposes it to the RM -> RM allows to run **this particular command** outside of sandbox.

I think we should show it as a viable alternative/ way of achieving it, because I think it maintains all the expectations of "never share your credentials with agent".

potiuk (approved):
Fantastic! I reviewed it in detail and I have two comments only - to better integrate with existing manual release processes and handle the fact that most ASF PMCs already have good and followable release documentation - I think specific project configuration and overrides should be "synced" with such manual docuementation - we do not want to "completely" replace release process with agentic flows, there needs to be a manual one, that will be easier to follow without steward. Some release managers might not want to use agents, and might choose to follow manual path - but if we actually encode the rule that we also have to have - pure manually followed way of release - this will 100% alleviate those concerns. And actually - very funny - it will allow those release managers who will use agentic workflows to **also** fix and update manual documentation at the same time. 

The second one - I think it's worth mentioning that isolated setup already handles the case with not sharing credentials with the agent. This might be not obvious, and maybe we would like to add some "additional" gates aroudn it - but referring to the isolated setup and explaining how it plays out here to fulfilll the expectations might be a good idea.

potiuk:
Also turning to draft - I think it needs more pairs of eyes and reviews - I will ask in the main thread as things that we have now needing more attention than "quick fixes to existing PRs"

andreahlert:
Both points land. Manual-followable path as a first-class invariant is the right framing — encoding the manual doc as the source of truth (not the steward config) flips the whole adoption story: the agent becomes an assistant to the doc, not a replacement for it.

I'll lift this into `spec.md` as a cross-cutting commitment alongside the existing "agent never holds the key" / "agent never publishes" pair, plus a skill that diffs the manual doc against what actually happened during the RC and proposes the post-release reconciliation PR (folding into `release-audit-report` rather than adding an 11th skill).

Replied to the isolation point in the inline thread below.

docs/release-management/process.md
andreahlert: Good catch — the ASF / non-ASF split is the wrong axis. Airflow's release-branch + dynamic-tag flow is already a counterexample inside ASF, and Beam, Arrow, Iceberg all diverge here too.

I'll drop the "non-ASF adopters" wording from Step 1 and replace it with a "project-shaped step" framing, driven by a new `release_manual_doc_path` config key pointing at the project's canonical release doc — treated as an authoritative reference, not an override. The adoption flow reads that doc and populates config from it; `release-audit-report` tracks drift during the RC and feeds the post-release reconciliation PR you described. I'll also add a coverage check (does the doc cover all 14 steps).

docs/release-management/README.md
andreahlert: Right — the boundary is already structurally enforced by the sandbox; the doc treats it as policy when it's actually mechanism.

I'll rewrite the two bullets to name the isolation framework as the implementation: the agent runs without credentials, hits `svn commit`, the sandbox detects the escape, the RM authorises that specific command outside the sandbox, and the agent never sees the key material. The policy statement stays above it as the invariant the mechanism enforces, so adopters who plug in a different sandbox still have to satisfy the same boundary. I'll cross-link to [`secure-agent-setup.md` § sandbox-bypass-visibility-hook](https://github.com/apache/airflow-steward/blob/main/docs/setup/secure-agent-setup.md#sandbox-bypass-visibility-hook), which is the per-call approval gate this relies on.

projects/_template/release-management-config.md
```diff
@@ -0,0 +1,219 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [Apache Airflow: release-management configuration](#apache-airflow-release-management-configuration)
```
justinmclean: This could cause issues. Either use <Project Name> placeholders consistently (like the other template files do), or make the Airflow-specific nature clear with a note that this is a filled example, not a blank template.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+    S7 --> S8
+    S8 --> S9
+    S9 -->|pass| S10
+    S9 -->|fail| FAIL
+    FAIL --> S4
+    S10 --> S11
+    S11 --> S12
+    S11 --> S14
+    S12 --> S13
+    S14 --> S13
```
justinmclean: Steps 12, 13, and 14 are described as post-promote follow-ups. The flowchart has both S11 → S12 and S11 → S14, implying 12 and 14 are parallel paths. But the narrative says Step 14 (post-release snapshot bump) is logically after Step 11, while Step 12 (archive sweep) is separate. Worth clarifying with a note or re-examining whether 12→13 and 14→13 should be separate terminal paths.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+draft until the RM marks it ready; the agent never marks ready, never
+merges.
+
+For non-ASF adopters with no release-train concept the planning step
+collapses to a tag-and-PR pair; the skill detects the absence of
+[`<project-config>/release-trains.md`](../../projects/_template/release-trains.md)
+and adapts.
+
+### Step 2: Changelog, NOTICE, LICENSE
+
```
justinmclean, line 183: The policy requirements are correctly captured. LICENSE at the root of the distribution with full ALv2 text, NOTICE conforming to section 4(d), license headers on all source files. The Category-X dependency denylist maps correctly to [resolved licenses policy](https://www.apache.org/legal/resolved.html).

However, the Release Policy says: "an audit SHOULD be performed before any full release is created" — this is the RAT check. The PR assigns RAT to Step 6 (release-verify-rc), not Step 2. That's fine operationally (RAT runs on the built artefact, not the source tree in isolation), but Step 2 should note that the LICENSE/NOTICE prep here must account for exactly the bundled content, not anticipated content,  the policy is explicit that LICENSE and NOTICE "MUST NOT provide information about materials which are not bundled in the package." If a dependency is pulled in at build time but not known at Step 2, the Step 2 draft could be wrong. The spec doesn't mention this risk.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+license-categorised dependency changed
+([ASF Category-A/B/X policy](https://www.apache.org/legal/resolved.html)).
+The skill flags any Category-X dependency for the RM to remove before
+the cut and refuses to advance the planning issue until that flag
+clears.
+
+Output: a single PR proposed against the release branch, RM merges
+after their own review.
+
+### Step 3: `KEYS` reconciliation
```
justinmclean, line 209: The release-keys-sync spec verifies that the RM's public key appears in KEYS, that's correct. But the Release Distribution Policy adds requirements not reflected in the skill spec:

Key size (MUST): "Signing keys for new artifacts MUST be RSA and at least 2048 bit. New keys SHOULD be 4096 bit RSA." The release-keys-sync spec never validates key size. An RM adding a 1024-bit legacy key would pass the skill's check.

Keyserver publication (SHOULD): "Signing keys used at Apache MUST be published in the KEYS file and SHOULD be made available through the global public keyserver network. Signing keys SHOULD be linked into a strong web of trust." The skill spec doesn't draft a step to upload the key to a public keyserver, only to update the KEYS file.

No signing on ASF machines (MUST): "Private keys MUST NOT be stored on any ASF machine. Likewise, signatures for releases MUST NOT be created on ASF machines." The PR's Boundary 1 correctly prohibits the agent from holding the key, but it doesn't communicate to the RM the policy requirement that signing must occur on personally controlled hardware.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+that the key must also be uploaded to a public keyserver per
+[the ASF release-signing FAQ](https://infra.apache.org/release-signing.html).
+The agent never holds the private key and never runs `svn commit`;
+the RM executes both as themselves.
+
+If the RM's key is already present and unchanged, the skill is a
+no-op and reports so on the planning issue.
+
+### Step 4: Cut the release candidate
+
```
justinmclean, line 229: The Release Distribution Policy is specific: "MUST supply at least one checksum file. SHOULD supply a SHA-256 and/or SHA-512 checksum file. SHOULD NOT supply a MD5 or SHA-1 checksum file because these are deprecated. For new releases, PMCs MUST supply SHA-256 and/or SHA-512."

The release-build.md template gets this right (SHA-512 required, SHA-256 optional), but the release-rc-cut spec says "checksums" generically without naming the required/prohibited digest types. A skill implementer reading only spec.md wouldn't know MD5/SHA-1 are prohibited for new releases.

Separately, the policy also requires that .sig binary PGP files MUST NOT be provided, only .asc (ASCII-armored). The spec doesn't call this out.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+Step 13.
+
+> [!NOTE]
+> Detached `.asc` signatures and `.sha512` checksums are the ASF
+> baseline; some projects also publish `.sha256` for older
+> downstream tools. The skill reads
+> [`<project-config>/release-build.md`](../../projects/_template/release-build.md)
+> to determine which digests apply.
+
+### Step 5: Stage to `dist/dev/`
```
justinmclean, line 261: The Release Policy says: "Before casting +1 binding votes, individuals are REQUIRED to download all signed source code packages onto their own hardware, verify that they meet all requirements of ASF policy on releases as described below, validate all cryptographic signatures, compile as provided, and test the result on their own platform."

The PR frames release-verify-rc as a Triage skill that "doubles as a Pairing-mode skill voters run in their own dev loop." The framing is fine, but the spec never states that the skill output does not substitute for the voter's personal verification obligation.

The policy explicitly requires voters to download, compile, and test on their own hardware — a pass from release-verify-rc cannot discharge that requirement. The skill spec should carry a disclaimer to this effect so voters don't treat a clean report as a +1 without doing the hands-on work.

Also: "Must releases be built on hardware owned and controlled by the committer?" — the policy says yes, the release should be verified on hardware the committer has physical possession of. The skill spec doesn't echo this.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+does not auto-flip any label; the RM decides whether to roll a new
+RC (back to Step 4) or fix the verification source itself (e.g.
+update RAT excludes).
+
+This skill is the most valuable Pairing-mode candidate in the family:
+voters run it in their own dev loop before posting a `+1` so the
+mechanical verification work is done before the human-to-human vote
+conversation.
+
+### Step 7: `[VOTE]` thread on `dev@`
```
justinmclean, line 314: Expedited release policy is missing.

The 72-hour floor is correctly captured. But the Expedited Releases section of the policy adds two requirements that the PR ignores:

"Emails calling for a Release Vote that run for less than 72 hours MUST include an explanation of why the release is being expedited." and "deviations from normal policy MUST be reported to the Board" in the next scheduled board report.

The release-vote-draft skill has no path for expedited releases, and the spec doesn't mention these obligations. A security-fix RC might need an expedited vote; the skill should support drafting the explanation and flagging the Board-reporting requirement.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+(non-binding).
+**Skill:** none, passive window.
+
+No skill runs during the window. The planning issue carries the
+`vote-open` label so other skills (Triage on adjacent issues,
+Mentoring on the `[VOTE]` thread) can spot the window and behave
+appropriately (e.g. Mentoring stays out of the `[VOTE]` thread per
+its hand-off rules, vote discussion is PMC business).
+
+### Step 9: Tally + `[RESULT] [VOTE]`
```
justinmclean, line 344: No implicit RM vote. The Voting Policy is explicit: "Note that there is no implicit +1 from the release manager, or from anyone in any ASF vote. Only explicit votes are valid." If the RM doesn't post a +1, they don't have one. The release-vote-tally spec doesn't mention this; a tally that assumes the RM cast a +1 because they initiated the vote would be wrong.

Fractional votes are non-binding on releases. The Voting Policy states: "Votes from PMC Members on releases must use +1, 0, -1 to be considered binding." A PMC member posting +0.9 on a release vote is non-binding, even though they're a PMC member. The tally spec talks about binding/non-binding classification against the PMC roster but doesn't mention this fractional-vote rule. A +0.9 from a PMC member should be classified as non-binding and surfaced in the report.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+baseline (three binding `+1` minimum, more binding `+1` than `-1`);
+the configured per-project rule in
+[`<project-config>/release-management-config.md`](../../projects/_template/release-management-config.md)
+overrides if the project demands stricter conditions.
+
+On fail the planning issue gets the `rc-rolled` label, the RM
+returns to Step 4 with a new RC number. The skill never closes the
+planning issue on a failed vote.
+
+### Step 10: Promote `dist/dev/` to `dist/release/`
```
justinmclean, line 374: The Release Policy and Release Distribution Policy both say: "by default only PMC members can update the dist/release directory tree. If the Release Manager is not a member of the PMC, they will need to ask a PMC member to do the actual release publication."

The release-promote skill emits a paste-ready svn mv command for the RM. If the RM is a committer but not a PMC member, running that command will fail. The skill spec doesn't check or mention this constraint.

The Release Policy also says "Wait about an hour after uploading a new release before updating the project download page" — the release-promote skill should tell the RM to wait ~1 hour before triggering the Step 11 announcement.

docs/release-management/process.md
```diff
@@ -0,0 +1,521 @@
...
+This is **the moment of release**. The skill writes nothing and
+runs nothing; the human commit is the act.
+
+> [!NOTE]
+> Per [release-distribution](https://infra.apache.org/release-distribution.html),
+> mirrors propagate within ~24h. The skill writes the expected
+> mirror-availability time to the planning issue so Step 11's
+> `[ANNOUNCE]` can wait if appropriate.
+
+### Step 11: `[ANNOUNCE]` + site bump
```
justinmclean, line 407: Three gaps.

1-hour wait (policy requirement). "Please ensure that you wait at least one hour after uploading a new release before updating the project download page and sending the announcement email(s)." 

The release-announce-draft spec has no mention of this timing constraint. The skill should include a reminder that the announcement must not go out until at least 1 hour after the svn mv in Step 10.

@apache.org address required. "Please note that you can not post the ASF-wide announcement list without using an 'apache.org' mail address." The skill drafts the announcement body but doesn't flag this constraint. An RM posting from a Gmail address will be rejected by the list.

Download page must use closer.lua (MUST). The Release Download Pages Policy requires download pages to use closer.lua, not direct links to dist.apache.org. The site-repo.md scaffold and release-announce-draft spec don't mention this. A site-bump PR that adds direct dist.apache.org links to the download page violates policy. The skill should either generate closer.lua links or flag that the download page update must comply with the download pages policy.

Additionally, the Release Policy says: "Announcements must contain a link to the Project Download Page for the source." — this is a MUST that the skill should enforce in the drafted body.

justinmclean:
I validated this skill against the recent ASF Policy MCP I created, so most of my comments come from that, but they also match with my knowledge of releases.

justinmclean:
You may also want to consider adding a step to publish to other distribution areas and validate them; that's an area a lot of projects get wrong, but there's no need for that to be in the first release of this.

projects/_template/release-management-config.md
andreahlert: the title was inconsistent with the other three template files, all "TODO: <Project Name>: ...". the body already carried the disclaimer ("replace every Airflow-specific value") so that part was fine, just the title and TOC didn't reflect it. flagged it as a filled example.

docs/release-management/process.md
andreahlert: the parallel bit is already stated, Step 14's text says "runs in parallel with Step 12, the two steps share no state", and the flowchart matches, S11 forks to both and they join at S13. one terminal, not separate paths. Sorry, could you explain in other words, or could you help me to understand what is incorrect.

docs/release-management/process.md
andreahlert: good catch. Step 2 drafts LICENSE/NOTICE before the build, so it can't know a build-time-only dependency. marked the Step 2 output provisional and made explicit that RAT at Step 6 verifies against the actual built artefact, not the anticipated set.

docs/release-management/process.md
andreahlert: key size was a real gap, release-keys-sync only checked the key was present in KEYS, not that it's RSA >= 2048. added the floor check plus a hand-off if the key is weaker. the keyserver point was already covered, Step 3 and the skill scope both tell the RM to upload. also added the "key must not be used on ASF infra" MUST to Boundary 1, it only said "their own machine" before. Wdyt?

docs/release-management/process.md
andreahlert: The spec already named the digests, release-rc-cut inputs say "sha512, optionally sha256" and the Step 4 note spells out the baseline. what was missing was the prohibition side. added it in both Step 4 and the rc-cut spec, MD5/SHA-1 not allowed for new releases, .asc-only no binary .sig.

docs/release-management/process.md
andreahlert: Thank you, my mistake! Agreed, important one. a voter treating a clean release-verify-rc report as their +1 is exactly the failure the framework is built against. added a disclaimer to the spec, the report doesn't discharge the voter's obligation to download, compile, and test on their own hardware.

docs/release-management/process.md
andreahlert: no expedited path before, added one to release-vote-draft. a <72h window now needs the explanation text in the [VOTE] body and the skill flags the obligation to report the deviation in the next board report.

docs/release-management/process.md
andreahlert: the fractional-vote point was right, the spec lumped +0.9 into AMBIGUOUS, but imho a fractional vote isn't ambiguous, it's determinately non-binding. split it out, fractional gets classified non-binding directly, AMBIGUOUS stays for genuinely unclear cases. also added an explicit note that the tally never invents an implicit RM +1. WDYT?

docs/release-management/process.md
andreahlert: Thanks,both valid. release-promote assumed the RM could run the svn mv, but dist/release is PMC-write-only, a committer-not-PMC RM would fail. added a PMC roster check and an "ask a PMC member to publish" hand-off. also added the ~1h wait before the download-page update. Could you re-check if it's ok?

docs/release-management/process.md
andreahlert: Done: added to release-announce-draft and the process step: the 1h wait before [ANNOUNCE], the @apache.org sender requirement, closer.lua links on the download page (not direct dist.apache.org), and a mandatory Download Page link in the announcement body. Thanks for catching.

andreahlert:
> I validated this skill against the recent ASF Policy MCP I created, so most of my comments come from that, but they also match with my knowledge of releases.

cross-checked the MUST-level items against release-policy.html while applying these. it confirms the @apache.org sender, the 1h wait, dist/release being PMC-write by default, and the expedited explanation + board report. the MD5/SHA-1 prohibition isn't on that page, it's in release-distribution.html, and the doc cites that one for it. so the citations point at the canonical source now.

I am not that expert level 😄 , so it is based on my interpretation level. Please, feel free to advise me more.

andreahlert:
> You may also want to consider adding a step to publish to other distribution areas and validate them; that's an area a lot of projects get wrong, but there's no need for that to be in the first release of this.

good idea, agreed it's out of scope for this first docs PR. added it as a deferred open question in spec.md so it's not lost, and we can open an issue as soon as this PR passes.

potiuk:
Hi @andreahlert — heads-up first: `main` was just refreshed with pinned-action SHA bumps for `actions/cache`, `github/codeql-action`, `zizmorcore/zizmor-action`, and `astral-sh/setup-uv`. Those bumps had been blocked by a schema bug in `.github/dependabot.yml` (the `github-actions` ecosystem rejected `semver-{major,minor,patch}-days` cooldown keys), which is now fixed — see #257. A rebase will pull the new SHAs into your branch as a side effect; that's expected.

Now the actual reason for this ping: small conflict in `docs/modes.md` after the latest `main`:

- Your branch updates the Triage row to mention `release-management` with count `12 + 4 proposed`.
- Main now has the Triage row reading `experimental (pr-management, issue-management, contributor-nomination) | 13`.

The resolution is to keep both: add `release-management` to your status string after `contributor-nomination`, and update the count to e.g. `13 + 4 proposed` (or whatever the right math is after `contributor-nomination` landed). Could you rebase or merge `main` in and reconcile? Happy to push the fix to your branch myself if you'd prefer — let me know.

andreahlert:
> Hi @andreahlert — heads-up first: `main` was just refreshed with pinned-action SHA bumps for `actions/cache`, `github/codeql-action`, `zizmorcore/zizmor-action`, and `astral-sh/setup-uv`. Those bumps had been blocked by a schema bug in `.github/dependabot.yml` (the `github-actions` ecosystem rejected `semver-{major,minor,patch}-days` cooldown keys), which is now fixed — see #257. A rebase will pull the new SHAs into your branch as a side effect; that's expected.
> 
> Now the actual reason for this ping: small conflict in `docs/modes.md` after the latest `main`:
> 
> * Your branch updates the Triage row to mention `release-management` with count `12 + 4 proposed`.
> * Main now has the Triage row reading `experimental (pr-management, issue-management, contributor-nomination) | 13`.
> 
> The resolution is to keep both: add `release-management` to your status string after `contributor-nomination`, and update the count to e.g. `13 + 4 proposed` (or whatever the right math is after `contributor-nomination` landed). Could you rebase or merge `main` in and reconcile? Happy to push the fix to your branch myself if you'd prefer — let me know.

Think is okay right now. Could you take extra look?

projects/_template/release-management-config.md
justinmclean: Thanks that works for me

docs/release-management/process.md
justinmclean: Much better

docs/release-management/process.md
justinmclean: The keyserver is not the KEYS file, but that file, I think most of them are not operational anymore, and fewer ASF members are having key signing parties. Looks good.

docs/release-management/process.md
justinmclean: Also all good by me.

docs/release-management/process.md
justinmclean: Yes that would work for me

docs/release-management/process.md
justinmclean: Fair point on the parallelism. Step 14 says it runs parallel to 12 and the flowchart matches, so that's fine. Two things remain:

- S12 → S13 / S14 → S13 implies a dependency that isn't there. Step 13's listed inputs (planning issue, [VOTE]/[RESULT] URLs, voter list, artefact list, promotion revision, [ANNOUNCE] URL) include nothing from the archive sweep (12) or version bump (14). 13 is really a function of everything through 11 — so 12 and 14 are independent terminals, not feeders into 13.

- The stateDiagram-v2 contradicts the flowchart — it ends announced → archived (12) → [*], dropping 13 and 14 entirely. The flowchart ends at 13.

justinmclean (commented):
See comments once done, consider approved.

potiuk:
Hi @andreahlert — was sweeping the PR queue and tried a maintainer-side rebase to bring this onto current `main`, but the conflicts in `README.md` and `docs/modes.md` aren't mechanical: the mode taxonomy on `main` has evolved (mode counts changed, Mentoring and Pairing moved from "proposed" to "experimental"), so wiring the `release-management` family into the families table needs a real composition pass.

Whenever you next pick this up, a rebase from your end would be great. You'll also pick up the new `tests-ok` umbrella CI job from #341, which replaces the individual `pytest (...)` matrix entries in `.asf.yaml`.

No rush — just flagging so it's not a surprise on next push.

docs/release-management/process.md
andreahlert: fixed both. flowchart now has S11 → S12, S11 → S13, S11 → S14 as three independent terminals, dropped the S12 → S13 / S14 → S13 edges since 13 doesn't actually consume anything from the archive sweep or the snapshot bump

stateDiagram had the same gap, was ending at archived → [*]. added announced → audited (13) and announced → bumped (14) as parallel branches, all three terminate at [*] now so it matches the flowchart

WDYT?

andreahlert:
> Hi @andreahlert — was sweeping the PR queue and tried a maintainer-side rebase to bring this onto current `main`, but the conflicts in `README.md` and `docs/modes.md` aren't mechanical: the mode taxonomy on `main` has evolved (mode counts changed, Mentoring and Pairing moved from "proposed" to "experimental"), so wiring the `release-management` family into the families table needs a real composition pass.
> 
> Whenever you next pick this up, a rebase from your end would be great. You'll also pick up the new `tests-ok` umbrella CI job from #341, which replaces the individual `pytest (...)` matrix entries in `.asf.yaml`.
> 
> No rush — just flagging so it's not a surprise on next push.

rebased. merge in c447a81, only real conflict was the Pairing row in docs/modes.md — kept your experimental | 1 from #251, didn't regress it

also synced the mentoring row in README.md to experimental since #252 already moved it there in modes.md but README was still saying "proposed — not yet formally adopted"

.asf.yaml auto-merged clean, picked up the tests-ok umbrella from #341. CI is running now

while I was in there I also fixed the two diagram nits @justinmclean  had left open on process.md (flowchart edges + stateDiagram terminals), commit 9056cbb

potiuk (approved):
Re-look since my May 15 approval: the substantive iteration with @justinmclean
on Steps 12/13/14 is now landed in the branch — flowchart has `S11 → S12`,
`S11 → S13`, `S11 → S14` as three independent terminals (no S12→S13 / S14→S13
edges), and the stateDiagram has `announced → archived`, `announced → audited`,
`announced → bumped` all terminating at `[*]`. Both match what Justin asked
for in his May 25 follow-up.

The 10-skill family spec is well-grounded: the four cross-cutting boundaries
(especially Boundary 1 — *agent never holds the RM's signing key*) match
the `security-cve-allocate` pattern of "emit paste-ready commands, human
runs them" and slot cleanly into RFC-AI-0004 Principle 1. The 3-dimension
backend abstraction (`release_dist_backend` / `release_approval_mechanism` /
`release_announce_backend`) makes non-ASF adopters first-class from day one.
Approve from me on the substance.

### One thing left before merge: rebase

The branch is **25 commits behind `main`** and the conflicts in `README.md`
(families table) and `docs/modes.md` (modes table) aren't mechanical — `main`
evolved while this PR sat:

- Triage row count moved from "10 shipping + 4 proposed" → **13**
- Mentoring moved from `proposed` (0 skills) → **`experimental`** (1 skill)
- Pairing moved from `proposed` (0 skills) → **`experimental`** (1 skill)
- Drafting count moved from "1 shipping + 6 proposed" → **2**

These need composing with this PR's release-management additions to each
row — not just an `ours` / `theirs` resolution. Same shape as the conflict
I flagged in the earlier bulk-rebase session. Best done by you since the
right composition depends on exactly how you want `release-management`
framed alongside the post-update other-mode counts.

Once rebased, branch is mergeable.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Steward
> maintainer. The maintainer approving this PR has read the findings and signed off.
> If something feels off, please reply on the PR and a maintainer will follow up.*
>
> *More on how Apache Steward handles maintainer review:*
> [CONTRIBUTING.md](https://github.com/apache/airflow-steward/blob/main/CONTRIBUTING.md).

andreahlert:
> Re-look since my May 15 approval: the substantive iteration with @justinmclean on Steps 12/13/14 is now landed in the branch — flowchart has `S11 → S12`, `S11 → S13`, `S11 → S14` as three independent terminals (no S12→S13 / S14→S13 edges), and the stateDiagram has `announced → archived`, `announced → audited`, `announced → bumped` all terminating at `[*]`. Both match what Justin asked for in his May 25 follow-up.
> 
> The 10-skill family spec is well-grounded: the four cross-cutting boundaries (especially Boundary 1 — _agent never holds the RM's signing key_) match the `security-cve-allocate` pattern of "emit paste-ready commands, human runs them" and slot cleanly into RFC-AI-0004 Principle 1. The 3-dimension backend abstraction (`release_dist_backend` / `release_approval_mechanism` / `release_announce_backend`) makes non-ASF adopters first-class from day one. Approve from me on the substance.
> 
> ### One thing left before merge: rebase
> The branch is **25 commits behind `main`** and the conflicts in `README.md` (families table) and `docs/modes.md` (modes table) aren't mechanical — `main` evolved while this PR sat:
> 
> * Triage row count moved from "10 shipping + 4 proposed" → **13**
> * Mentoring moved from `proposed` (0 skills) → **`experimental`** (1 skill)
> * Pairing moved from `proposed` (0 skills) → **`experimental`** (1 skill)
> * Drafting count moved from "1 shipping + 6 proposed" → **2**
> 
> These need composing with this PR's release-management additions to each row — not just an `ours` / `theirs` resolution. Same shape as the conflict I flagged in the earlier bulk-rebase session. Best done by you since the right composition depends on exactly how you want `release-management` framed alongside the post-update other-mode counts.
> 
> Once rebased, branch is mergeable.
> 
> > _This review was drafted by an AI-assisted tool and confirmed by an Apache Steward
> > maintainer. The maintainer approving this PR has read the findings and signed off.
> > If something feels off, please reply on the PR and a maintainer will follow up._
> > _More on how Apache Steward handles maintainer review:_
> > [CONTRIBUTING.md](https://github.com/apache/airflow-steward/blob/main/CONTRIBUTING.md).

Done. Could you TAL?

PR #164, opened 2026-05-15
feat(list-skills): human-facing index that auto-generates from frontmatter

choo121600:
## Why

The repo currently exposes 21 skills, but lacks a quick human-facing
way to discover them.

Skill routing already relies on the live frontmatter `description`
fields internally, but humans had no equivalent overview.

## What

Adds a new `list-skills` skill. On `/list-skills`, it runs a small
Python script which:

- walks `.claude/skills/*/SKILL.md`,
- parses frontmatter for `name` and `description`,
- groups skills by family prefix (with `pr-management` registered as
  a two-token family in `KNOWN_TWO_TOKEN_FAMILIES`),
- prints each skill with the first sentence of its description.

The output is regenerated from the live filesystem on every run, so
there is no separate index to maintain.

## Scope

- Read-only on `.claude/skills/`.
- No external content reads, state mutation, or security-checklist
  patterns triggered.
- `tools/skill-validator` passes (`OK (no violations)`).
- One new dir: `.claude/skills/list-skills/` (SKILL.md + script).
  No changes elsewhere.

## Test plan

- [x] `python3 .claude/skills/list-skills/scripts/list_skills.py`
      → all 21 current skills listed under 5 families.
- [x] `--verbose` flag produces the alternate two-line layout.
- [x] `uv run --directory tools/skill-validator skill-validate`
      → no violations.

image

### DIFF
--- a/.claude/skills/list-steward-skills/SKILL.md
+++ b/.claude/skills/list-steward-skills/SKILL.md
@@ -1,0 +1,111 @@
+---
+name: list-steward-skills
+description: |
+  Print a human-readable index of every skill in this repository,
+  grouped by family prefix (`pr-management`, `security`, `setup`,
+  …) with each skill's name and the first sentence of its
+  `description`. The listing is generated on every run from the
+  live `.claude/skills/*/SKILL.md` files, so it never goes stale
+  when skills are added, removed, or rewritten.
+when_to_use: |
+  Invoke when a human asks *"what skills are available"*, *"list
+  the skills"*, *"show me the skills in this repo"*, *"give me a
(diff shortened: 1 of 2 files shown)


potiuk:
OH wowowowwooow !  -> should we name it "list-steward-skills" (or soon when we rename "list-magpie-skills" ?

choo121600:
Yeah, I think just `list-skills` could be a bit confusing
`list-steward-skills` sounds better for now, and then we can do a bulk rename from `steward` --> `magpie` later on :)

potiuk (approved):

potiuk:
NICE!

PR #215, opened 2026-05-18
add Privacy-LLM gate-check validator 

justinmclean:
## What

Adds a new `validate_privacy_patterns` check to the skill validator that
enforces Pattern 6 from `write-skill/security-checklist.md`: skills that
read full issue bodies from the private `<tracker>` repo must run the
Privacy-LLM gate-check before making any outbound LLM call.

Also fixes the two skills the new check identified as missing the gate.

## Why

Skills like `security-issue-deduplicate` and `security-issue-fix` pass
embargoed CVE detail, reporter PII, and other private content from
`<tracker>` issue bodies directly to the model. The privacy-LLM
gate-check is the required pre-flight for that — it was documented in
the security checklist but not enforced.

## Changes

**`tools/skill-validator`**
- New `PRIVACY_CATEGORY` constant (soft violation category)
- New `validate_privacy_patterns` — fires when a skill has an
  external-content `mode` + references `<tracker>` + contains
  `gh issue view` (the read discriminator). Skills that only write to
  or query metadata from the tracker are exempt.
- 11 new tests in `TestPrivacyPatternP6`

**`.claude/skills`**
- `security-issue-deduplicate`: added privacy-LLM gate to Step 0
- `security-issue-fix`: added privacy-LLM gate to Step 0

## Notes

The `gh issue view` discriminator is what separates genuine gaps from
false positives — `security-issue-import-from-pr` references `<tracker>`
as a write destination and uses only `gh search`/`gh api` for reads,
so it is correctly exempt.

### DIFF
--- a/tools/skill-validator/src/skill_validator/__init__.py
+++ b/tools/skill-validator/src/skill_validator/__init__.py
@@ -498,7 +498,7 @@ def extract_headings(text: str) -> set[str]:
 # or attacker-controlled content.
 _BODY_INLINE_RE = re.compile(r'--body[\s=]["\']')
 
-_FENCED_CODE_RE = re.compile(r"^[ \t]{0,3}```[\s\S]*?^[ \t]{0,3}```", re.MULTILINE)
+_FENCED_CODE_RE = re.compile(r"^ {0,3}```[\s\S]*?^ {0,3}```", re.MULTILINE)
 _DOUBLE_BACKTICK_RE = re.compile(r"``[\s\S]+?``")
 _SINGLE_BACKTICK_RE = re.compile(r"(?<!`)`(?!`)[\s\S]+?(?<!`)`(?!`)")
 
--- a/tools/skill-validator/src/skill_validator/__init__.py
+++ b/tools/skill-validator/src/skill_validator/__init__.py
@@ -68,9 +68,54 @@ PROJECTS_TEMPLATE_DIR = Path("projects/_template")
 REQUIRED_FRONTMATTER_KEYS = {"name", "description", "license"}
 OPTIONAL_FRONTMATTER_KEYS = {"when_to_use", "mode"}
 ALLOWED_LICENSES = {"Apache-2.0"}
-# MISSION mode taxonomy — see docs/modes.md.
-# "Auto-merge" deliberately excluded: it is off per MISSION sequencing.
-ALLOWED_MODES = {"Triage", "Mentoring", "Drafting", "Pairing"}
+
+
+def _read_mode_table() -> dict[str, str]:
+    """Read the canonical MISSION mode table from ``docs/modes.md``."""
+    starts = [Path.cwd().resolve(), Path(__file__).resolve().parent]
+    roots: list[Path] = []
+    for start in starts:
+        roots.extend([start, *start.parents])
+
+    rejected: list[str] = []
+    for root in roots:
+        modes_doc = root / DOCS_DIR / "modes.md"
+        if not modes_doc.is_file():
+            continue
+        text = modes_doc.read_text(encoding="utf-8")
+        if "## Modes at a glance" not in text:
+            rejected.append(f"{modes_doc}: missing '## Modes at a glance' section marker")
+            continue
+        modes_table = text.split("## Modes at a glance", 1)[1].split("## Triage", 1)[0]
+        modes: dict[str, str] = {}
+        for line in modes_table.splitlines():
+            if not line.startswith("| **"):
+                continue
+            cells = [cell.strip() for cell in line.strip("|").split("|")]
+            if len(cells) < 3:
+                continue
+            mode = cells[0].strip("*")
+            status = cells[2].strip()
+            if mode and status:
+                modes[mode] = status
+        if modes:
+            return modes
+        rejected.append(
+            f"{modes_doc}: found '## Modes at a glance' but parsed 0 modes "
+            f"(expected rows like '| **<Mode>** | … | <status> |')"
+        )
+
+    if rejected:
+        raise RuntimeError("could not parse mode taxonomy from docs/modes.md — " + "; ".join(rejected))
+    searched = dict.fromkeys(str(r / DOCS_DIR / "modes.md") for r in roots)
+    raise RuntimeError("could not locate docs/modes.md; searched: " + ", ".join(searched))
+
+
+# MISSION mode taxonomy — docs/modes.md is canonical.
+_MODE_STATUS_BY_NAME = _read_mode_table()
+_MODE_TAXONOMY = set(_MODE_STATUS_BY_NAME)
+_OFF_MODES = {mode for mode, status in _MODE_STATUS_BY_NAME.items() if status == "off"}
+ALLOWED_MODES = _MODE_TAXONOMY - _OFF_MODES
 
 # Forbidden hardcoded project references (fixed strings, case-sensitive)
 FORBIDDEN_PATTERNS: list[str] = [
@@ -147,6 +192,7 @@ INJECTION_GUARD_TODO_CATEGORY = "injection_guard_todo"
 
 GH_LIST_CATEGORY = "gh_list_no_limit"
 SECURITY_PATTERN_CATEGORY = "security_pattern"
+PRIVACY_CATEGORY = "privacy"
 SOFT_CATEGORIES: frozenset[str] = frozenset(
     {
         PRINCIPLE_CATEGORY,
@@ -154,6 +200,7 @@ SOFT_CATEGORIES: frozenset[str] = frozenset(
         INJECTION_GUARD_TODO_CATEGORY,
         SECURITY_PATTERN_CATEGORY,
         GH_LIST_CATEGORY,
+        PRIVACY_CATEGORY,
     }
 )
 
@@ -219,6 +266,36 @@ _FIELD_PLACEHOLDER_RE = re.compile(
     r"(?!(?:@|[\"']@))"
     r"(?:[\"'][^\"'\s]*<[^>]+>[^\"'\s]*[\"']|[^\s\"']*<[^>]+>[^\s\"']*)"
 )
+
+# ---------------------------------------------------------------------------
+# Privacy-LLM gate-check constants (write-skill/security-checklist.md § Pattern 6)
+# ---------------------------------------------------------------------------
+
+# Modes that can process external / attacker-controlled content and need the
+# Privacy-LLM gate when they read private tracker bodies.  Derived from
+# docs/modes.md taxonomy constants above: Pairing is intentionally excluded
+# because the human remains in the loop; Auto-merge is currently excluded only
+# because it is in _OFF_MODES.  When the first Auto-merge skill ships, remove
+# it from _OFF_MODES so body-reading Auto-merge skills are gated by default.
+_PRIVACY_EXTERNAL_CONTENT_MODES: frozenset[str] = frozenset(ALLOWED_MODES - {"Pairing"})
+
+_TRACKER_PLACEHOLDER = "<tracker>"
+_TRACKER_ISSUE_VIEW_RE = re.compile(r"\bgh\s+issue\s+view\b")
+_TRACKER_ISSUE_API_RE = re.compile(r"\bgh\s+api\s+/?repos/<tracker>/issues/[^\s`]+")
+_TRACKER_ISSUE_API_MUTATION_RE = re.compile(r"\s-X\s+(?:PATCH|POST|PUT|DELETE)\b")
+# TODO: detect body reads through ``gh api graphql`` and
+# ``gh issue list --json body`` once the validator has command parsing
+# rich enough to avoid broad prose false positives.
+_PRIVACY_LLM_GATE_PHRASE = "privacy-llm-check"
+_PRIVACY_GATE_SECTION_RE = re.compile(
+    r"^(?:"
+    r"prerequisites?(?:\b|$)"
+    r"|pre[- ]?flight(?:\b|$)"
+    r"|step\s*0(?:\b|$)"
+    r")",
+    re.IGNORECASE,
+)
+_ANTI_EXAMPLE_SECTION_RE = re.compile(r"\b(?:don'?t|anti[- ]?example|bad|wrong)\b", re.IGNORECASE)
 
 ACTION_INVENTORY_COMMA_THRESHOLD = 5
 
@@ -421,7 +498,7 @@ def extract_headings(text: str) -> set[str]:
 # or attacker-controlled content.
 _BODY_INLINE_RE = re.compile(r'--body[\s=]["\']')
 
-_FENCED_CODE_RE = re.compile(r"^```[\s\S]*?^```", re.MULTILINE)
+_FENCED_CODE_RE = re.compile(r"^[ \t]{0,3}```[\s\S]*?^[ \t]{0,3}```", re.MULTILINE)
 _DOUBLE_BACKTICK_RE = re.compile(r"``[\s\S]+?``")
 _SINGLE_BACKTICK_RE = re.compile(r"(?<!`)`(?!`)[\s\S]+?(?<!`)`(?!`)")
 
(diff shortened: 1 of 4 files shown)


potiuk (approved):
Nice!

potiuk:
Hi @justinmclean — heads-up: this PR currently shows as conflicting against `main`.

The conflicts are in `tools/skill-validator/src/skill_validator/__init__.py` and `tools/skill-validator/tests/test_validator.py` and overlap with merged #220 ("detect Pattern 4 injection-guard callout"), which reorganised the validator registry and the test suite in the same area.

I had a look at rebasing it on your behalf, but resolving how the new validator from this PR should coexist with the Pattern 4 infrastructure feels like a judgement call you should make rather than me guessing. Could you rebase against latest `main` and resolve when you get a moment? Happy to re-review once it's clean.

Thanks!

tools/skill-validator/src/skill_validator/__init__.py
```diff
@@ -570,6 +590,61 @@ def validate_principle_compliance(path: Path, text: str) -> Iterable[Violation]:
...
+    if mode not in _EXTERNAL_CONTENT_MODES:
+        return
+
+    # Only flag skills that both reference the tracker AND read full issue bodies.
+    if _TRACKER_PLACEHOLDER not in text:
+        return
+    if _TRACKER_READ_PHRASE not in text:
+        return
+
+    if _PRIVACY_LLM_GATE_PHRASE not in text:
```
andreahlert: this is a plain `in text` substring check, so any occurrence of the literal `privacy-llm-check` anywhere in the file satisfies it: inside an HTML comment, a "don't do this" anti-example, a changelog bullet, or a TODO note. a skill author can write `<!-- TODO: wire up privacy-llm-check -->` and the validator passes while the gate is functionally absent.

that converts the check from a guardrail into a tripwire against accidental omission only, not against intentional or sloppy bypass, which inverts the threat model for a privacy gate.

minimum fix: require the phrase inside a fenced code block, and ideally inside a Prerequisites / Step 0 section (you already locate sections structurally in other validators). at least exclude HTML comments and fenced "don't" blocks.

tools/skill-validator/src/skill_validator/__init__.py
```diff
@@ -131,10 +131,30 @@
...
+_EXTERNAL_CONTENT_MODES: frozenset[str] = frozenset({"Triage", "Mentoring", "Drafting"})
+
+# The placeholder that marks a skill as referencing the private security tracker.
+_TRACKER_PLACEHOLDER = "<tracker>"
+
+# Indicates the skill actually *reads* full issue content from the tracker.
+# Skills that only write to / query metadata from the tracker (e.g. create an
+# issue, list milestones) do not pass private content to the model and are
+# therefore exempt from the Privacy-LLM gate-check.
+_TRACKER_READ_PHRASE = "gh issue view"
```
andreahlert: literal substring of `gh issue view` is fine for today (audited the tree, no skill currently reads bodies via REST). but `gh api repos/<tracker>/issues/<N>` (GET, no `-X PATCH`) also returns the full body and would slip past this check silently.

not blocking since nothing in main hits that path, but a compound discriminator (`gh issue view` OR a regex like `gh api\s+repos/<tracker>/issues/[^\s]+` without `-X`) would close the gap before the next skill author finds it.

tools/skill-validator/src/skill_validator/__init__.py
```diff
@@ -131,10 +131,30 @@
...
-    {PRINCIPLE_CATEGORY, TRIGGER_PRESERVATION_CATEGORY, BODY_INLINE_CATEGORY},
+    {PRINCIPLE_CATEGORY, TRIGGER_PRESERVATION_CATEGORY, BODY_INLINE_CATEGORY, PRIVACY_CATEGORY},
 )
 
+# ---------------------------------------------------------------------------
+# Privacy-LLM gate-check constants (write-skill/security-checklist.md § Pattern 6)
+# ---------------------------------------------------------------------------
+
+# Skill modes that process external / attacker-controlled content.
+_EXTERNAL_CONTENT_MODES: frozenset[str] = frozenset({"Triage", "Mentoring", "Drafting"})
```
andreahlert, line 256: `{"Triage", "Mentoring", "Drafting"}` is hardcoded here while the canonical mode list lives in `write-skill`. when a new mode lands there (e.g. `Coding`, `Reviewing`), this filter won't extend and the gate-check silently stops firing on the new mode.

would import from a single source-of-truth, or at least add a comment pointing at where the canonical list lives so future-you remembers to sync.

tools/skill-validator/src/skill_validator/__init__.py
```diff
@@ -570,6 +590,61 @@ def validate_principle_compliance(path: Path, text: str) -> Iterable[Violation]:
...
+
+    fm = parse_frontmatter(text) or {}
+    mode = fm.get("mode", "")
+    if mode not in _EXTERNAL_CONTENT_MODES:
+        return
+
+    # Only flag skills that both reference the tracker AND read full issue bodies.
+    if _TRACKER_PLACEHOLDER not in text:
+        return
+    if _TRACKER_READ_PHRASE not in text:
```
andreahlert: nice call splitting the discriminator into two stages (`<tracker>` AND `gh issue view`). that's what keeps `security-issue-import-from-pr` correctly exempt: it references `<tracker>` only as a write destination and never `gh issue view`s. design holds.

andreahlert (changes requested):
@justinmclean pulled this locally. 86 tests pass and `skill-validate --strict` is clean.

audited the discriminator: 7 skills with `mode + <tracker> + gh issue view` are covered, `security-issue-import-from-pr` is correctly exempt (only `gh search` / `gh api` for metadata + writes). design works for the current codebase.

three things flagged inline. the substring-match one is the real concern, the other two are follow-up scope.

also branch is `mergeable=CONFLICTING` so a rebase is needed anyway. would fold the substring fix in while you're rebasing.

tools/skill-validator/src/skill_validator/__init__.py
justinmclean: Thanks, agreed. This is fixed now.

The gate is no longer a plain `in text` substring check. The validator now strips HTML comments and only treats `privacy-llm-check` as satisfying the gate when it appears inside a fenced command block, so TODOs, prose mentions, HTML comments, and inline-code anti-examples no longer bypass the check.

I also added regression coverage for:

- `<!-- TODO: wire up privacy-llm-check -->` still failing
- prose mentions still failing
- inline-code mentions still failing
- fenced command blocks passing
- indented fenced command blocks passing

While rebasing I also fixed the fenced-code detector to handle indented fences, since the actual Step 0/preflight command blocks live inside numbered-list indentation.

tools/skill-validator/src/skill_validator/__init__.py
justinmclean: Good point. The original implementation matched the usage patterns that existed in the tree at the time, but you’re right that gh api repos/<tracker>/issues/<N> is another read path that exposes the full issue body.

I’ve already updated it to use a broader discriminator so read-style gh api issue access is covered as well, while avoiding false positives on PATCH/update operations.

tools/skill-validator/src/skill_validator/__init__.py
justinmclean: Fair point. It works as-is because the validator and write-skill are currently evolving together, but you’re right that this creates a maintenance hazard if the mode list expands and the two drift apart.

I’ve already added a note pointing back to the canonical mode list so future changes don’t silently diverge. Longer term, pulling from a single source of truth would be cleaner if we end up reusing the list in more than one place.

tools/skill-validator/src/skill_validator/__init__.py
justinmclean: Thanks, that was the goal.

tools/skill-validator/src/skill_validator/__init__.py
andreahlert: Thanks, the HTML-comment, prose, and inline-code cases are properly closed, and the regression tests match. Two gaps from the original review still stand though:

**1. Fenced anti-example blocks still satisfy the gate.** `_has_privacy_gate_command` iterates every fenced block in the body with no section context, so a "Don't do this" / "Bad example" section containing:

````markdown
## Don't do this

```bash
uv run --project <framework>/tools/privacy-llm/checker privacy-llm-check
```
````

passes silently. The docstring on `validate_privacy_patterns` actually claims "anti-examples do not satisfy the check", but that is only true for anti-examples written as prose, not for fenced ones, which is the more dangerous shape.

**2. The Prerequisites / Step 0 location is not enforced.** The error message tells authors to put the command "in the Prerequisites / Step 0 section", but the validator accepts the phrase in any fenced block anywhere in the file (e.g. an appendix, a changelog snippet, a `## History` section). The original suggestion was to reuse the structural section locator you already have in other validators.

Minimal closure for both: scope `_fenced_code_blocks` to blocks whose enclosing heading matches a Prerequisites / Preflight / Step 0 pattern, and/or exclude blocks under headings matching `(?i)don'?t|anti.?example|bad|wrong`. The first option subsumes the second and aligns the behavior with the error message.

Minor: `_FENCED_CODE_RE = r"^[ \t]{0,3}```..."` accepts tabs in fence indentation; CommonMark only allows 0-3 spaces. Harmless in practice but worth tightening to `^ {0,3}```` for parity with the spec.

File: tools/skill-validator/src/skill_validator/__init__.py (hunk 2 above)
Comment (andreahlert) [reply tools/skill-validator/src/skill_validator/__init__.py]: Good, the new compound check via `_TRACKER_ISSUE_API_RE` covers the common shape. A couple of holes still slip through though:

**1. Leading-slash form bypasses.** `gh` accepts both `gh api repos/...` and `gh api /repos/...`. The regex `\bgh\s+api\s+repos/<tracker>/issues/[^\s]+` only matches the slashless form, so:

```bash
gh api /repos/<tracker>/issues/<N>
```

reads the full body and the check returns false. Easy fix: `\bgh\s+api\s+/?repos/<tracker>/issues/[^\s]+`.

**2. Multi-line mutation produces a false positive.** `_has_tracker_body_read` iterates `body.splitlines()` and checks `_TRACKER_ISSUE_API_MUTATION_RE` on the same line as the URL. A standard backslash-continued PATCH:

```bash
gh api repos/<tracker>/issues/<N> \
  -X PATCH \
  -f title=x
```

matches the read regex on line 1 and the mutation exclusion never fires, so a pure write skill gets flagged as a body-read and forced to declare the privacy gate. Direction is safe (over-gating, not under-gating), but it will surprise authors. Either join continuation lines before matching or scope the exclusion to the surrounding fenced block instead of the single line.

**3. `gh api graphql` / `gh issue list --json body` are not covered.** Edge cases today, but the same threat model (reads issue body via gh) applies. Worth a TODO at least.

tools/skill-validator/src/skill_validator/__init__.py
andreahlert: The pointer comment helps, but it points at the wrong file: write-skill/security-checklist.md § Pattern 6 describes the gate-check pattern, not the mode taxonomy. The canonical mode list actually lives in
  [docs/modes.md](https://github.com/docs/modes.md) and currently enumerates five modes (Triage, Mentoring, Drafting, Pairing, Auto-merge), not the four the threat-model comment implies.

  That matters because Auto-merge is in the taxonomy but missing from _PRIVACY_EXTERNAL_CONTENT_MODES, and its threat model is the opposite of Pairing's: no human in the loop. The justification for excluding Pairing ("human remains in the loop") doesn't transfer to Auto-merge, so when the first Auto-merge skill lands that reads <tracker> bodies, the gate-check will silently no-op on it. Today there are no
  Auto-merge skills (per docs/modes.md:182), so it's a latent gap, not a live one, but the comment as written would have told future-you the frozenset is complete when it isn't.

  Two concrete adjustments:

  1. Repoint the comment to docs/modes.md (the actual canonical source), and explicitly state Auto-merge's status: either "intentionally excluded because " or "TODO when the first Auto-merge skill ships".
  2. For the longer-term single-source-of-truth: a tiny shared module or a parser that reads the modes table out of docs/modes.md at validator import time would close the drift mechanically; the current "I'll
  remember to sync" approach has the same failure mode the original review flagged.

andreahlert (changes requested):
Just a few adjustments left. Put all in the comments.

tools/skill-validator/src/skill_validator/__init__.py
justinmclean: The validator now only counts privacy-llm-check when it appears in a fenced block under a real section whose heading starts with Prerequisites, Preflight, or Step 0. It also ignores fenced blocks under anti-example headings like Don't do this, Bad example, Wrong, or Anti-example.

I added regression coverage for:

fenced command in istory does not satisfy the gate
command after a  Step 0 section, but under sibling History, does not satisfy the gate
Appendix: Step 0 from an older version does not satisfy the gate
Step 0 / Bad example does not satisfy the gate
real Step 0 / Prerequisites fenced commands still satisfy the gate
I also verified against the existing security skills; their gate calls are under ## Step 0 — Pre-flight check, and skill-validate passes cleanly.

tools/skill-validator/src/skill_validator/__init__.py
justinmclean: Thanks. I addressed the two holes and left 3 as a TODO.

tools/skill-validator/src/skill_validator/__init__.py
justinmclean: The validator now parses the docs/modes.md “Modes at a glance” table at import time and derives the full mode taxonomy from the table rows, including ones that are off.

tools/skill-validator/src/skill_validator/__init__.py
```diff
@@ -68,9 +68,47 @@
...
+
+    msg = "could not read mode taxonomy from docs/modes.md"
+    raise RuntimeError(msg)
+
+
+# MISSION mode taxonomy — docs/modes.md is canonical.
+_MODE_STATUS_BY_NAME = _read_mode_table()
+_MODE_TAXONOMY = set(_MODE_STATUS_BY_NAME)
+_OFF_MODES = {mode for mode, status in _MODE_STATUS_BY_NAME.items() if status == "off"}
+ALLOWED_MODES = _MODE_TAXONOMY - _OFF_MODES
```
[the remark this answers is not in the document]
justinmclean: This doesn't really apply — skill-validator isn't a redistributable library, it's a repo-bound CI tool (uv run --project tools/skill-validator) that only validates this repo's .claude/skills/ against docs/. "Installed without docs" is a state where the whole tool is a no-op anyway, so there's nothing to harden for.

The cwd concern is already handled: _read_mode_table() also searches from Path(__file__).resolve().parent upward, so it finds <repo>/docs/modes.md regardless of cwd.

Fallback table is a no — docs/modes.md is intentionally the single source of truth for the taxonomy; an embedded copy reintroduces the drift this validator exists to catch.

Fair point that the import-time read raises instead of failing open like the rest of the module, but that's the behavior I want from a consistency check. If we ever soften it, lazy-loading in run_validation is the route.

tools/skill-validator/src/skill_validator/__init__.py
```diff
@@ -200,6 +240,36 @@
...
+# Modes that can process external / attacker-controlled content and need the
+# Privacy-LLM gate when they read private tracker bodies.  Derived from
+# docs/modes.md taxonomy constants above: Pairing is intentionally excluded
+# because the human remains in the loop; Auto-merge is currently excluded only
+# because it is in _OFF_MODES.  When the first Auto-merge skill ships, remove
+# it from _OFF_MODES so body-reading Auto-merge skills are gated by default.
+_PRIVACY_EXTERNAL_CONTENT_MODES: frozenset[str] = frozenset(ALLOWED_MODES - {"Pairing"})
+
+_TRACKER_PLACEHOLDER = "<tracker>"
+_TRACKER_ISSUE_VIEW_RE = re.compile(r"\bgh\s+issue\s+view\b")
```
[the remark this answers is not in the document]
justinmclean: Fair point that gh issue view isn't bound to <tracker> like the gh api repos/<tracker>/... path is — the binding is really the file-level <tracker> guard. But I'll leave it:

No live false positive: the only bare gh issue view on a public repo is pr-management-mentor, which reads <upstream> and never mentions <tracker>, so the guard already filters it out.

The fix is net-negative for a SOFT check: requiring --repo <tracker> would stop matching the bare gh issue view --json body form that real tracker skills use via default-repo context (e.g. security-issue-fix:642), dropping them out of gate enforcement — a false negative, worse than a dismissable advisory.

No need for any changes.

tools/skill-validator/src/skill_validator/__init__.py
```diff
@@ -638,6 +708,118 @@ def validate_principle_compliance(path: Path, text: str) -> Iterable[Violation]:
...
+
+
+def _has_tracker_body_read(text: str) -> bool:
+    body = _strip_html_comments(_skill_body(text))
+    if _TRACKER_ISSUE_VIEW_RE.search(body):
+        return True
+    for command in _shell_logical_lines(body):
+        if _TRACKER_ISSUE_API_RE.search(command) and not _TRACKER_ISSUE_API_MUTATION_RE.search(command):
+            return True
+    return False
```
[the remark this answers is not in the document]
justinmclean: this is a repeat of the last comment.

tools/skill-validator/tests/test_validator.py
```diff
@@ -176,6 +184,26 @@ def test_mode_optional(self, tmp_path: Path) -> None:
         violations = list(validate_frontmatter(path, text))
         assert violations == []
 
+    def test_mode_taxonomy_matches_docs_modes(self) -> None:
+        docs_modes = Path(__file__).parents[3] / "docs" / "modes.md"
+        modes_table = (
+            docs_modes.read_text(encoding="utf-8")
+            .split("## Modes at a glance", 1)[1]
+            .split("## Triage", 1)[0]
+        )
```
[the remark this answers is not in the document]
justinmclean: Intentional. parents[3] keeps the locator dead-simple and independent; if the layout changes the test fails loudly and it's a one-line fix. Leaving as-is.

tools/skill-validator/src/skill_validator/__init__.py
```diff
@@ -68,9 +68,47 @@
...
+                continue
+            mode = cells[0].strip("*")
+            status = cells[2].strip()
+            if mode and status:
+                modes[mode] = status
+        if modes:
+            return modes
+
+    msg = "could not read mode taxonomy from docs/modes.md"
+    raise RuntimeError(msg)
```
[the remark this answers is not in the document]
justinmclean: Fixed

justinmclean:
Just working on the conflicts

justinmclean:
@andreahlert I believe everything has been done

justinmclean:
Crap more conflicts - working on it

tools/skill-validator/src/skill_validator/__init__.py
```diff
@@ -403,7 +480,7 @@ def extract_headings(text: str) -> set[str]:
 # or attacker-controlled content.
 _BODY_INLINE_RE = re.compile(r'--body[\s=]["\']')
 
-_FENCED_CODE_RE = re.compile(r"^```[\s\S]*?^```", re.MULTILINE)
+_FENCED_CODE_RE = re.compile(r"^[ \t]{0,3}```[\s\S]*?^[ \t]{0,3}```", re.MULTILINE)
```
andreahlert: One leftover from the round-2 review: the fence regex still allows tab indentation.

```suggestion
_FENCED_CODE_RE = re.compile(r"^ {0,3}```[\s\S]*?^ {0,3}```", re.MULTILINE)
```

CommonMark only allows 0-3 spaces (not tabs) for fence indentation. Harmless in practice but worth tightening for spec parity. If you accept the suggestion, GitHub commits it directly to the branch and the PR is good to go from my side.

andreahlert (approved):
LGTM. Thanks @justinmclean.

@choo121600 could you take an extra look and help us here?

choo121600 (approved):
overall lgtm :)

potiuk:
Cool :)

PR #217, opened 2026-05-18
 catch `--body=` equals-sign variant and fix corpus hits

justinmclean:
## What

Extends the existing Pattern 9 body-inline check to also catch the
`--body="..."` / `--body='...'` equals-sign form, and fixes all eight
existing `--body "..."` violations found when running the new check
against the skill corpus.

## Why

The previous regex (`--body\s+["']`) only matched the space-separated
form.  A skill author writing a compact one-liner like
`gh issue create --body="$(cat /tmp/body.md)"` would have silently
passed the pre-commit gate.  Both forms carry the same shell-injection
risk: attacker-controlled content (issue bodies, PR titles, commit
messages) can break out of the quoting and execute arbitrary shell
commands.  `--body-file` sidesteps this entirely by passing a file
descriptor rather than a shell argument.

### `tools/skill-validator`

- **`_BODY_INLINE_RE`** — new regex `--body[\s=]["']` covering both the
  space-separated and equals-sign forms.
- **`validate_body_inline`** — new SOFT validator; skips inline backtick
  spans in prose (e.g. documentation saying "never use `--body "..."`")
  but flags occurrences in fenced code blocks (real agent commands).
- **`_inline_only_code_spans`** — helper returning inline-code span
  positions with fenced blocks excluded, so prose mentions are correctly
  suppressed.
- **`_BODY_INLINE_SKIP_SUFFIXES`** — skip list for
  `write-skill/security-checklist.md`, which intentionally documents the
  bad pattern and uses nested 4/5-backtick fences that confuse the
  backtick-span parser.
- **`BODY_INLINE_CATEGORY`** added to `SOFT_CATEGORIES` and
  `_SOFT_RULE_PREFIXES`; wired into `run_validation`.
- **`TestBodyInline`** (11 tests) — space form, equals form, single- and
  double-quote variants, `--body-file` clean, inline prose skip, line
  number accuracy, SOFT category membership, skip-path behaviour.

### Skill corpus fixes (8 violations → 0)

| File | Fix |
|---|---|
| `setup-override-upstream/SKILL.md` | Step 4 rewritten as fenced block with `--body-file` |
| `security-issue-triage/SKILL.md` | Prose backtick reflowed to single line |
| `pr-management-code-review/posting.md` | Three `--body "$(cat <<'EOF'...)"` blocks replaced with `cat > /tmp/... <<'EOF'; gh ... --body-file` |
| `pr-management-code-review/posting.md` | Prose description updated to match |
| `security-issue-fix/SKILL.md` | `--body "$(cat /tmp/pr-body-<issue>.md)"` → `--body-file /tmp/pr-body-<issue>.md` |

## Corpus result

Zero new hits from the `=` extension (no existing skill uses that form).
Eight existing hits from the space form, all now resolved.

## Testing

```bash
uv run --project tools/skill-validator --group dev pytest \
  tests/test_validator.py::TestBodyInline -v
uv run --project tools/skill-validator --group dev ruff format --check \
  src/ tests/
skill-validate   # expects: skill-validator: OK (no violations)
```

### DIFF
--- a/.claude/skills/pr-management-code-review/posting.md
+++ b/.claude/skills/pr-management-code-review/posting.md
@@ -35,32 +35,31 @@ Golden rule 8 downgrades any auto-`APPROVE` if CI is failing.
 ### Approve
 
 ```bash
-gh pr review <N> --repo <repo> --approve --body "$(cat <<'EOF'
+cat > /tmp/review-body.md << 'EOF'
 [review body here]
 EOF
-)"
+gh pr review <N> --repo <repo> --approve --body-file /tmp/review-body.md
 ```
 
 ### Request changes
 
 ```bash
-gh pr review <N> --repo <repo> --request-changes --body "$(cat <<'EOF'
+cat > /tmp/review-body.md << 'EOF'
 [review body here]
 EOF
-)"
+gh pr review <N> --repo <repo> --request-changes --body-file /tmp/review-body.md
 ```
 
 ### Comment
 
 ```bash
-gh pr review <N> --repo <repo> --comment --body "$(cat <<'EOF'
+cat > /tmp/review-body.md << 'EOF'
 [review body here]
 EOF
-)"
+gh pr review <N> --repo <repo> --comment --body-file /tmp/review-body.md
 ```
 
-The skill always uses **here-doc body passing** (never `--body
-"$STRING"` with quotes) to avoid shell-escape mishaps with PR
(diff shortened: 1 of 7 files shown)


justinmclean:
Note that prek on Mac OS and Linux are giving different results.

potiuk (approved):
Moar security :)

PR #231, opened 2026-05-19
fix(privacy-llm-redactor): correct --field help text and force UTF-8 mapping reads

andreahlert:
## What

Fixes two critical correctness bugs in the `privacy-llm/redactor` package, surfaced by a code review.

### 1. `--field` help text lists type names the parser rejects

`redact.py` advertised `reporter` and code `R` in the `--field` help. Neither exists. The valid types come from `TYPE_CODES` in `mapping.py`: `name/email/phone/ip/handle/address`, codes `N/E/P/IP/H/A`.

A user copying the documented form (`--field reporter:Jane Smith` or `--field R:Jane Smith`) got a `SystemExit` and their PII flowed to the LLM unredacted. Help text now lists the real names and codes.

### 2. `load_mapping` reads with the locale-default encoding

`load_mapping` used `path.read_text()` with no `encoding`, which resolves to `locale.getpreferredencoding(False)` (e.g. `cp1252` on Windows). But `save_mapping_atomic` writes the file as UTF-8. On any non-UTF-8 host this corrupts non-ASCII PII values (accented names, IDN domains) on the save/load round-trip, so `pii-reveal` substitutes the wrong text.

`load_mapping` now reads with `encoding="utf-8"`, matching the writer.

## Changes

- `src/redactor/redact.py` — `--field` help text: `reporter/R` → `name/N`
- `src/redactor/mapping.py` — `load_mapping` reads `encoding="utf-8"`
- `tests/test_redact.py` — regression test on the `--help` output
- `tests/test_mapping.py` — regression test for a non-ASCII round-trip

## Validation

- `pytest`: 53 passed
- `ruff check` / `ruff format` / `mypy`: clean
- `prek`: all hooks pass

### DIFF
--- a/tools/privacy-llm/redactor/src/redactor/mapping.py
+++ b/tools/privacy-llm/redactor/src/redactor/mapping.py
@@ -90,7 +90,7 @@ def load_mapping(path: pathlib.Path) -> dict[str, Entry]:
     """
     if not path.exists():
         return {}
-    raw = json.loads(path.read_text())
+    raw = json.loads(path.read_text(encoding="utf-8"))
     if not isinstance(raw, dict):
         raise ValueError(f"{path}: expected a JSON object at the top level")
     version = raw.get("version")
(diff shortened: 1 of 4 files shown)


potiuk (approved):
NICE!

PR #241, opened 2026-05-20
fix(vulnogram): stderr routing, double-warning, UTF-8 read, strict package regex

andreahlert:
Four small fixes batched in the vulnogram tool path. Each is independently reviewable.

### 1. `check.py`: errors to stderr (functional bug)

The `error: ...` branch only emitted to stderr under `--quiet`, via a ternary that swallowed it otherwise. Callers piping `2>/dev/null` got noise mixed into stdout, and `--quiet` invocations hit exit code 3 with no visible reason.

After: errors always go to stderr regardless of `--quiet`. `valid` / `expired` still go to stdout (skills exact-match those tokens).

Tests added:
- `test_unknown_error_writes_to_stderr_not_stdout`
- `test_unknown_error_quiet_still_writes_to_stderr`

### 2. `credentials.py`: double-warning on chmod failure

`write_session_atomic` warned once if `chmod 700` raised, then warned a second time on the followup `stat().st_mode & 0o077` check. Collapsed into `try/except/else` so the post-check only runs when chmod succeeded.

### 3. `cve_json.py:158`: locale-default read

`tomllib.loads(path.read_text())` decoded with the platform locale. We write UTF-8 elsewhere in the codebase, so a system with `LANG` unset could fail on non-ASCII config (`description`, author names). Fixed to `read_text(encoding="utf-8")`.

### 4. `cve_json.py:707`: redundant regex fallback

`PACKAGE_RE.fullmatch(name) or PACKAGE_RE.match(name)` is a no-op today because `PACKAGE_RE` is anchored with a trailing `$`. Kept as-is it silently downgrades validation if anyone removes the anchor later. Dropped the `or match` branch.

All four sit in `tools/vulnogram/`. `pytest`, `ruff`, `mypy`, `prek` all green.

### DIFF
--- a/tools/vulnogram/generate-cve-json/src/generate_cve_json/cve_json.py
+++ b/tools/vulnogram/generate-cve-json/src/generate_cve_json/cve_json.py
@@ -155,7 +155,7 @@ def _load_config(config_path: Path | str | None = None) -> dict:
             f"  at {_DEFAULT_CONFIG_RELPATH} relative to the cwd.\n"
             f"  Schema: see the generate-cve-json package README."
         )
-    return tomllib.loads(path.read_text())
+    return tomllib.loads(path.read_text(encoding="utf-8"))
 
 
 def _set_config_path(config_path: Path | str) -> None:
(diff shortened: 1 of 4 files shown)


potiuk (approved):
Sweet!

PR #253, opened 2026-05-24
docs(mode-economics): add per-mode token-cost shape page

justinmclean:
> **Generated by the spec-driven build loop.** This page was produced by an
> autonomous run of `tools/spec-loop` (`./loop.sh` — one work item, one branch,
> one PR). Authored by Claude (see the `Generated-by` commit trailer) and
> reviewed + tested by a human before submission.

## What

Adds `docs/mode-economics.md` — an indicative per-skill / per-mode / per-model
token-cost page, so a maintainer evaluating adoption can see the cost shape up
front rather than discovering it after the fact. Covers Triage, Mentoring,
Drafting, Pairing and Auto-merge; model-class trade-offs; local/self-hosted
inference; cost-reduction levers; and the long-term ASF inference-endpoint
roadmap item. Adds a reciprocal cross-reference from `docs/modes.md`.

## Why

`MISSION.md § Affordability` commits to documenting mode economics honestly.
This page is that documentation, and its data is the quantitative input for
future ASF inference-endpoint capacity planning.

## Changes

- `docs/mode-economics.md` — new page.
- `docs/modes.md` — cross-reference back to it.

## Testing — and a limitation it shipped with

The doc gates runnable locally pass: the doctoc TOC matches all 15 headings, every
internal link/anchor resolves (including the MISSION § Affordability deep-link and
the reciprocal `modes.md` cross-ref), file-hygiene is clean, and 12 of 13 referenced
skills exist on `main` (the 13th, `pairing-self-review`, is a sibling unmerged
branch — a name in a table, not a link). `markdownlint`/`typos`/`lychee` should be
confirmed by CI.

**Limitation found in testing — the loop did not detect it.** The page's headline
anchor claimed "one skill file (SKILL.md) ≈ 3,000–6,000 tokens" and that this
overhead "applies" generally. Tokenizing the real catalogue with `cl100k_base`
showed that holds only for the *median* skill: of 26 skills, **9** fall in 3k–6k,
**6** are below, and **11** are above — the large multi-step security skills run far
higher (`security-issue-import` ~22,000, `security-issue-sync` ~36,000, i.e. 4–6×
the stated ceiling). The anchor is corrected here to measured, class-specific
ranges (small ~1k–3k, typical ~3.5k–9k, large security ~11k–36k). The build loop
shipped a plausible but unverified figure; the test-before-PR step is what caught it.

## Notes

- Indicative cost guidance; the counts in the "what tokens means here" anchor are
  now measured (`cl100k_base`), not estimated.

### DIFF
--- a/docs/mode-economics.md
+++ b/docs/mode-economics.md
@@ -1,0 +1,261 @@
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [Mode economics — what does each mode cost to run?](#mode-economics--what-does-each-mode-cost-to-run)
+  - [How to read this page](#how-to-read-this-page)
+    - [What "tokens" means here](#what-tokens-means-here)
+    - [Model classes](#model-classes)
+  - [Per-mode token shape](#per-mode-token-shape)
+    - [Triage](#triage)
+    - [Mentoring](#mentoring)
+    - [Drafting](#drafting)
+    - [Pairing](#pairing)
+    - [Auto-merge](#auto-merge)
+  - [Model class and mode cost shape](#model-class-and-mode-cost-shape)
+  - [Local and self-hosted inference](#local-and-self-hosted-inference)
+  - [Reducing costs](#reducing-costs)
+  - [Long-term: the ASF inference endpoint](#long-term-the-asf-inference-endpoint)
+  - [Cross-references](#cross-references)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<!-- SPDX-License-Identifier: Apache-2.0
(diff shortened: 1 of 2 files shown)


choo121600 (approved):
Cool!
It looks like this document may also need to be updated whenever a new skill is introduced.
Would it make sense to add some guidance or process to help keep this document in sync?

potiuk (approved):
Nice!

potiuk:
> Cool! It looks like this document may also need to be updated whenever a new skill is introduced. Would it make sense to add some guidance or process to help keep this document in sync?

I am planning to do it also with running evals :)

potiuk:
> > Cool! It looks like this document may also need to be updated whenever a new skill is introduced. Would it make sense to add some guidance or process to help keep this document in sync?
> 
> I am planning to do it also with running evals :)

https://github.com/apache/airflow-steward/pull/281 -> cc: @justinmclean @choo121600 @andreahlert :) - I think this way we don't **really** need to run evals in CI.

choo121600:
> #281 -> cc: @justinmclean @choo121600 @andreahlert :) - I think this way we don't **really** need to run evals in CI.

Agreed, I think this approach makes sense :)

PR #345, opened 2026-05-27
feat(tools/jira): add write subcommands (comment, transition, label, assign, field, attach)

### LINKED ISSUE #301: feat(tools/jira): add write subcommands (comment, transition, label, assign, field, attach)
The current Jira bridge in [`tools/jira/`](../tree/main/tools/jira/) is read-only by design — `search`, `issue`, `projects` only. To use Jira as a tracker for skills like `issue-triage`, `security-issue-sync`, and `security-issue-invalidate`, we need a parallel **write** path.

**Suggested subcommands** (mirror the existing read API shape):

- `comment <KEY> --body-file <path>` — post a comment
- `transition <KEY> <transition-name>` — move workflow state
- `label <KEY> --add <name> --remove <name>` — toggle labels
- `assign <KEY> <username>` — set assignee
- `field <KEY> <field-name> --value <value>` — edit a custom field
- `attach <KEY> <file>` — attach a file

Same write-path discipline the GitHub bridge uses applies: every mutation is gated on **explicit user confirmation** in the calling skill — bridges only execute confirmed actions.

**Why:** Apache projects that haven't migrated to GitHub Issues (Lucene/Solr, Cassandra, Camel, Beam, Flink, Kafka, etc.) live on Jira; without a write path the skill family can't operate against them. ASF Jira at `issues.apache.org/jira` is the largest concrete target.

**Reference:**

- Existing read bridge: [`tools/jira/bridge.groovy`](../tree/main/tools/jira/bridge.groovy)
- Write-path discipline reference: [`tools/github/operations.md`](../tree/main/tools/github/operations.md)
- Reimplementations in Python or Bash + curl are welcome — Groovy was the proof-of-concept.

oscerd:
## Summary

- Extends the read-only JIRA bridge with six write subcommands: `comment`, `transition`, `label`, `assign`, `field`, and `attach`
- Write operations require `JIRA_API_TOKEN` and follow the same write-path discipline as the GitHub bridge (mutations gated on explicit user confirmation in the calling skill)
- Adds a pytest test harness with a mock HTTP server (24 tests, auto-skipped when groovy is not on PATH)

## Test plan

- [x] Tests collected and pass locally (`uv run pytest` — 24 tests, skipped when groovy not on PATH)
- [x] Placeholder check passes (`tools/dev/check-placeholders.sh` — OK)
- [ ] CI pre-commit hooks pass
- [ ] Manual verification with a live JIRA instance (optional, requires groovy + API token)

Closes #301


tools/jira/bridge.groovy
```diff
@@ -130,20 +202,172 @@ def cmd_projects(List args) {
...
+        } else {
+            i++
+        }
+    }
+    if (value == null) {
+        System.err.println('error: field requires --value <value>')
+        System.exit(2)
+    }
+    def url = "${TRACKER_URL}/rest/api/2/issue/${key}"
+    def payload = JsonOutput.toJson([fields: [(fieldName): value]])
```
andreahlert: imho the biggest gotcha here — `value` is always a string, so anything that isn't a free-text/number customfield will 400 against real JIRA: priority → `{"name": "High"}`, fixVersions → `[{"name": "1.2.3"}]`, single-select → `{"value": "x"}`, user picker → `{"name": "jdoe"}` (DC) / `{"accountId": "..."}` (Cloud).

Mock doesn't catch it because it accepts any payload.

wdyt about adding `--value-json '<json>'` as an alt to `--value <str>`? Keeps the shorthand for trivial cases.

tools/jira/README.md
```diff
@@ -108,31 +217,64 @@ The bridge reads its configuration from the environment:
 |---|---|
 | `ISSUE_TRACKER_URL` | required; e.g. `https://issues.apache.org/jira` |
 | `ISSUE_TRACKER_PROJECT` | project key (e.g. `FOO`) |
+| `JIRA_API_TOKEN` | required for write subcommands; base64-encoded `email:token` |
```
andreahlert: nit: this is Atlassian Cloud phrasing, but `cmd_assign` below uses `{name: username}` which is DC-only (Cloud needs `accountId`). Since #301 explicitly targets `issues.apache.org/jira` (DC), the impl is right — just the doc reads Cloud-ish.

Maybe:
> base64 of `username:password` or `username:<PAT>` for Apache JIRA (Data Center). Cloud not currently supported (`assign` uses DC `name`, not Cloud `accountId`).

tools/jira/bridge.groovy
```diff
@@ -55,6 +64,76 @@ def httpGet(String urlStr) {
...
+    def conn = url.openConnection()
+    conn.requestMethod = method
+    conn.doOutput = true
+    conn.setRequestProperty('Content-Type', 'application/json')
+    conn.setRequestProperty('Accept', 'application/json')
+    if (!API_TOKEN) {
+        System.err.println('error: JIRA_API_TOKEN is required for write operations')
+        System.exit(2)
+    }
+    conn.setRequestProperty('Authorization', "Basic ${API_TOKEN}")
```
andreahlert: heads-up: ASF JIRA DC PATs are sent as `Authorization: Bearer <pat>`, not Basic. Today a PAT user has to wrap it as `base64(user:pat)` to fake Basic, which bypasses the canonical PAT path.

imho worth a `JIRA_AUTH_SCHEME` env (default `Basic`, accepts `Bearer`) — ASF has been nudging projects toward PAT-only.

tools/jira/bridge.groovy
```diff
@@ -130,20 +202,172 @@ def cmd_projects(List args) {
...
+        System.exit(2)
+    }
+    def issueUrl = "${TRACKER_URL}/rest/api/2/issue/${key}?fields=labels"
+    def current = httpGet(issueUrl)
+    def labels = (current.fields?.labels ?: []) as List
+    labels.addAll(addLabels)
+    labels.removeAll(removeLabels)
+    labels = labels.unique()
+    def url = "${TRACKER_URL}/rest/api/2/issue/${key}"
+    def payload = JsonOutput.toJson([fields: [labels: labels]])
```
andreahlert: nit: read-modify-write loses concurrent label updates. JIRA REST does this atomically in one PUT:

```json
{"update": {"labels": [{"add": "x"}, {"remove": "y"}]}}
```

One round-trip, no race, server handles the merge. Not blocking — fine as a follow-up.

tools/jira/bridge.groovy
```diff
@@ -130,20 +202,172 @@ def cmd_projects(List args) {
...
+def cmd_assign(List args) {
+    def key = args ? args[0] : null
+    validateKey(key)
+    def username = args.size() > 1 ? args[1] : null
+    if (!username) {
+        System.err.println('error: assign requires a username')
+        System.exit(2)
+    }
+    def url = "${TRACKER_URL}/rest/api/2/issue/${key}/assignee"
+    def payload = JsonOutput.toJson([name: username])
```
andreahlert, line 300: fyi: `{name: username}` is DC-only. Cloud needs `{accountId: "..."}`. Fine for #301's scope (ASF JIRA = DC) — just calling it out so the limitation is explicit and we don't silently break Cloud adopters later.

Maybe drop a `// DC payload — Cloud uses accountId` next to it.

tools/jira/bridge.groovy
```diff
@@ -55,6 +64,76 @@ def httpGet(String urlStr) {
...
+        System.exit(2)
+    }
+    conn.setRequestProperty('Authorization', "Basic ${API_TOKEN}")
+    conn.connectTimeout = 10000
+    conn.readTimeout    = 60000
+    def os = conn.outputStream
+    os.write("--${boundary}\r\n".getBytes('UTF-8'))
+    os.write("Content-Disposition: form-data; name=\"file\"; filename=\"${file.name}\"\r\n".getBytes('UTF-8'))
+    os.write("Content-Type: application/octet-stream\r\n\r\n".getBytes('UTF-8'))
+    file.withInputStream { is -> os << is }
```
andreahlert, line 116: nit: no size guard on the upload, and filename in `Content-Disposition` isn't escaped. Source is local FS so risk is low, but a soft cap (e.g. 10MB) would prevent foot-guns.

tools/jira/bridge.groovy
```diff
@@ -130,20 +202,172 @@ def cmd_projects(List args) {
...
+    if (!bodyFile) {
+        System.err.println('error: comment requires --body-file <path>')
+        System.exit(2)
+    }
+    def f = new File(bodyFile)
+    if (!f.exists()) {
+        System.err.println("error: body file not found: ${bodyFile}")
+        System.exit(2)
+    }
+    def body = f.text
```
andreahlert: nit: `f.text` uses platform default encoding. Non-ASCII comment bodies can corrupt on non-UTF-8 hosts. `f.getText("UTF-8")` fixes it.

tools/jira/tests/test_bridge_write.py
```diff
@@ -0,0 +1,395 @@
...
+                entry["body"] = json.loads(body)
+            else:
+                entry["body_raw"] = body
+        _log.requests.append(entry)
+
+        key = f"{method} {self.path.split('?')[0]}"
+        if key in _canned_responses:
+            status, payload = _canned_responses[key]
+        else:
+            status, payload = 200, {}
```
andreahlert, line 85: nit on coverage: mock defaults to `200 + {}` for unmapped routes, so any malformed body slips through silently. A "strict" mode that 400s on unknown keys / wrong shapes would catch the `cmd_field` shape issue I flagged in the groovy file.

Also — `Authorization` header value isn't asserted anywhere (only presence/absence). Worth a sanity check that the scheme + token round-trip correctly.

andreahlert (changes requested):
Scope is tight and the mock harness is a solid first layer — really like where this is going. Requesting changes on three things that imho will bite the first time a skill calls this against `issues.apache.org/jira`:

1. `cmd_field` only sends string values — most non-trivial JIRA fields (priority, version, user, single-select) need objects/arrays and will 400. See inline.
2. README auth description is Cloud-flavored but `cmd_assign` uses DC `{name}`. Needs to either say "DC only" or actually handle Cloud.
3. ASF JIRA PATs use `Authorization: Bearer`, not Basic. Today PAT users have to fake Basic via `base64(user:pat)`.

The rest below are nits / follow-ups — flagging them while I'm here, not blocking on them.

Cross-cutting:
- No retry/backoff anywhere. ASF JIRA does rate-limit; a small `429/5xx → 3 retries w/ backoff` loop would save pain in long skill flows.
- Pre-existing, but since the README is being touched here: `@Grab('org.apache.groovy:groovy-json:4.0.21')` requires Groovy 4 (3.x uses `org.codehaus.groovy`), yet `README.md:66` still says "Groovy 3.x or newer". Easy fix in passing.

tools/jira/bridge.groovy
oscerd: Fixed — added `--value-json <json>` as an alternative to `--value <str>`. Structured values (priority objects, version arrays, select fields) go through `--value-json` and are parsed before being sent to JIRA. Added two test cases covering object and array payloads.

tools/jira/README.md
oscerd: Fixed — rewrote the Configuration section to explicitly state DC-only scope, and added auth notes explaining Basic vs Bearer with concrete guidance for ASF PAT users.

tools/jira/bridge.groovy
oscerd: Fixed — added `JIRA_AUTH_SCHEME` env var (default `Basic`, accepts `Bearer`). All three auth header sites (`httpGet`, `httpWrite`, `httpMultipart`) now use `"${AUTH_SCHEME} ${API_TOKEN}"`. Added a test that verifies the Bearer header round-trips correctly.

tools/jira/bridge.groovy
oscerd: Fixed in this round — switched to `{"update": {"labels": [{"add": "x"}, {"remove": "y"}]}}`. Eliminates the GET round-trip and the race. Tests updated to assert on the atomic payload shape.

tools/jira/bridge.groovy
oscerd: Added `// DC payload — Cloud uses accountId` inline comment and documented the DC-only limitation in the README assign section.

tools/jira/bridge.groovy
oscerd: Acknowledged — risk is low since source is always local FS and the calling skill controls what gets attached. Good candidate for a follow-up (size cap + filename sanitization) if adoption surfaces real foot-guns.

tools/jira/bridge.groovy
oscerd: Fixed — changed to `f.getText('UTF-8')`. Good catch.

tools/jira/tests/test_bridge_write.py
oscerd: Added auth header value assertions — `test_basic_auth_header` verifies `Basic <token>` round-trips and `test_bearer_auth_scheme` verifies `Bearer <pat>`. Strict mock mode is a good idea for a follow-up; keeping the permissive default for now to keep the harness simple.

oscerd:
@andreahlert Thanks for the thorough review — addressed all three blocking items and most nits in 2dcfadc.

**Retry/backoff:** Agreed this will matter in long skill flows. It applies equally to the read path, so imho better as a cross-cutting follow-up than scoped to this PR.

**Groovy version:** Fixed — README now says "Groovy 4.x" and explains the `org.apache.groovy` group ID.

andreahlert (approved):
Thanks for addressing this. LGTM.

potiuk (changes requested):
Thanks for the iteration — implementation is clean and the prior-feedback
round (Bearer scheme, atomic `update` API, UTF-8 file reads) shows good
attention to detail. Strict key regex (`^[A-Z][A-Z0-9_]*-\d+$`) closes the
URL-injection path, and the 28 pytest tests cover the right happy + error +
auth surfaces.

### Blocking — wire the test suite into CI

The new `tools/jira/` is a proper uv project with tests, but neither the
workflow matrix nor the prek hooks pick it up — so the 28 tests land as
dead weight in CI and a future regression would never block a PR. Two small
additions, both mechanical, mirror the pattern the other tools use:

**1. Add `jira` to the pytest matrix** in `.github/workflows/tests.yml`:

```suggestion
        project:
          - name: oauth-draft
            path: tools/gmail/oauth-draft
          - name: generate-cve-json
            path: tools/vulnogram/generate-cve-json
          - name: skill-and-tool-validator
            path: tools/skill-and-tool-validator
          - name: privacy-llm-checker
            path: tools/privacy-llm/checker
          - name: privacy-llm-redactor
            path: tools/privacy-llm/redactor
          - name: vulnogram-oauth-api
            path: tools/vulnogram/oauth-api
          - name: sandbox-lint
            path: tools/sandbox-lint
          - name: agent-isolation
            path: tools/agent-isolation
          - name: jira
            path: tools/jira
```

**2. Add prek hooks for `jira`** in `.pre-commit-config.yaml`, mirroring the
sibling pattern (the `skill-and-tool-validator` block is the closest analogue
— four hooks: `ruff-check`, `ruff-format`, `mypy`, `pytest`). Each hook's
`files:` regex should match `^tools/jira/(src|tests|pyproject\.toml|bridge\.groovy)`.

Once both land, `pytest (jira)` shows up as a matrix entry under the
`tests-ok` umbrella and prek runs the lint pass locally and in CI.

Happy to apply this as a maintainer-side fixup if you'd rather not chase it
— just say the word. Otherwise, push the two edits when you have a moment
and I'll re-review.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Steward
> maintainer. After you've addressed the points above and pushed an update, an
> Apache Steward maintainer will take the next look at your PR.*
>
> *More on how Apache Steward handles maintainer review:*
> [CONTRIBUTING.md](https://github.com/apache/airflow-steward/blob/main/CONTRIBUTING.md).

oscerd:
@potiuk Done in 5aa58e6:

1. Added `jira` to the pytest matrix in `.github/workflows/tests.yml`
2. Added four prek hooks (`ruff-check`, `ruff-format`, `mypy`, `pytest`) in `.pre-commit-config.yaml` — `files:` pattern includes `bridge.groovy` so tests re-run when the bridge itself changes
3. Added `ruff` and `mypy` to dev dependencies in `pyproject.toml` with mypy config mirroring `sandbox-lint`

All four checks pass locally (ruff check, ruff format, mypy, pytest collect — 28 tests).

potiuk (approved):
LGTM — the CI wire-in addresses the prior REQUEST_CHANGES exactly:

- `.github/workflows/tests.yml` — `jira / tools/jira` matrix entry slots in
  alongside the other tool packages; the `tests-ok` umbrella now gates the
  jira suite alongside the rest.
- `.pre-commit-config.yaml` — four hooks under a `repo: local` block
  (`ruff check`, `ruff format --check`, `mypy`, `pytest`), each with a
  `files:` regex matching `^tools/jira/(src|tests|pyproject\.toml|bridge\.groovy)`
  exactly as proposed.
- `pyproject.toml` — `ruff` and `mypy` added to the `dev` dep group; uv.lock
  reflects.

Verified locally on the rebased branch: `ruff check` clean, `ruff format
--check` clean, `mypy` reports no issues, `pytest` skips the Groovy-dependent
cases (groovy not installed locally) but the test suite's skip semantics
are correct and CI ships groovy. The 28 test assertions still match the
prior round.

The substantive review from earlier still stands: clean Groovy
implementation of six write subcommands; strict key regex (`^[A-Z][A-Z0-9_]*-\d+$`)
closes URL injection; atomic JIRA `update` API for labels; DC-only scope
called out; Bearer auth supported.

Thanks for the quick turnaround on the CI wiring.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Steward
> maintainer. The maintainer approving this PR has read the findings and signed off.
> If something feels off, please reply on the PR and a maintainer will follow up.*
>
> *More on how Apache Steward handles maintainer review:*
> [CONTRIBUTING.md](https://github.com/apache/airflow-steward/blob/main/CONTRIBUTING.md).

PR #346, opened 2026-05-27
feat(pr-management-triage): fetch all PRs upfront, classify in batch

potiuk:
## Summary

- Restructure PR triage from per-page interactive loop to fetch-all → classify-once → present-groups. Step 1 walks every page until `hasNextPage=false` without prompting; Step 2 classifies the full set; Step 3 forms groups that span the entire queue.
- A `mark-ready` group can carry 30+ PRs across what was previously six pages — one screen, one decision.
- Add `Bash(gh api graphql *)` to `permissions.allow` (in `.claude/settings.json`, `tools/sandbox-lint/expected.json` baseline, and the isolation-setup template at `docs/setup/secure-agent-setup.md`) so the read-only fetch loop bypasses the `-F`/`-f` ask rules. Mutations still hit ask.

## Why

The per-page loop forced context-switching between action classes (mark-ready on page 1, draft on page 2, back to mark-ready on page 3, …) and required maintainer attention throughout the long fetch phase. Running the skill against the Apache Airflow queue (200+ open PRs) made the cost visible — maintainer attention was the bottleneck, not GraphQL budget.

Fetching everything up front lets the maintainer walk away during the long fetch, then come back to one batched decision per action class.

## Out of scope

`issue-triage` and `security-issue-triage` would benefit from the same refactor; left untouched here to keep this PR's scope tight.

## Test plan

- [ ] `skill-and-tool-validate` exits 0 (verified locally on the touched files; pre-existing soft warnings in unrelated skills are not introduced here).
- [ ] `pytest (sandbox-lint)` passes — `expected.json` baseline updated in lockstep with `.claude/settings.json`.
- [ ] Run a triage pass on a small queue (e.g. `triage label:area:scheduler`) to confirm the fetch loop ends cleanly and groups present across the result set.
- [ ] Confirm `gh api graphql *` calls during the fetch loop no longer prompt under the new allow rule.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -44,6 +44,9 @@
     }
   },
   "permissions": {
+    "allow": [
+      "Bash(gh api graphql *)"
+    ],
     "deny": [
       "Read(~/.aws/**)",
       "Read(~/.ssh/**)",
--- a/.claude/skills/pr-management-triage/SKILL.md
+++ b/.claude/skills/pr-management-triage/SKILL.md
@@ -48,11 +48,22 @@ of scope here.
 
 This skill is the successor to the triage mode of
 `breeze pr auto-triage`. It drops the full-screen TUI in favour
-of a CLI conversation: PRs are presented to the maintainer one
-*group* at a time (grouped by suggested action), and the
-maintainer either bulk-confirms the group, pulls individual PRs
-out for case-by-case handling, or skips. Detail files in this
-directory break the logic out topic-by-topic:
+of a CLI conversation. The flow is:
+
+1. **Fetch the entire candidate set up front** by paginating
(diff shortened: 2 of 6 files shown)


potiuk:
Nice UX improvement for triage -> instead of paging results, we fetch **all** PRs to memory and pre-classify them before we get triager's attention. This should all get with very little interaction - so when attention of triager is grabbed, there should already be a summary of things to do and groups of changes to act on .. This is a very nice improvement making the triage process far less "attention disrupting".

potiuk:
cc: @paulk-asert -> I am also applying it to "security-issue-triage"  - possibly similar approach could be used for "issue-triage".

PR #348, opened 2026-05-28
feat(pr-management-stats): add deterministic HTML dashboard renderer

choo121600:
## Summary

Adds tools/pr-management-stats/dashboard.py, a directory-portable
Python script that renders the full 11-panel HTML dashboard
described in render.md, without requiring agent execution.

## Why

`reference.py` intentionally stopped at fetch + classify and left
aggregation/rendering to downstream consumers. The README already
described extending the tool in this direction; this PR provides
that deterministic implementation directly.

`dashboard.py` is **directory-portable**: it imports the fetch +
classify primitives from its sibling `reference.py`, so the whole
`tools/pr-management-stats/` directory travels together. It is run as a
script (`python3 dashboard.py …`) — the script's own directory is on
`sys.path[0]`, so the sibling import resolves from any working
directory. It is not a package (the directory name is not a valid module
identifier) and cannot be run with `python3 -m`. The scripts are
stdlib-only; `pytest` is a dev-only dependency for the test suite.

## What changed

- **New**: `tools/pr-management-stats/dashboard.py`
  - Reuses `reference.py` classification + fetch primitives
  - Adds aggregation and HTML rendering logic implementing the
    `render.md` dashboard specification
  - Includes inline SVG/CSS helpers used by the renderer
  - Surfaces incomplete fetches: an INCOMPLETE DATA banner in the HTML
    and `partial: true` in the JSON sidecar
- **Modified**: `tools/pr-management-stats/reference.py`
  - Fixes the `paginated_search` cursor-insertion bug (two consecutive
    `-F` flags silently stopped pagination after page 1) — fixed at the
    source so every consumer benefits
  - Retries transient `5xx` / `RATE_LIMITED` failures with backoff and
    reports a partial-fetch signal
  - `classify(*, partial=False)` makes the reduced closed-PR schema an
    explicit contract; the project-specific defaults (`DEFAULT_TRIAGE_MARKER`,
    `DEFAULT_AI_FOOTER`, `DEFAULT_READY_LABEL`, `DEFAULT_AREA_PREFIX`) are each
    a CLI override (`--triage-marker` / `--ai-footer` / `--ready-label` /
    `--area-prefix`), so an adopter for another project supplies their own
    without editing the file
- **New**: `tools/pr-management-stats/tests/` + `pyproject.toml`
  - pagination (bug regression + retry + partial), aggregations,
    partial-classify contract, render helpers, and an end-to-end
    `reference` ↔ `dashboard` JSON-parity check (all `gh` calls stubbed)
  - wired into prek as the `pytest (pr-management-stats)` hook
- **Modified**: `tools/pr-management-stats/README.md`
  - Documents `dashboard.py`, the test suite, the directory-portable
    model, and the configurable project-specific defaults

image

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] For Python packages touched: `uv run pytest` passes (24 tests, `gh` stubbed)
- [ ] Other:

## RFC-AI-0004 compliance

- [x] **Vendor neutrality** — project-specific defaults are individual `DEFAULT_*` constants, each overridable per invocation via its `--flag` (`--triage-marker` / `--ai-footer` / `--ready-label` / `--area-prefix`) and documented in the README's "Configuration and vendor neutrality" section; placeholders used in tool prose (the `check-placeholders` prek hook is the mechanical gate)

Second commit addresses the review:
- pagination bug fixed at the source (`reference.py`), local override removed;
- the "self-contained / no shared modules" claim corrected to "directory-portable";
- a `tests/` suite added (was the main blocker), including a golden JSON-parity
  test for the reference ↔ dashboard sidecar contract;
- partial-fetch handling (retry + banner + sidecar flag), explicit
  `classify(partial=…)` contract replacing the `setdefault` shim, and the area
  recommendation count switched from total contribs to the untriaged pile.

ruff-format / mypy hooks were intentionally not wired for this tool: it is a
pair of flat scripts rather than a typed src-layout package, so only the pytest
gate is added (noted in `.pre-commit-config.yaml`).

## Known follow-ups (deferred, non-blocking)

Two review nits are intentionally deferred to a follow-up rather than this PR:

- **Eager top-level import of `reference`** (`dashboard.py:76`) — the reviewer
  suggested moving it into `main()` for `--help` robustness. The literal move
  is unsafe here because the module-level aggregation helpers reference the
  imported symbols (`parse_iso`, `is_bot`, `classify`, `COLLAB_ASSOCIATIONS`, …)
  as globals; the clean fix is a qualified-access (`import reference` →
  `reference.parse_iso`) refactor across ~57 call sites, kept out of this diff.
- **Splitting `dashboard.py` into `aggregate.py` / `render.py` / CLI glue**
  (`dashboard.py:1874`) — the reviewer explicitly marked this "Fine as a
  follow-up."

### DIFF
--- a/tools/pr-management-stats/README.md
+++ b/tools/pr-management-stats/README.md
@@ -42,13 +42,28 @@ exists for two reasons:
 ```text
 tools/pr-management-stats/
 ├── README.md     (this file)
-└── reference.py  (Python implementation: fetch + classify + emit intermediates)
+├── reference.py  (Python implementation: fetch + classify + emit intermediates)
+└── dashboard.py  (full HTML render extending reference.py — all 11 panels, self-contained)
 ```
+
+The full HTML dashboard render (per `render.md`) lives in
+[`dashboard.py`](dashboard.py); it imports the fetch + classify
+primitives from `reference.py` and inlines its own SVG / CSS helpers
+so adopters can carry over the whole directory with no external
+dependencies.
 
 ## Invocation
 
 ```bash
+# Reference (fetch + classify + JSON sidecar only)
 python3 tools/pr-management-stats/reference.py \
+    --repo <upstream> \
+    --viewer <maintainer-handle> \
+    --since 2026-04-12 \
+    --out /tmp/dashboard.html
+
+# Full dashboard (all 11 panels rendered as self-contained HTML)
+python3 tools/pr-management-stats/dashboard.py \
     --repo <upstream> \
     --viewer <maintainer-handle> \
     --since 2026-04-12 \
@@ -79,10 +94,11 @@ When the agent invokes the skill, it MUST:
 
 ## Parity implementations
 
-This script is a fetch + classify reference. The full render lives
-in the agent-emitted version per `render.md`. Adopters who want a
-deterministic CI-runnable equivalent should extend this script with
-the aggregation + HTML emission directly; we welcome PRs.
+`reference.py` provides fetch + classify only. `dashboard.py`
+extends it with the aggregation + HTML emission for all 11 panels
+declared in `render.md`, and is the recommended path for CI-rendered
+dashboards. Adopters who want a different language target are
+welcome to add additional parity implementations.
 
 ## Cross-references
 
--- a/tools/pr-management-stats/README.md
+++ b/tools/pr-management-stats/README.md
@@ -5,6 +5,8 @@
 - [pr-management-stats reference implementation](#pr-management-stats-reference-implementation)
   - [Layout](#layout)
   - [Invocation](#invocation)
+  - [Configuration and vendor neutrality](#configuration-and-vendor-neutrality)
+  - [Tests](#tests)
   - [Contract for the agent](#contract-for-the-agent)
   - [Parity implementations](#parity-implementations)
   - [Cross-references](#cross-references)
@@ -41,16 +43,26 @@ exists for two reasons:
 
 ```text
 tools/pr-management-stats/
-├── README.md     (this file)
-├── reference.py  (Python implementation: fetch + classify + emit intermediates)
-└── dashboard.py  (full HTML render extending reference.py — all 11 panels, self-contained)
+├── README.md       (this file)
+├── pyproject.toml  (pins the pytest harness; the tool itself is stdlib-only)
+├── reference.py    (Python implementation: fetch + classify + emit intermediates)
+├── dashboard.py    (full HTML render extending reference.py — all 11 panels)
+└── tests/          (pytest suite: pagination, aggregations, render, JSON parity)
 ```
 
 The full HTML dashboard render (per `render.md`) lives in
 [`dashboard.py`](dashboard.py); it imports the fetch + classify
-primitives from `reference.py` and inlines its own SVG / CSS helpers
-so adopters can carry over the whole directory with no external
-dependencies.
+primitives from `reference.py` and inlines its own SVG / CSS helpers.
+
+The tool is **directory-portable**, not single-file or installable: the
+two scripts plus their resources travel together as one directory, and
+`dashboard.py` imports its sibling `reference.py` by name. Run it as a
+script (`python3 dashboard.py …`) — the script's own directory is on
+`sys.path[0]`, so the sibling import resolves from any working directory.
+Because the directory name (`pr-management-stats`) is not a valid Python
+module identifier, the tool is **not** a package and cannot be run with
+`python3 -m`. The only third-party dependency is `pytest`, and that is
+dev-only (for the test suite); the scripts themselves are stdlib-only.
 
 ## Invocation
 
@@ -83,6 +95,53 @@ The script:
    `is_engaged` requires ANY maintainer touch (issue comment, review,
    review-thread comment, label add, draft conversion).
 5. Writes a JSON sidecar with all the counts that feed the dashboard.
+
+## Configuration and vendor neutrality
+
+The default triage marker, AI footer, ready-label, and area-prefix are
+example values for the reference instance these scripts were built
+against — they are **not** vendor-neutral, and they do not need to be:
+every one is a CLI override, so an adopter for another project supplies
+their own without editing the tool. (The framework's placeholder
+convention governs repo slugs and URLs in prose, not these runtime
+config defaults.) Override per invocation:
+
+```bash
+python3 dashboard.py --repo <upstream> --viewer <handle> \
+    --triage-marker "<your quality-criteria marker>" \
+    --ai-footer "<your bot footer>" \
+    --ready-label "<your ready label>" \
+    --area-prefix "<your area label prefix>"
+```
+
+When pagination is cut short (a `gh` error, a rate limit, or the page
+cap is reached), the run does not silently publish a truncated view: a
+visible **INCOMPLETE DATA** banner is added to the HTML and `partial:
+true` is written to the JSON sidecar. Transient `5xx` / `RATE_LIMITED`
+failures are retried once with backoff before that happens.
+
+## Tests
+
+```bash
+# from the tool directory
+uv run pytest            # or: python3 -m pytest
+```
+
+The suite is stdlib + `pytest` only and stubs all `gh` calls — no
+network access:
+
+- `tests/test_pagination.py` — guards the cursor-pagination fix, the
+  retry/backoff path, and the partial-fetch signal.
+- `tests/test_aggregations.py` — pure aggregation functions
+  (`compute_hero_counts`, `compute_pressure_by_area`,
+  `compute_recommendations`, weekly velocity).
+- `tests/test_classify_partial.py` — the explicit partial closed-PR
+  classify contract.
+- `tests/test_html_render.py` — render helpers, including the
+  incomplete-data banner toggle and HTML escaping.
+- `tests/test_json_parity.py` — runs both `reference.py` and
+  `dashboard.py` over one fixture and asserts the dashboard sidecar is a
+  superset of reference's with identical values on every shared key.
 
 ## Contract for the agent
 
--- a/tools/pr-management-stats/dashboard.py
+++ b/tools/pr-management-stats/dashboard.py
@@ -1,0 +1,1859 @@
+#!/usr/bin/env python3
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+"""
+Full HTML-rendering extension on top of reference.py.
+
+reference.py stops at fetch + classify + JSON sidecar; this script
+reuses its primitives, then computes every aggregate from aggregate.md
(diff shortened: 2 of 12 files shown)


tools/pr-management-stats/dashboard.py
```diff
@@ -0,0 +1,1859 @@
...
+            sys.stderr.write(f"  page {page}: errors {d['errors'][:1]}\n")
+            break
+        nodes = d["data"]["search"]["nodes"]
+        all_nodes.extend(nodes)
+        pi = d["data"]["search"]["pageInfo"]
+        sys.stderr.write(f"  page {page}: +{len(nodes)} (total {len(all_nodes)})\n")
+        if not pi["hasNextPage"]:
+            break
+        cursor = pi["endCursor"]
+    return all_nodes
```
andreahlert: **Blocker (paired with `reference.py:165`).**

The upstream bug this works around is one line. In `reference.paginated_search`:

```python
if cursor:
    cmd.insert(4, "-F"); cmd.insert(5, f"after={cursor}")
```

After both inserts the argv becomes `[..., "-F", "-F", "after=X", "first=N", "-F", "q=Q", ...]` — two `-F` in a row and `gh api graphql` silently stops paginating after page 1. Against `apache/airflow` that means heavily truncated results.

The stated reason for not fixing it upstream ("preserve JSON sidecar parity") doesn't hold IMHO — the JSON shape is unaffected by the subprocess argv order.

Proposal: fix `reference.py:165` to:

```python
if cursor:
    cmd.extend(["-F", f"after={cursor}"])
```

and drop this entire override block, replacing it with `from reference import paginated_search`.

tools/pr-management-stats/dashboard.py
```diff
@@ -0,0 +1,1859 @@
...
+import argparse
+import html
+import json
+import subprocess
+import sys
+from collections import defaultdict
+from datetime import datetime, timedelta, timezone
+from pathlib import Path
+
+from reference import (
```
andreahlert, line 59: **Blocker.** PR body says "no shared modules, no `sys.path` manipulation", but this imports 9 names from `reference`. Only works because CPython sticks the script dir on `sys.path[0]` when you run `python3 dashboard.py`. Running as a module blows up:

```
$ python3 -m tools.pr-management-stats.dashboard --help
ModuleNotFoundError: No module named 'reference'
```

IMHO either reword the PR body to "directory-portable" (accurate), or rename the dir to `pr_management_stats` and use a relative import. Current claim is misleading.

tools/pr-management-stats/dashboard.py
```diff
@@ -0,0 +1,1859 @@
...
+
+    # closed PRs need _is_engaged too for trend coverage — classify with a
+    # minimal shim (no commits / reviewThreads / timelineItems in closed query)
+    for pr in closed_prs:
+        pr.setdefault("isDraft", False)
+        pr.setdefault("commits", {"nodes": []})
+        pr.setdefault("latestReviews", {"nodes": []})
+        pr.setdefault("reviewThreads", {"nodes": []})
+        pr.setdefault("timelineItems", {"nodes": []})
+        classify(pr, ctx)
```
andreahlert: These `setdefault`s are silently papering over the fact that `CLOSED_PRS_QUERY` doesn't return `commits / latestReviews / reviewThreads / timelineItems`, and then we hand the half-populated PR to `classify()`. The moment `classify()` starts reading any new field, closed PRs blow up with a `KeyError` — and nothing here will tell us.

Two cleaner options:

- extend `CLOSED_PRS_QUERY` to actually return those fields, or
- add `classify(pr, ctx, *, partial=False)` in `reference.py` and make the contract explicit.

At minimum, log a warning when we go down the "partial closed PR" path.

tools/pr-management-stats/dashboard.py
```diff
@@ -0,0 +1,1859 @@
...
+        if cursor:
+            cmd.extend(["-F", f"after={cursor}"])
+        r = subprocess.run(cmd, capture_output=True, text=True)
+        if r.returncode != 0:
+            sys.stderr.write(f"  page {page}: error {r.stderr[:300]}\n")
+            break
+        d = json.loads(r.stdout)
+        if "errors" in d:
+            sys.stderr.write(f"  page {page}: errors {d['errors'][:1]}\n")
+            break
```
andreahlert: No retry, no backoff, and on any error we just `break` and write a partial dashboard. Running this on a cron against `apache/airflow` is going to hit a rate limit or a 502 sooner or later and silently publish an incomplete HTML page.

Minimal ask:

- 1 retry w/ small backoff on 5xx / `RATE_LIMITED`
- a `partial: true` flag in the JSON sidecar + a visible banner in the HTML when pagination was cut short

Otherwise we're flying blind when this misfires.

andreahlert (commented):
Overall shape LGTM, dashboard works end-to-end. Three blockers IMHO before merge:

1. `reference.paginated_search` bug worked around instead of fixed — leaves `reference.py` broken for any other consumer. See inline on `dashboard.py:330`.
2. PR body claims "self-contained, no shared modules" but `dashboard.py:49` imports 9 symbols from `reference`. Running as a module breaks.
3. 1859 LOC added, zero tests. Every sibling tool has a `tests/`.

Rest inline. Marked `nit` where non-blocking.

Note: a couple of comments are anchored on `dashboard.py` even though the underlying issue is in `reference.py` — GitHub only lets me inline-comment on files touched by the PR.

tools/pr-management-stats/dashboard.py
```diff
@@ -0,0 +1,1859 @@
...
+        "recommendation_count": len(recs),
+        "pressure_areas": [
+            {"area": a, **v} for a, v in pressure
+        ],
+        "funnel": funnel,
+        "weekly_velocity_totals": {
+            "merged": sum(w["merged"] for w in weekly),
+            "closed_not_merged": sum(w["closed_not_merged"] for w in weekly),
+        },
+    }
```
andreahlert: "Preserved bit-for-bit" is a strong claim for something with no test backing it. First refactor on either side and it drifts without anyone noticing.

A golden-file test would cover it cheaply: run `reference.py` on a fixture, run `dashboard.py` on the same fixture, assert dashboard's JSON is a superset with identical values on the shared keys.

tools/pr-management-stats/README.md
```diff
@@ -42,17 +42,32 @@ exists for two reasons:
 ```text
 tools/pr-management-stats/
 ├── README.md     (this file)
-└── reference.py  (Python implementation: fetch + classify + emit intermediates)
+├── reference.py  (Python implementation: fetch + classify + emit intermediates)
+└── dashboard.py  (full HTML render extending reference.py — all 11 panels, self-contained)
 ```
 
+The full HTML dashboard render (per `render.md`) lives in
+[`dashboard.py`](dashboard.py); it imports the fetch + classify
```
andreahlert, line 54: Since this README is what adopters read first: 1859 LOC added in `dashboard.py` and no `tests/` in this tool. Every other tool in this repo ships tests (`agent-isolation/tests`, `spec-status-index/tests`, `vulnogram/*/tests`, `skill-evals/tests`, `sandbox-lint/tests`, `skill-validator/tests`). The PR template's `uv run pytest` checkbox is unchecked.

IMHO bare minimum to merge:

- `tests/test_aggregations.py` — pure functions, easy to fixture (`compute_hero_counts`, `compute_pressure_by_area`, `compute_recommendations`).
- `tests/test_html_render.py` — snapshot of the HTML for a fixed fixture.
- `tests/test_json_parity.py` — covers the parity claim from the sidecar comment in `dashboard.py`.

tools/pr-management-stats/dashboard.py
```diff
@@ -0,0 +1,1859 @@
...
+    OPEN_PRS_QUERY,
+    classify,
+    compute_codeowners_panel,
+    compute_weekly_velocity,
+    fetch_codeowners,
+    fetch_ready_pr_files,
+    is_bot,
+    parse_iso,
+    weeks_buckets,
+)
```
andreahlert, line 76: nit (related to the imports above): defaults pulled in from `reference` (`DEFAULT_TRIAGE_MARKER = "Pull Request quality criteria"`, `DEFAULT_AI_FOOTER = "AI-assisted triage tool"`, `DEFAULT_READY_LABEL = "ready for maintainer review"`, `DEFAULT_AREA_PREFIX = "area:"`) are very `apache/airflow`-shaped. Configurable via flags, but RFC-AI-0004 "Vendor neutrality" box is unchecked in the PR body. Either tick it with a note in the README, or move the airflow defaults behind a `DEFAULTS_AIRFLOW` profile in `reference.py`.

tools/pr-management-stats/dashboard.py
```diff
@@ -0,0 +1,1859 @@
...
+    print(f"\nDashboard written to {out_path}", file=sys.stderr)
+    print(f"Intermediate state written to {side}", file=sys.stderr)
+    print(json.dumps({k: intermediates[k] for k in (
+        "open_count", "closed_count", "ready_count",
+        "untriaged_count", "untriaged_4w_count", "engaged_count",
+        "ai_triaged_count", "health_rating", "recommendation_count",
+    )}, indent=2))
+
+
+if __name__ == "__main__":
```
andreahlert, line 1874: nit: 1859 LOC in one file with inline CSS f-string (~70 lines), two hand-rolled SVG generators, plus aggregate + render + main. "Directory-portable" doesn't require a single file — splitting into `aggregate.py` / `render.py` / `dashboard.py` (CLI glue) would still be portable and a lot easier to navigate. Fine as a follow-up.

tools/pr-management-stats/dashboard.py
andreahlert, line 76: nit: `reference` is imported eagerly at top level, so any unrelated import error in `reference.py` takes down even `dashboard.py --help`. Moving the import inside `main()` keeps the CLI surface robust.

tools/pr-management-stats/dashboard.py
```diff
@@ -0,0 +1,1859 @@
...
+        area, v = pressure[0]
+        if v["u4w"] + v["u14w"] >= 5:
+            recs.append(
+                {
+                    "priority": "medium",
+                    "icon": "📍",
+                    "title": f"Area \"{area}\" has {v['contribs']} contributor PRs ({v['u4w']} untriaged >4w)",
+                    "detail": "One area dominating the untriaged queue — scoped pass clears bulk.",
+                    "action": f"/pr-management-triage label:area:{area}",
+                    "count": v["contribs"],
```
andreahlert: nit: the recommendation uses `count: v["contribs"]`, but the actual signal driving the rec is `u4w` (untriaged > 4 weeks). Areas with lots of contribs but a shallow untriaged pile will look more urgent than they are. Probably want `count: v["u4w"]`.

tools/pr-management-stats/dashboard.py
choo121600: Fixed at the source:
`cmd.extend(["-F", f"after={cursor}"])`,
override block removed, now importing `paginated_search` from `reference`.

Regression test in `test_pagination.py` asserts clean argv generation and no consecutive `-F`.

tools/pr-management-stats/dashboard.py
choo121600: Reworded to `"directory-portable"` across the PR body, docstring, and README.
As you noted, the directory name and `sys.path[0]` sibling-import approach remain unchanged.

tools/pr-management-stats/dashboard.py
choo121600: Went with the explicit contract approach:
`classify(*, partial=False)`,
`isDraft` is now selected in `CLOSED_PRS_QUERY`,
and the partial path logs a warning.

The `setdefault` shim is gone.
Covered by `test_classify_partial.py`.

tools/pr-management-stats/dashboard.py
choo121600: Added:
- 1 retry with backoff on 5xx / RATE_LIMITED
- `partial: true` in the sidecar
- `INCOMPLETE DATA` banner in the HTML

choo121600:
@andreahlert Thanks! Looks like most of the issues are resolved now. 🙏

potiuk:
Indeed !

PR #353, opened 2026-05-28
add good first issue skill

justinmclean:
## Summary

- Extracts pure helper functions from `render.py` into `_render_helpers.py`
  (YAML parser, `deep_merge`, `parse_dt`, bucket functions, `eval_predicate`,
  `is_bot_body`, JS serialisers) so they can be imported and tested without
  triggering `render.py`'s file-I/O module-level code.
- Adds `pyproject.toml` + `uv.lock` to give the tool a proper uv project
  (dependency-only; `tool.uv.package = false` — no wheel).
- Ships `tests/`: 78 unit tests for the helper functions and 12 integration
  tests that run `render.py` end-to-end against fixture cache data in
  `tests/fixtures/`.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:


potiuk:
Hmmm. I think those "first time issues" with all details explained might be a bit too "easy" - the two fixes we got are most likely juist bots picking them up -  My thinking with such issues is that they could be rather generally described without going to too much details, because we need some "input" from the people ? 

but maybe that's a bit to cautious?

potiuk (approved):

choo121600:
I do have some concerns, but overall I think it’s probably okay.

I think good first issues are primarily meant for onboarding new contributors. So if maintainers use Magpie to generate easy and detailed good first issues, then another contributor simply feeds that directly into their own agent prompt to solve it, and finally the maintainers review it again, the whole flow feels a bit awkward to me.

That said, I also think this kind of flow is difficult to prevent in practice.

choo121600 (approved):

justinmclean:
People are already doing this, using AI to submit PRs. It reminds me of a Paul Kelly song: "from little things big things grow." Contributors have to start somewhere in open source, and if there aren't easy tickets, they have nowhere to begin.

I tried a similar approach on another project and it attracted 250+ contributors. Most were drive-by, but a few stayed, a couple became committers, and one ended up with 100+ PRs. Yes, some will be bots, but the same flow is what surfaces the humans who'd otherwise never have started.

choo121600:
We were actually able to attract a new contributor to our project through this issue creation flow.
So I definitely think it can be quite effective :)

potiuk:
> We were actually able to attract a new contributor to our project through this issue creation flow. So I definitely think it can be quite effective :)

Agree. This is most important :)

potiuk:
> I tried a similar approach on another project and it attracted 250+ contributors. Most were drive-by, but a few stayed, a couple became committers, and one ended up with 100+ PRs. Yes, some will be bots, but the same flow is what surfaces the humans who'd otherwise never have started.

And we have Magpie to deal with it :)

justinmclean:
I also note it gives us PRs, etc we can test skills on. I also note it doesn't find a good first issue; just create one on X and validates that it is a good issue.

potiuk:
are you undrafting it @justinmclean ? Happey to merge it :)

choo121600:
COOL :)

PR #370, opened 2026-05-29
skill-evals: field-aware grading, intersection key matching, wrap non-JSON output

justinmclean:
## Summary

The exact-equality comparator in `tools/skill-evals` was rejecting
candidate answers that carried the right decision but differed in
wording, and was failing whole cases on schema details the model
chose not to emit. This change rewrites comparison around three
ideas, all default-on with `--exact` as the opt-out.

1. Field-aware grading. Decision fields (booleans, enums, counts,
   ordering, IDs) stay on exact equality. Free-text fields
   (rationale, reason, drop_reason, blockers, notes, summary,
   explanation, details, description) are sent to a cheap judge
   model with a fixed rubric and parsed back as
   `{match, reason}`. Default grader is `claude -p --model haiku`;
   override with `--grader-cli "<command>"`. Per-fixtures-dir
   `grading-schema.json` overrides the prose-field set.

2. Batched grader call. All prose-field mismatches for a single
   case are collected and sent in one rubric prompt, so a case
   with N wording diffs costs one Haiku call, not N. Decision-
   field failures skip the grader entirely.

3. Intersection-only key matching. Only keys present in both
   actual and expected are asserted. Extra keys in the model
   output are ignored; keys declared in expected that the model
   did not emit are skipped, not failed. expected.json describes
   what the answer should say where it speaks, not a required-
   keys schema. Suite authors should keep expected.json focused
   on the keys that carry the eval's signal.

Non-JSON model output and non-zero CLI exits are now wrapped
silently as `{raw_output, ...}` and fed to the same intersection
comparator. Cases that produce a prose refusal still pass unless
expected.json declares a `raw_output` key. Statuses remain
PASS / FAIL / ERROR / MANUAL; the wrap is an implementation
detail. `--exact` mode preserves the previous strict behavior end
to end.

Test coverage: 40 cases across the new helpers
(`collect_diffs`, `batch_grade_prose_fields`, `compare_with_grader`)
and the end-to-end `--cli` + `--grader-cli` path, plus three small
grader stub scripts under `tests/`. README updated.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

### DIFF
--- a/tools/skill-evals/README.md
+++ b/tools/skill-evals/README.md
@@ -92,6 +92,58 @@ automatic JSON-equality comparison is meaningless. Those cases report
 `MANUAL` and the runner skips the CLI call; review them by re-running
 without `--cli` (or with `--verbose`).
 
+### Field-aware grading (default in `--cli` mode)
+
+Pure JSON-equality on `expected.json` is too strict for free-text fields
+like `rationale`, `reason`, `drop_reason`, and `blockers`: a candidate
+answer can carry the right decision but be flagged FAIL on wording
+alone. By default, `--cli` mode now sends those fields to a cheap judge
+model and grades them by meaning instead of by string equality.
+
+```bash
+# Decision fields exact; prose fields graded by the default Haiku judge.
+PYTHONPATH=tools/skill-evals/src python3 -m skill_evals.runner --cli "claude -p" \
+    tools/skill-evals/evals/issue-triage/
+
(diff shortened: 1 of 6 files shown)


potiuk:
NICE! Step closer to reproducibility of evals across LLMs but also for repeatable tests :)

PR #456, opened 2026-06-05
feat(agent-isolation): add opt-in terminal-tint helper for waiting state

potiuk:
## Summary

- Ships `tools/agent-isolation/claude-term-bg.sh`, an **opt-in quality-of-life helper (not a security control)** that keeps a calm baseline terminal background and tints it **only when Claude genuinely wants the operator to act** — never while it is working.
- Two states across five user-scope hooks: `Stop` → tint (your turn); `Notification` → tint *only* for permission prompts (calm for the plain idle ping); `PreToolUse` + `UserPromptSubmit` + `SessionStart` → calm (working, after a reply, fresh session — the last also clears a stale tint a prior session left in the terminal).
- Why it needs care: hooks run with **no controlling terminal**, so the helper walks the process tree to the Claude pty to write the OSC escape; and iTerm2 honours OSC 11 (set) but not OSC 111 (reset), so a deterministic reset needs an explicit colour via `CLAUDE_RESET_BG`.

## Type of change

## Test plan

- [x] `prek run --all-files` passes (doctoc, markdownlint, typos, check-placeholders, skill-and-tool-validate)
- [x] `bash -n` clean on the script
- [x] Manually verified on iTerm2 + macOS: background tints at end-of-turn and on a permission prompt; returns to a deterministic black while working, after a reply, and on a fresh session; stale tint from a closed session is cleared on `SessionStart`.

## RFC-AI-0004 compliance

- [x] **HITL** — the install step is explicitly optional and **default-off**; the operator is asked during setup before any hook is wired.
- [x] **Sandbox** — no new host access; the helper only writes an OSC escape to its own terminal device.
- [x] **Write-access discipline** — no outbound messages; purely local terminal cosmetics.

## Linked issues

(none)

## Notes for reviewers

This is deliberately a **UX convenience, not a security mechanism**, and the doc/README say so in several places — it rides the existing user-scope-hook install machinery (same as the sandbox status line), which is why it lives under `agent-isolation/` and in `secure-agent-setup.md`. If you'd prefer it housed outside the security-focused setup surface, happy to relocate it.


potiuk:
This is really nice opt-in :)... Shows which terminal needs your attention.


Screenshot 2026-06-05 at 18 32 20

PR #467, opened 2026-06-08
feat(spec-loop): scope update runs incrementally from a .last-sync marker

justinmclean:
## Summary

The update beat re-audited every skill, tool, and modes.md row on every
run, which was the bulk of the streaming volume on a slow link. This adds
a tools/spec-loop/.last-sync marker holding the BASE SHA the specs were
last synced against, and an update_scope_context step that tells the agent
to diff only the paths changed since that SHA and skip everything else. If
the diff is empty the run exits without creating a branch or commit.

loop.sh owns the marker: after a successful sync it advances .last-sync to
the current BASE HEAD, amending it into the agent's sync commit when there
is one, or making a marker-only branch off BASE when there was nothing to
sync. PROMPT_update.md is updated to honor the incremental scope and to
leave the marker file alone, since loop.sh maintains it.

Also skip the open-PR context (a gh network round-trip) in update mode,
which never picks a work item, and add SPEC_LOOP_OUTPUT_FORMAT (default
text) to switch the agent to stream-json with --verbose when debugging a
wedged run.

## Type of change

- [ ] Groovy reference impl
- [X] Other: tool update

## Test plan

- [ ] Other:


justinmclean:
Looks like the CI failure is a download issue and not a real issue - can someone rerun the CI for me?

potiuk:
Rebased + merged . Good one.

PR #471, opened 2026-06-08
fix(gmail): guarantee plain-text drafts on both backends

potiuk:
## What

Make plain-text-only an enforced, documented invariant for Gmail drafts created by either backend (`oauth_curl` and the claude.ai Gmail MCP).

## Why

The project's outbound correspondence (security replies, ASF-security relays, advisory text) is plain text by policy. Nothing previously *guaranteed* that — the oauth backend happened to emit plain text but had no guard against a regression, and the MCP path will emit an HTML part if `htmlBody` is ever passed.

## Changes

- **`oauth_curl`** (`tools/gmail/oauth-draft/src/oauth_draft/create_draft.py`): comment on `build_mime` locking the intent — `set_content(str)` yields a single `text/plain` part; never add an HTML alternative.
- **Test** (`tests/test_create_draft.py`): new `test_build_mime_is_plain_text_not_html` asserting the built MIME is single-part `text/plain`, not multipart, with no `text/html` anywhere and no `text/html` in the raw bytes.
- **`claude_ai_mcp`**: documented the hard rule — pass only `body`, never `htmlBody` (which adds a `text/html` part) — in `operations.md` (call-shape example + bullet + "Hard rules that apply to both backends") and `draft-backends.md` ("Limitations that apply to both backends").

Docs + one Python package. `prek`-equivalent hooks ran on commit and pass: ruff, mypy, pytest (60 tests), markdownlint, typos, check-placeholders.

### DIFF
--- a/tools/gmail/draft-backends.md
+++ b/tools/gmail/draft-backends.md
@@ -188,6 +188,15 @@ are involved (either backend); always do the per-thread check too.
 
 ## Limitations that apply to both backends
 
+- **Plain text only — never HTML.** Both backends produce plain-text
+  (`text/plain`) drafts, and must keep doing so. `oauth_curl` builds a
+  single `text/plain` part via `EmailMessage.set_content` (its test
+  suite asserts the message stays single-part `text/plain` with no
+  `text/html` alternative). For `claude_ai_mcp`, this is guaranteed by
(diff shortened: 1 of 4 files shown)


potiuk:
It turned out that this is the best way to guarantee Gmail not overriding the URLs.

PR #484, opened 2026-06-10
Fix setup-status eval prompts to surface decision rules

justinmclean:
## Summary

Add eval tests and discovered some bugs.

The step-1-command and step-3-adjust-decision evals extracted skill
sections that did not contain the rules for the fields under test, so
the model guessed instead of following the skill:

- step-1-command anchored on "Step 1 - Render the dashboard", which
  never mentions the flags; repoint it at "Inputs" (the flag table)
  and state in output-spec that no_adjust is independent of --format.
- step-3-adjust-decision could not see the command-mapping rules
  (a peer "Step B" section) or the --no-adjust short-circuit (an intro
  above the anchor). Nest the mapping under Step A, renumber the old
  Step B/C, and add the short-circuit at the top of Step A.

Document the single-section extraction behaviour in the eval README
so future steps anchor step_heading at the section holding the rules.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

### DIFF
--- a/skills/setup-status/adjust.md
+++ b/skills/setup-status/adjust.md
@@ -14,6 +14,11 @@ runs it.
 
 ## Step A — Detect the deltas
 
+**Short-circuit first:** if `no_adjust` is set, the adjust flow is
+skipped entirely — return `offer_adjustments: false` with empty
+`deltas` and `delegated_commands`, and run no detection at all.
+Only when `no_adjust` is unset do you evaluate the table below.
+
 From the collected JSON, surface each of these that applies:
 
 | Delta | Signal in the JSON |
@@ -29,7 +34,7 @@ Order the offers most → least impactful (drift and dangling links
 before optional family additions). If no delta applies, say the
 adoption is fully wired and stop — do not invent work.
 
-## Step B — Map each delta to a `/magpie-setup` command
+### Map each delta to a `/magpie-setup` command
 
 | Adjustment | Delegated command |
 |---|---|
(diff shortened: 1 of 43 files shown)


justinmclean:
Run teh review skill on this:


## Correctness

[blocking] step-2-present/fixtures/output-spec.md:17 — the output-spec, which is appended to the model's system prompt, literally states the answer: "Per the skill's OUTPUT CONTRACT, the correct answer is always verbatim." Cases 2 and 3 are supposed to test whether the model resists pressure to paraphrase; with the answer in-context they pass-stamp regardless of whether the skill's OUTPUT CONTRACT actually reached the model. Remove that line (the field-semantics lines above it are fine).
[advisory] step-1-command / step-2-present anchors verified correct — the OUTPUT CONTRACT blockquote and Step-A rules fall inside the extraction window, so the skill rule does reach the model. The leak above is the only wiring problem.

## Conventions

[blocking — verified] Malformed SPDX header in fixture report.md files. The header opens with <\!-- (literal backslash) instead of <!-- — confirmed in the raw bytes via cat -A. That is not a valid comment opener, so a license scanner keying on a proper comment may not recognise it. Affects ~13 fixture report.md files across all four steps. Likely a fixture-generation escaping bug; fix the generator/template, not each file by hand.
[advisory] evals/setup-status/README.md uses uv run --project tools/skill-evals — your convention is --directory (per your own runner notes). Will fail if copy-pasted.

Some more work needs to be done here.

potiuk (approved):
Nice, more evals.

potiuk:
But I will wait for confirmation on that one @justinmclean

potiuk:
Nice!

PR #488, opened 2026-06-11
fix(skills): stabilize setup-verify eval and extend coverage

justinmclean:
## Summary

Follow-up to the heading fix, which let the eval run and exposed
latent fixture failures.

- step-1-classify: add grading-schema.json marking 'evidence' as a
  prose field so equivalent wording is graded on conclusion, not
  verbatim string match (evidence is not in the runner default set).
- case-3 / case-6: correct check status from X to partial for a
  missing PostToolUse/sandbox-error-hint.sh, matching the documented
  'missing error-hint reports partial, not missing' rule in SKILL.md.
- step-2 case-4: broaden expected follow_up reason to span the same
  scope as a correct answer (.git/HEAD read failure consequence plus
  helper remediation), consistent with case-3/case-5 style.
- Extend check 1 to enumerate the sandbox.filesystem allowlist
  (allowRead/allowWrite) in both SKILL.md and the cited canonical
  doc, so the model reports it consistently.

Validation: eval suite fully green (11/11)

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

### DIFF
--- a/docs/setup/secure-agent-setup.md
+++ b/docs/setup/secure-agent-setup.md
@@ -1747,8 +1747,9 @@ below and report ✓ done / ✗ missing / ⚠ partial, with the evidence
 
 1. Project `.claude/settings.json` exists and has
    `sandbox.enabled: true`, the `permissions.deny` block, the
-   `permissions.ask` block, and the
-   `sandbox.network.allowedDomains` block.
+   `permissions.ask` block, the
+   `sandbox.network.allowedDomains` block, and the
+   `sandbox.filesystem` allowlist (`allowRead`/`allowWrite`).
 2. User-scope `~/.claude/settings.json` has the `PreToolUse`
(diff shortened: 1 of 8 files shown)


justinmclean:
## Correctness

[advisory] step-2-recommend/case-4-project-root-missing/expected.json — reason was expanded, but no grading-schema.json was added/changed for step-2-recommend. If the runner exact-matches undeclared string fields, case-4 will fail. Confirm step-2-recommend grades reason as prose (or add a grading-schema there). The new step-1-classify/grading-schema.json only covers step 1.
[advisory] case-6-injection-attempt status ✗→⚠ now matches the SKILL rule (missing sandbox-error-hint.sh is ⚠, not ✗) — correct.

## Security / Conventions

no findings.

potiuk (approved):

PR #493, opened 2026-06-11
feat(validator): add ASF-coupling advisory lint

justinmclean:
## What

Adds a new SOFT advisory lint to `skill-and-tool-validator` that flags
ASF-coupled tokens in skill bodies, tokens a non-ASF adopter cannot satisfy
without editing the skill (e.g. `svn` commands, `announce@apache.org`,
Vulnogram URLs, bare PMC/ICLA/incubator mentions).

## Why

Skills are meant to be project-agnostic, but ASF-specific assumptions leak
into prose over time. This surfaces them so maintainers can generalise the
coupling (or consciously keep the ASF default) before it becomes a hidden
adoption blocker.

## How

`validate_asf_coupling()` scans skill bodies against a tiered pattern list:

- **high** confidence: almost never legitimate outside an ASF workflow
  (svn release commands, hardcoded announce list, dist/dev paths).
- **low** confidence: common ASF prose that may be intentional
  (bare PMC, ICLA, incubator).

Each hit is tagged with a confidence level and a remedy class
(placeholder / adapter / capability-flag) so maintainers know how to
generalise it. The rule is advisory only: it surfaces on stderr and
never fails the run. It reuses the existing allowlist-path and
inline-allow-marker machinery, plus extra markers for lines that already
name a generalisation mechanism (`release-dist-backend`, `<announce-list>`,
`ASF default:`, etc.).

## Testing

New `TestValidateAsfCoupling` class, 17 cases covering high/low patterns,
remedy-class reporting, allowlisted paths, and inline-allow-marker
suppression. All pass:

    uv run --directory tools/skill-and-tool-validator --with pytest \
      pytest tests/test_validator.py::TestValidateAsfCoupling -v

Ran against the repo's 42 SKILL.md files as a sanity check: 77 advisory
hits (1 high, 76 low), all behaving as designed. The single high hit is a
hardcoded `announce@apache.org` in `security-issue-sync` (left as a
follow-up, not in scope here).

## Type of change

- [ ] Groovy reference impl
- [X] CI / dev loop (`prek`, workflows, validators)
- [ ] Other:

## Test plan

- [ ] Other:

### DIFF
--- a/tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
+++ b/tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
@@ -17,7 +17,7 @@
 
 """Validate framework skill definitions.
 
-This module validates nine aspects of every skill under
+This module validates ten aspects of every skill under
 skills/:
 
 1. YAML frontmatter — every SKILL.md must have a valid frontmatter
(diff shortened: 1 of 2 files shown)


potiuk:
Good idea. Let me rebase that one too after the merges.

PR #509, opened 2026-06-12
feat(skill): add issue-stale-sweep skill with eval suite

justinmclean:
## Summary

Adds the `issue-stale-sweep` general-issue triage skill, closing the stale-handling gap noted in the triage-mode spec Known Gaps. The skill sweeps open issues dormant past configurable warn/close/hard-close thresholds, classifies each as REQUEST-UPDATE or CLOSE-STALE, drafts a comment for each candidate, and posts only after per-item maintainer confirmation. CLOSE-STALE items require a mandatory two-step confirmation before the issue is actually closed. Security-signal issues are skipped and surfaced for private routing. Comes with the stale-sweep-config.md template, symlinks, capability-map entry, and 22-case eval suite.

## Type of change

- [X] Skill change (`.claude/skills/<name>/`) — eval fixtures updated below
- [ ] Tool / bridge contract (`tools/<system>/*.md`)
- [ ] Python package (`tools/*/` with `pyproject.toml`)
- [ ] Groovy reference impl
- [ ] Cross-cutting (RFC, AGENTS.md, sandbox, privacy-LLM)
- [ ] Documentation (`docs/`, `README.md`, `CONTRIBUTING.md`)
- [ ] Project template (`projects/_template/`)
- [ ] CI / dev loop (`prek`, workflows, validators)
- [ ] Other:

## Test plan

- [X] `prek run --all-files` passes
- [ ] For Python packages touched: `uv run pytest` / `ruff check` / `mypy` passes
- [ ] For Groovy bridges touched: command-line invocation tested end-to-end
- [X] For skill changes: eval suite passes for the affected skill
      (`PYTHONPATH=tools/skill-evals/src python3 -m skill_evals.runner tools/skill-evals/evals/<skill>/`)
- [ ] For skill *behaviour* changes: a new or updated eval fixture is included in this PR
      (a regression test for the bug fixed / the behaviour added — see CONTRIBUTING.md)
- [ ] Other:

PR #533, opened 2026-06-20
feat(skill): add release-vote-tally skill with eval suite

justinmclean:
## Summary

Adds the `release-vote-tally` skill (`magpie-release-vote-tally`), which a
Release Manager invokes after an RC's approval window closes. It fetches the
approval signal for the RC, classifies each reply as +1 / 0 / -1 and binding
or non-binding against the configured roster, tallies the result against the
`dev-list-vote` ASF baseline (binding +1 >= 3 and binding +1 > binding -1),
and drafts the `[RESULT] [VOTE]` email. It never sends mail and never applies
a label without explicit RM confirmation, and it blocks in pre-flight if the
vote window has not yet elapsed.

### Changes

- `skills/release-vote-tally/SKILL.md` — the skill (Triage mode,
  `capability:triage`), with installed copies under `.agents/`, `.claude/`,
  and `.github/`.
- `docs/labels-and-capabilities.md` — maps the skill to `capability:triage`.
- `tools/skill-evals/evals/release-vote-tally/` — eval suite of 9 cases
  across three steps: pre-flight (window-elapsed / mechanism-mismatch),
  classify (binding vs non-binding, fractional, ambiguous/conditional), and
  tally (passed, failed, and a prompt-injection-in-vote-thread safety case).
  Steps 2 and 3 ship `assertions.json` so prose and decision fields are
  auto-graded rather than reviewed by hand.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

### DIFF
--- a/.agents/skills/magpie-release-vote-tally
+++ b/.agents/skills/magpie-release-vote-tally
@@ -1,0 +1,1 @@
+../../skills/release-vote-tally
--- a/.claude/skills/magpie-release-vote-tally
+++ b/.claude/skills/magpie-release-vote-tally
@@ -1,0 +1,1 @@
+../../skills/release-vote-tally
--- a/.github/skills/magpie-release-vote-tally
+++ b/.github/skills/magpie-release-vote-tally
@@ -1,0 +1,1 @@
+../../.agents/skills/magpie-release-vote-tally
--- a/docs/labels-and-capabilities.md
+++ b/docs/labels-and-capabilities.md
@@ -153,6 +153,7 @@ Capabilities for every skill currently in
 | `security-issue-import-from-scan` | `capability:intake` |
 | `security-issue-sync` | `capability:intake` *(+ `capability:reconciliation` once [#337](https://github.com/apache/airflow-steward/issues/337) lands the ASF-dashboard step)* |
 | `setup-shared-config-sync` | `capability:intake` + `capability:setup` *(reconciles user-scope config to a sync repo; the act is intake, the subject is setup)* |
+| `release-vote-tally` | `capability:triage` *(reads the vote thread / approval signal, classifies each reply as binding or non-binding, tallies the result, and drafts the `[RESULT] [VOTE]` email for RM review — triage over the vote-thread queue)* |
 | `release-announce-draft` | `capability:resolve` *(drafts the `[ANNOUNCE]` email and opens the site-bump PR that complete the release lifecycle)* |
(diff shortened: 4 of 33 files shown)


justinmclean:
## Correctness
No findings.

Cross-checked the eval suite against the harness rather than trusting the fixtures at face value: all 9 expected.json parse, every step-config.json step_heading matches a real heading in SKILL.md (so the harness extracts the right step), and the grading wiring is right — step-2 declares its non-default prose fields in grading-schema.json (similarity_summary, closing_comment, cross_ref_comment, injection_warning), step-0's blockers is already in the harness DEFAULT_PROSE_FIELDS so it grades soft without a schema, and step-3's action lists are exact-matched, which is what you want for deterministic output. All three relay symlinks resolve.

## Security
No findings.

The skill makes every state change a confirmed proposal (Golden rules 1–2) and the adversarial eval (step-2 case-2-injection-in-body) correctly treats an injected HTML-comment directive as data: sets injection_warning, keeps proposed: true, and doesn't let it swap the kept/duplicate assignment. No credentials, injection, or untrusted-deserialization patterns introduced.

## Conventions

[advisory] docs/mode-economics.md (not in diff) — new skill not added to the Triage cost table.

AGENTS.md → "Keeping evals and mode-economics in sync" lists "a new skill" as a trigger; the table already carries security-issue-deduplicate.

Rule: AGENTS.md#keeping-evals-and-mode-economics-in-sync. Note: the preceding sibling issue-stale-sweep (#509) also omitted it and the table is non-exhaustive, so this is precedent-consistent — optional.

[advisory] skills/issue-deduplicate/SKILL.md:397 — reference link text doesn't match its target.

- [docs/triage/spec.md](../../tools/spec-loop/specs/triage-mode.md)

Rule: link label should name the real path; docs/triage/spec.md doesn't exist (target tools/spec-loop/specs/triage-mode.md does, so lychee passes but the label misleads).

[advisory] skills/issue-deduplicate/SKILL.md:405 — link points at the directory, not the file.

[<project-config>/issue-tracker-config.md](../../projects/_template/)

Rule: link the file (.../projects/_template/issue-tracker-config.md).

[advisory] commit 81cc0f5 "fix tests" — vague subject and no Generated-by: trailer.

AGENTS.md commit conventions: keep the message focused on the user-visible change; agent-generated commits carry a Generated-by: trailer. Moot if you squash-merge (the lead commit has the trailer and no Co-Authored-By:, which is the rule that matters).

PR #543, opened 2026-06-21
feat(skill): add release-rc-cut skill with eval suite

justinmclean:
## Summary

Adds the `release-rc-cut` skill (Drafting mode) covering Steps 4 to 5 of the
release-management lifecycle: it emits paste-ready command sequences to tag an
RC, build artefacts, sign each one, generate checksums, and stage to the
adopter's distribution backend. The skill runs nothing locally and never
handles the RM's signing key. Registered in `.agents` / `.claude` / `.github`
skill dirs and in the labels-and-capabilities table (`capability:resolve`).

Ships a 9-case behavioral eval suite across three steps (pre-flight, tag/build/
sign/checksum, staging), plus a small grading enhancement to the eval runner.

## Skill behaviour highlights

- Golden rules enforce: agent never runs commands locally, never touches the
  signing key, SHA-512 only by default (SHA-256 when configured, MD5/SHA-1
  never), every state change is a proposal, and a `dist/release/` denylist
  (RC staging is `dist/dev/` only).
- External content (planning issues, build config) is treated as data, with an
  adversarial eval case for prompt injection.

## Eval runner change

- Deterministic assertions now honor `"negate": true`, so a predicate can
  assert the *absence* of a match (`has_no_passphrase_arg`, `has_no_md5sum`,
  `has_no_sha1sum`). A spec error (`None`) passes through unflipped, so a real
  `--passphrase` leak is still caught. Covered by two new unit tests.

## Eval fixtures

- `has_sha256sum` lives in the suite-level `assertions.json` (the runner reads
  assertions only from the suite `fixtures/` dir; per-case files are ignored).
  Removed five dead per-case `assertions.json` files.
- Added `grading-schema.json` marking `section_2_build_command` as a prose
  field, since the model may emit the build command bare or wrapped with the
  spec's `# Run at the release tag <version>-<rcN>` comment header. Both are
  correct, so exact match was brittle.
- README run commands corrected (`--directory` +

## Type of change

- [X ] Skill change (`.claude/skills/<name>/`) — eval fixtures updated below
- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:


justinmclean:
## Correctness

[advisory] tools/skill-evals/src/skill_evals/runner.py:707 — negate is read with a truthiness check, not a strict boolean.

if spec.get("negate") and holds is not None:

Rule: a stray non-bool value (e.g. "negate": "false") would silently invert. load_assertions validates type but not negate. Low risk given fixtures are the only source, but a bool guard in load_assertions would fail typos loudly, consistent with how unknown predicate types are already rejected.

[advisory] step-3-staging fixtures — the dist/release/ denylist check is vacuous. staging_url_not_dist_release and proposed_true lack a has_/mention_ prefix, so they route to the field-aware comparator and are intersection-skipped (the model output has no field of those names). The injection case therefore passes without actually verifying the security property. Rename to has_-prefixed keys (now that negate works) to exercise it.

## Security

No findings in the changed code. The skill is read-only by design (golden rules 1 to 5: no local execution, no key handling, SHA-512/256 only, proposal-gated writes, dist/release/ denylist), and the runner change is grading-only. The negate refactor preserves fail-loud behaviour: a spec error returns None and is passed through un-inverted (covered by test_assert_negate_passes_spec_error_through), and a real --passphrase leak still evaluates to False (covered by test_assert_negate_inverts_regex_result).

## Conventions
No blocking findings. SPDX headers present on new .md files; JSON fixtures are comment-incapable and correctly exempt. The negate feature is documented in both the module docstring and the eval README.

potiuk:
All those PRs need rebase :(

PR #592, opened 2026-06-27
docs: remove legacy project name details from the readme

johnslavik:
I figured that the reader might be interested in what Magpie is more, rather than how it got its current name

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,27 +20,6 @@
 
 # Apache Magpie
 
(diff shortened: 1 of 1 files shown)


johnslavik:
Outside of README there's [a couple more](https://github.com/search?q=repo%3Aapache%2Fmagpie+steward&type=code) occurrences to remove later

Issue #320, opened 2026-05-26
feat(adapter/amazon-q): Amazon Q Developer CLI skill runtime


potiuk:
**Goal:** Bring Amazon Q Developer CLI (`q`) to feature parity with Claude Code as a skill runtime. Grounds: [RFC-AI-0004 Principle 3 — Vendor neutrality](../tree/main/docs/rfcs/RFC-AI-0004.md).

**Why Amazon Q:** Default agentic CLI in many AWS-shop environments — corporate adopters, federal / regulated environments where Amazon Q is the only sanctioned LLM tool. Several ASF projects with AWS-aligned committer bases have asked about Q compatibility. Coverage matters less for individual maintainers, more for *organisational* adoption.

**What "parity" means:**

- Skills under `.claude/skills/<name>/SKILL.md` are invokable from `q` (Q Developer CLI)
- The `tools/*` bridges are reachable — `q` supports custom tools via its agent configuration
- Sandbox parity: `q` has its own command-permission model (allow-lists, "trust" prompts) — document the mapping
- HITL primitives: `q` supports interactive confirmation for tool use — verify and document

**Suggested approach:**

- Map `SKILL.md` → Q's agent configuration format (`~/.aws/amazonq/agents/<name>.json` + system prompts)
- Document skill-invocation pattern
- Map sandbox parity — `q` allowlists are the native fit for the framework's command-permission discipline

**Reference:**

- RFC-AI-0004: [`docs/rfcs/RFC-AI-0004.md`](../tree/main/docs/rfcs/RFC-AI-0004.md)
- Amazon Q Developer CLI: https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/command-line.html
- Q agents: https://docs.aws.amazon.com/amazonq/latest/qdeveloper-ug/cli-agents.html
- Existing skill shape: [`.claude/skills/`](../tree/main/.claude/skills/)

Alexhans:
Happy to help in this space.  I benchmark/eval both coding harness agents and SDK agents and kiro-cli (and previously q-cli) are ones we use extensively and have access to.  

Some easy tricks to get progressive disclosure are using a `skills` folder as the actual source for all skills and creating symlinks for the different vendor specific hidden folders such as `.kiro` , `.claude`, like so:

```                                                                                                                                                      
my-project/                                                                                                                                              
├── skills/                     # Canonical source                                                                                                       
│   ├── my-skill/                                                                                                                                        
[...]

PR #648, opened 2026-06-30
fix(setup): remove stray in-source skill symlinks

Alexhans:
## Summary

Remove three stray, self-referential symlinks committed inside skill source
dirs. Each resolves back to its own directory, forming a filesystem cycle; the
canonical `.agents/skills/magpie-<skill>` links already provide discovery, so
these in-source links are redundant — the same self-adoption-drift class fixed
in #516 and #599.

- `skills/good-first-issue-sweep/good-first-issue-sweep` → `../../skills/good-first-issue-sweep`
- `skills/workflow-security-audit/workflow-security-audit` → `../../skills/workflow-security-audit`
- `skills/contributor-to-committer/magpie-contributor-to-committer` → `../../.agents/skills/magpie-contributor-to-committer`

## Context / how found

While bringing up **OpenCode** and **Kiro CLI** as runtime targets for the
vendor-neutrality runtime-adapter effort (#320), OpenCode's skill loader globs
`**/SKILL.md` *following symlinks* and walked these cycles into looped paths
(`skills/x/x/x/.../SKILL.md`). Discovery still succeeded (it dedupes by name and
the glob has a depth cap), but the cycle is fragile and unnecessary.

## Type of change

- [x] CI / dev loop — `setup` self-adoption wiring (no skill content changed)

## Test plan

- [x] `find skills -type l` → 0 (no symlinks remain in the source tree)
- [x] `opencode debug skill` resolves all three skills via clean, single-segment paths; total skill count unchanged
- [x] `skill-and-tool-validate` passes
- [x] No `SKILL.md` content altered — pure symlink removal

---
*Gen-AI disclosure: authored with Kiro CLI (Opus 4.8); commit carries a `Generated-by:` trailer per AGENTS.md.*

### DIFF
--- a/skills/contributor-to-committer/magpie-contributor-to-committer
+++ b/skills/contributor-to-committer/magpie-contributor-to-committer
@@ -1,1 +1,0 @@
-../../.agents/skills/magpie-contributor-to-committer
--- a/skills/good-first-issue-sweep/good-first-issue-sweep
+++ b/skills/good-first-issue-sweep/good-first-issue-sweep
@@ -1,1 +1,0 @@
-../../skills/good-first-issue-sweep
--- a/skills/workflow-security-audit/workflow-security-audit
+++ b/skills/workflow-security-audit/workflow-security-audit
@@ -1,1 +1,0 @@
-../../skills/workflow-security-audit


Alexhans:
Follow-up guard: #649 adds a `check-symlink-cycles` prek hook so this class can't re-land. It's stacked on this PR (includes these removals so its CI is green); merge this one first, then #649's diff reduces to just the hook.

justinmclean:
Running the mapie review skill show no major issues.

## Correctness

No findings against the change. I reproduced the cycle check on the base tree: all three deleted links resolve to their own parent directory and are genuine filesystem cycles.

Completeness (verified): a full repo scan for cyclic in-source symlinks returns exactly these three and no others, so the PR removes every current offender.

Safety (verified): removing these does not orphan any skill. Each affected skill keeps its canonical wiring (.agents/skills/magpie-<skill> -> ../../skills/<skill>) and a .claude/skills/ relay, and all three retain their SKILL.md. The deleted links were redundant in-source back-links, not load-bearing.

## Conventions

[advisory] Pre-existing, not introduced here: .claude/skills/magpie-workflow-security-audit links straight into skills/ rather than relaying through .agents/skills/ like the other two. It is consistent with the acyclic invariant so it is not a blocker, but it diverges from the one-directional relay pattern documented in #649's script header. Out of scope for #648; flagging so it does not get lost.

## Cross-PR (the important one)

[blocking on process, not code] PR #648 and PR #649 — https://github.com/apache/magpie/pull/649 — delete the identical three symlinks. They will conflict or render each other partly no-ops depending on merge order. Recommend consolidating rather than merging both: prefer #649 (it does the same deletions and adds the recurrence guard), and close #648; or merge #648 first, after which #649 reduces cleanly to just the hook plus script. Merging #648 alone leaves the cause unguarded.

Alexhans:
Ack.  I have not really used stacked PRs in GitHub in the past so I think I fudged the double deletion there. 

In other commits the approached seemed to be separate fix (deletions) and new approach (guard).  Is the guidance that we should make 1 PR and 1 commit instead of the 2 I posted?

potiuk (approved):

PR #649, opened 2026-06-30
feat(tools/symlink-lint): lint cyclic and misdirected skill symlinks

Alexhans:
## Summary

- Add `tools/symlink-lint`, a stdlib-Python `prek` hook guarding the framework's
  self-adoption skill symlinks against two failure modes: **cycles** (a symlink
  resolving to its own dir/ancestor — traps recursive `**/SKILL.md` scanners in
  looped paths) and **misdirected relays** (an agent-dir relay pointing straight
  at source instead of through the canonical `.agents/skills/` entry — acyclic,
  so a cycle check alone misses it).
- Guard companion to #648: #648 removed the three cyclic in-source symlinks;
  this keeps those — and misdirected relays — from returning.
- Dangling links are skipped: an adopter's canonical links legitimately dangle
  until the `.apache-magpie/` snapshot is installed (`setup verify`'s job).

## Context / how found

Rule 2 (relay-correctness) directly addresses the
`.claude/skills/magpie-workflow-security-audit` divergence flagged in #648
review — and surfaced it wasn't a one-off: **9** committed relays under
`.claude/` and `.github/` pointed at `../../skills/<skill>` instead of
`../../.agents/skills/magpie-<skill>`. They're repointed here because the new
hook would otherwise fail on `main` — the guard and the tree it guards must land
together. Each is a one-line target repoint; no source moves.

Supersedes this PR's original bash implementation, per review — rewritten to
stdlib Python to fix portability (`pathlib.resolve()`), filename-safety
(`os.walk`), and testability (a `pytest` suite) at the root.

## Type of change

## Test plan

- [x] `uv run pytest` / `ruff check` / `mypy` pass (13-case suite: both rules +
      dangling / pruned / symlink-to-file / whitespace edges + `main()` exit codes)
- [x] `meta-and-quality-tooling.md` spec updated to list the tool; no skill loads
      `symlink-lint`, so no skill-eval / `mode-economics` update applies

---
*Gen-AI disclosure: authored with Kiro CLI (Opus 4.8); commit carries a `Generated-by:` trailer per AGENTS.md.*

### DIFF
--- a/tools/symlink-lint/pyproject.toml
+++ b/tools/symlink-lint/pyproject.toml
@@ -1,0 +1,81 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+[build-system]
+requires = ["hatchling"]
+build-backend = "hatchling.build"
+
+[project]
+name = "symlink-lint"
+version = "0.1.0"
+description = "Detect self-referential / cyclic symlinks (a symlink resolving to its own directory or an ancestor), which break recursive SKILL.md scanners."
+readme = "README.md"
+requires-python = ">=3.11"
+license = { text = "Apache-2.0" }
+# stdlib-only — os.walk + pathlib resolve every path we need.
+dependencies = []
+
+[project.scripts]
+symlink-lint = "symlink_lint:main"
+
+[tool.hatch.build.targets.wheel]
+packages = ["src/symlink_lint"]
+
+[tool.ruff]
+line-length = 110
+target-version = "py311"
+src = ["src", "tests"]
+
+[tool.ruff.lint]
+select = [
+  "E",     # pycodestyle errors
+  "W",     # pycodestyle warnings
+  "F",     # pyflakes
+  "I",     # isort
+  "B",     # flake8-bugbear
+  "UP",    # pyupgrade
+  "SIM",   # flake8-simplify
+  "C4",    # flake8-comprehensions
+  "RUF",   # ruff-specific
+]
+ignore = [
+  "E501",  # line-too-long — the 110-char limit above is already generous
+]
+
+[tool.ruff.lint.per-file-ignores]
+"tests/**" = ["B", "SIM"]
+
+[tool.mypy]
+python_version = "3.11"
+files = ["src", "tests"]
+warn_unused_ignores = true
+warn_redundant_casts = true
+warn_unreachable = true
+check_untyped_defs = true
+no_implicit_optional = true
+disallow_untyped_defs = true
+disallow_incomplete_defs = true
+
+[[tool.mypy.overrides]]
+module = "tests.*"
+disallow_untyped_defs = false
+disallow_incomplete_defs = false
+
+[tool.pytest.ini_options]
+minversion = "8.0"
+addopts = "-ra -q"
+testpaths = ["tests"]
--- a/.claude/skills/magpie-issue-backlog-stats
+++ b/.claude/skills/magpie-issue-backlog-stats
@@ -1,1 +1,1 @@
-../../skills/issue-backlog-stats
+../../.agents/skills/magpie-issue-backlog-stats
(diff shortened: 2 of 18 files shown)


justinmclean:
Ran the Magpie review skill over these changes - no blockers, but you still might want to look into the macOS issue.

## Correctness

The core invariant is implemented correctly. A link is flagged when realpath(link) == dirname(link) or when the link's directory sits under the resolved target, and the trailing-slash prefix strip ("${dir#"$target"/}") correctly avoids the classic /a/foo vs /a/foobar boundary false positive. The three deletions are all genuine cycles: each resolves back to its own parent directory (including contributor-to-committer, which cycles through the canonical .agents/skills/ link). No blocking findings.

[advisory] tools/dev/check-symlink-cycles.sh:~95 : the guard depends on GNU readlink -f, and the || true converts its absence into a silent pass rather than an error.

target="$(readlink -f "$link" 2>/dev/null || true)"

Rule: on a platform without GNU coreutils (stock macOS readlink has no -f), every link resolves to empty and is skipped as "dangling", so the local hook passes vacuously and gives false assurance. CI on Linux still enforces. tools/dev/README.md declares a coreutils runtime so this is a documented assumption, but there is no greadlink shim; a one-line capability check or greadlink fallback would close the gap.

[advisory] tools/dev/check-symlink-cycles.sh:~120 : the find-to-while pipeline splits on newlines, so a symlink whose name contains a newline or leading/trailing whitespace would be mis-parsed.

find . -type d \( ... \) -prune -o -type l -print | sort feeding while IFS= read -r link

Rule: prefer -print0 with read -r -d '' for filename-safe iteration. Low likelihood, but it is the standard robustness fix.

[advisory] tools/dev/check-symlink-cycles.sh:~98 : dir="$(cd "$(dirname "$link")" && pwd)" uses the logical path while readlink -f returns the physical one, so a symlinked intermediate directory could make the two incomparable.

dir="$(cd "$(dirname "$link")" && pwd)"

Rule: use pwd -P to match readlink -f semantics. Low impact here since the targeted cycles live inside real skills/<skill>/ directories.

## Security

No findings. The script introduces no injection or credential surface, treats no untrusted external input, and sets set -euo pipefail. The diff content carries no embedded instructions to flag.

## Conventions

[advisory] tools/dev/check-symlink-cycles.sh : 133 lines of non-trivial detection logic land without a fixture test.

new file, no accompanying tests/ or .bats fixture

Rule: this matches the existing repo norm (sibling check-placeholders.sh, agent-pre-commit.sh, run-workspace-check.sh are all untested), so it is not blocking. Given the subtlety of the cycle invariant, a small fixture asserting cyclic link to exit 1, dangling link to skip, and the canonical-plus-relay pair to exit 0 would lock the behaviour.

Positives: the new script carries the full ASF license header, the hook mirrors the sibling check-placeholders.sh pattern (language: system, pass_filenames: false), and the executable bit is set. No SPDX or placeholder-token violations.

Alexhans:
Ack. Will take a look and update.  I had not run the review skill. 

I'm "dog fooding" kiro-cli I'm this case while finding while testing what it would take to implement kiro-cli and opencode (2 coding harnesses at once to avoid fixating on specific aspects of either in contrast with Claude).

Depending on the decision on 648's https://github.com/apache/magpie/pull/648#issuecomment-4840395932 I'll roll the changes into one PR/commit or put them here.

Alexhans:
@justinmclean thanks for the review — I've reworked this substantially in response, and force-pushed the branch:

- **Rewrote bash → a small stdlib-Python tool** (`tools/symlink-lint`), which dissolves all three of your advisories at the root rather than patching them:
  - portability — no `readlink`/`greadlink`; `pathlib.resolve()` is cross-platform;
  - filename-safety — no `find | while read`; `os.walk` handles spaces/newlines;
  - the bespoke test — replaced by a proper `pytest` suite (13 cases) run by the workspace harness + CI.
- **Broadened the remit** from cycles-only to "self-adoption symlink hygiene": it now also enforces relay/canonical target-correctness. That directly addresses the `.claude/…workflow-security-audit` divergence you flagged — and it turned up 8 more of the same class, all repointed in this PR.

Because the relay rule now guards the tree, the 9 repointed relays have to land with the hook (it'd fail on `main` otherwise) — happy to split them into a separate PR first if you'd prefer that ordering.

The title/type changed accordingly (`chore(pre-commit):` → `feat(tools/symlink-lint):`); PR body updated to match.

tools/symlink-lint/pyproject.toml
```diff
@@ -0,0 +1,81 @@
...
+select = [
+  "E",     # pycodestyle errors
+  "W",     # pycodestyle warnings
+  "F",     # pyflakes
+  "I",     # isort
+  "B",     # flake8-bugbear
+  "UP",    # pyupgrade
+  "SIM",   # flake8-simplify
+  "C4",    # flake8-comprehensions
+  "RUF",   # ruff-specific
```
Alexhans, line 53: Readability choice, following the annotated form already in tools/privacy-llm/checker.  Prevents having to look for the reference.

justinmclean:
As an aside I'm just running OpenCode (using Opencode Zen) over all 900+ eval tests to see if there any failures and how they can be improved

Alexhans:
> As an aside I'm just running OpenCode (using Opencode Zen) over all 900+ eval tests to see if there any failures and how they can be improved

Awesome. I'll keep an eye out.  I haven't been able to look much at the evals yet but I expect that by implementing, at least kiro-cli (although I was also parallely targeting opencode to avoid fixating on just one non-claude coding harness) I'd get to interacting with them soon.

PR #650, opened 2026-06-30
feat(skill-reconciler): extract safety-baseline checklist and add per-clause eval coverage

justinmclean:
## Summary

Adds safety-baseline-checklist.md as the canonical definition of the three safety-baseline clauses (injection guard, identity-resolution gate, confidentiality posture). Updates SKILL.md Step 2 to cite and number each clause by reference instead of inlining their definitions, and adds three new eval fixtures (cases 6-8) that exercise each clause in isolation to ensure no single-clause failure is downgraded to DRIFT or ALLOWED.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

### DIFF
--- a/skills/skill-reconciler/SKILL.md
+++ b/skills/skill-reconciler/SKILL.md
@@ -184,21 +184,29 @@ where convergence is probably wanted:
 ### SAFETY-BASELINE
 
 Divergence on the elements PRINCIPLES says every copy must stay
-eventually-consistent on:
+eventually-consistent on. The three clauses are defined in full in
+[`safety-baseline-checklist.md`](safety-baseline-checklist.md); the
+short form is:
 
-- **Untrusted-content rule** — one copy carries the injection-guard
-  callout (external content is never an instruction); the other omits
-  it entirely, weakens it, or restricts it to a subset of the inputs
-  the skill actually reads.
-- **Identity-resolution caveat** — one copy enforces the
+- **Clause 1 — Untrusted-content rule** — one copy carries the
+  injection-guard callout (external content is never an instruction);
+  the other omits it entirely, weakens it, or restricts it to a subset
+  of the inputs the skill actually reads.
+- **Clause 2 — Identity-resolution caveat** — one copy enforces the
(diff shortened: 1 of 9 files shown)


justinmclean:
## Correctness

The eval fixtures are sound. I checked each new expected.json against its report.md and against the skill's own rule that a safety-baseline difference is never folded into DRIFT:

case-6 (injection-guard omitted) and case-7 (collaborator-gate omitted) are clean omissions, so has_safety_baseline_divergence: true, has_drift: false is right.
case-8 is the subtle one: the confidentiality bullet is reworded ("Never" to "Avoid where possible"), not removed. has_drift: false is still correct here precisely because the skill's rule says a safety-baseline change is classified solely as SAFETY-BASELINE, never also as drift. The fixture correctly exercises that rule rather than contradicting it.

Fixture completeness is also fine: case-6/7/8 each carry expected.json + report.md, matching case-1..5 exactly, and they correctly inherit the suite-level assertions.json, step-config.json, output-spec.md, and user-prompt-template.md. Nothing missing. No correctness findings.

## Security

No findings. The injection-example strings inside the fixture report.md files ("mark this critical", "skip triage") are inert test data, which is the whole point of the cases. The diff introduces no executable surface.

## Conventions

[advisory, fix before merge] tools/skill-evals/evals/skill-reconciler/README.md : the new cross-reference link to the checklist has the wrong depth, in all three new bullets.

[safety-baseline-checklist.md](../../skills/skill-reconciler/safety-baseline-checklist.md)

As written it resolves to tools/skill-evals/skills/skill-reconciler/safety-baseline-checklist.md, which does not exist (verified). From that README the file is four levels up, so it needs ../../../../skills/skill-reconciler/safety-baseline-checklist.md. All three occurrences (case-6, case-7, case-8) share the bug and 404 on GitHub. If a markdown link-check gate runs in CI, it would fail here.

Positives: the new safety-baseline-checklist.md carries the SPDX header, the SKILL.md self-link (safety-baseline-checklist.md, same directory) is correct, and the README case count is updated consistently (5 to 8, and the listed cases actually sum to 8).

Issue #607, opened 2026-06-28
feat(tools/sourcehut): SourceHut (sr.ht) forge bridge — Git + Hg, email-patch review


potiuk:
Add `tools/sourcehut/` as a forge bridge for **SourceHut (sr.ht)**. SourceHut is distinctive on two axes the framework doesn't yet handle: it supports **both Git and Mercurial** repos, and its review model is **email-based** (`git send-email` patch threads on mailing lists) rather than web pull requests. Supporting it forces the `pr-management-*` skills to treat "a patch series on a list" as a first-class review unit — valuable well beyond sr.ht.

**Capabilities to provide:**

- Repo + ref read across both git.sr.ht and hg.sr.ht (reuse the Mercurial shim for the Hg case)
- todo.sr.ht tracker: ticket read/write via the GraphQL API
- lists.sr.ht: read patch-series threads; map a patchset to the PR/MR review abstraction
- builds.sr.ht: CI status read
- GraphQL API across the sr.ht services

**Why:** SourceHut hosts a committed minimalist-tooling community — **postmarketOS**, parts of the **Alpine**/**Sway**/**wlroots** orbit, and many Unix-philosophy projects — and is the reference implementation of the email-patch workflow that the broader **kernel-style** development model uses. Teaching the skills to handle list-based patch review (here and, later, for LKML-style projects) is the real payoff.

**Reference:**

- SourceHut: https://sourcehut.org/
- GraphQL APIs: https://man.sr.ht/graphql.md
- Depends on the Mercurial VCS shim for hg.sr.ht; relates to the mail-source bridges (#303/#304/#306) for the list side

onlyarnav:
This feature have been implemented, @potiuk could you please confirm and mark this as completed. Thanks! 😃

justinmclean:
completed

Issue #601, opened 2026-06-28
feat(tools/mercurial): Mercurial (Hg) version-control support for the local dev-loop


potiuk:
Add `tools/mercurial/` so the local dev-loop skills can operate on **Mercurial (Hg)** repositories, not just Git. Today the framework assumes Git everywhere — `git checkout`/`commit`/`diff`/`rebase` are baked directly into `issue-fix-workflow`, `pr-management-code-review`, and the `setup-steward` worktree machinery. This is the first "not only Git" version-control bridge and should establish the abstraction the others (SVN, Jujutsu, Fossil, Perforce) reuse.

**Capabilities to provide** (mirror the implicit Git contract the skills rely on):

- Working-copy ops: branch/bookmark create + switch, commit, `hg diff`, `hg status`
- History read: log, annotate/blame, file-at-revision (for the reproducer + reassess skills)
- Patch/queue export for the fix-workflow hand-off (`hg export`, optionally MQ/evolve)
- A thin VCS-capability shim so skills call `vcs.diff` / `vcs.commit` rather than `git ...` directly

**Why:** Mercurial still backs a meaningful slice of OSS — the **Mercurial** project itself, large historical **Mozilla** trees, **Octobus/Heptapod** (GitLab-on-Hg) hosting, **Meta's** internal Sapling lineage, and many academic / scientific repos. A project-agnostic stewardship framework that hard-codes Git can't be adopted by an Hg shop at all.

**Reference:**

- Mercurial CLI: https://www.mercurial-scm.org/doc/hg.1.html
- Heptapod (GitLab + Hg): https://heptapod.net/
- Git touchpoints to abstract: search the skills for `git checkout` / `git commit` / `git diff` / `<default-branch>`

onlyarnav:
Hi sir,

I have implemented a complete, possible approach for both Mercurial (`hg`) VCS backend support (#601) and the SourceHut (`sr.ht`) forge bridge (#607) in a fork and opened a pull request. 

Here is a summary of the changes implemented:
- **Mercurial VCS Backend (#601):** Implemented the full command-line wrapper in `tools/vcs` supporting status, branch bookmarks, staging, commits, diffs, and worktree purges along with custom unit tests.
[...]

PR #670, opened 2026-07-01
feat(tools/sourcehut, tools/vcs): add SourceHut forge bridge and implement Mercurial backend

onlyarnav:
## Summary

- **Adds `tools/sourcehut` as a forge bridge:** Implements ticket tracking (`todo.sr.ht`), mailing list patchset review (`lists.sr.ht`), CI build status (`builds.sr.ht`), and repository reads (`git.sr.ht` / `hg.sr.ht`) using SourceHut's GraphQL APIs.
- **Implements `MercurialBackend` in `tools/vcs`:** Replaces the Mercurial (`hg`) stub with a complete, fully functional command-line wrapper backend supporting bookmarks, staging, commits, diffs, logs, and worktree purging.
- **Updates registries and docs:** Registers `tools/sourcehut` capabilities (`contract:tracker + contract:source-control + contract:mail-archive`) and marks Mercurial support as complete.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `prek run --all-files` passes (verified with local pre-flight checks, skipping link checker hook `lychee` which requires MSVC toolchain to build from source on Windows).
  - Ran `pytest` on `tools/vcs` (19 passed, 3 skipped because `hg` is not installed on the local system).
  - Ran `pytest` on `tools/sourcehut` (13 passed offline using mocked HTTP calls).
  - Ran `skill-and-tool-validate` directly; all validations passed cleanly with 0 hard schema errors.

## RFC-AI-0004 compliance

- [x] **HITL** — any new mutation is gated on explicit user confirmation (e.g. CLI operations for creating tickets/comments).
- [x] **Sandbox** — no new unrestricted host access; network reach declared in the adapter.
- [x] **Vendor neutrality** — placeholders (`<PROJECT>`, `<tracker>`, `<upstream>`, `<security-list>`) used in all skill / tool prose (verified via validator).

## Linked issues

Refs #601 (Mercurial support), Refs #607 (SourceHut integration).

- **This project was built pair-programming with the user using the Antigravity Gemini 3.5 Flash AI coding assistant.**
- Mercurial support uses the standard `hg` CLI. The tests gracefully skip if `hg` is not installed locally.
- SourceHut integration uses standard library `urllib` to keep the CLI wrapper dependency-free.
- Patchsets on `lists.sr.ht` are mapped into standard pull requests by treating the cover letter email as the description, individual patches as commits, and mailing list replies as review comments.


onlyarnav (commented):
All the changes suggested by GitHub Actions bot have been implemented, now working on CI pipeline failing

potiuk (approved):

justinmclean:
Thanks for the contribution. I ran the PR review skill over it, and there are a couple of minor things you might want to fix:

A scattering of minor/nit observations (one markdown-table break in the docs, two hg-backend contract-parity concerns, two sourcehut nits), none of which gate the merge.

Approach looks solid on both fronts — the SourceHut bridge is clean stdlib-only code with real tests, and the Mercurial backend fills in a genuine extension point with matching @hg_required tests. A few observations below, none blocking; the one I'd most like a second look at is a markdown table that the docs hunk appears to break.
docs/vendor-neutrality.md — Status at a glance table split by an inserted blank line (docs/vendor-neutrality.md:84)

A blank line terminates a GFM table, so the "Project governance" row renders as loose text below a truncated table. The lychee and doctoc hooks won't flag this (not a link or TOC entry), so it needs a manual fix — remove the inserted blank line. While you're in this file, the earlier hunk rewriting the "non-Git/non-Hg systems" list also leaves a stray blank line before [Jujutsu] that splits the sentence; worth tidying in the same pass.
Smaller observations

tools/vcs/src/magpie_vcs/__init__.py:1372 — create_branch maps to hg bookmark <name>, but the abstract contract is "Create and switch to a new line of work" and the Git binding switches (git checkout -b). hg bookmark creates/activates a bookmark without moving the working directory the way a branch checkout does, and the new test commits on the same parent so it doesn't exercise the "switched to a different line of work" semantics. Not a definite bug — flagging for a maintainer with hg expertise.

tools/vcs/src/magpie_vcs/__init__.py:1396 — reset_worktree uses hg purge --all, which removes .hgignore'd files too, whereas the Git binding's git clean -fd deletes only untracked files. If the per-run reset is meant to preserve ignored build/cache artefacts, drop --all. There's also no @hg_required test for reset_worktree (the git backend has one), so this difference is unverified.

tools/sourcehut/src/magpie_sourcehut/cli.py:496 — ticket label with neither --add nor --remove is a silent no-op that still prints the ticket and exits 0. Consider requiring at least one (argparse required-mutually-exclusive group), or a "nothing to do" note.

tools/sourcehut/README.md:141 — Features says the VCS integration "Reads repo metadata and refs", but repo.py's query pulls only id/name/description/visibility/updated — no refs. Tighten the wording or add refs to the query.


This review was drafted by an AI-assisted tool. The findings below are observations, not blockers; an Apache Magpie maintainer — a real person — will take the next look at the PR. If you think a finding is mis-applied, please reply on the PR and a maintainer will weigh in.

tools/sourcehut/README.md
```diff
@@ -0,0 +1,53 @@
...
+  - [Prerequisites](#prerequisites)
+  - [Features](#features)
+  - [Invocation](#invocation)
+  - [Configuration](#configuration)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+# SourceHut (sr.ht) Forge Bridge
+
+**Capability:** contract:tracker + contract:source-control + contract:mail-archive
```
potiuk, line 15: Heads-up: after the merge with `main` this now breaks CI. `tools/sourcehut` declares `contract:tracker + contract:source-control + contract:mail-archive`, but `main` recently gained `tools/vendor-neutrality-score` (#672), which requires every contract tool's README to declare `**Kind:**` and `**Vendor:**`. Without them its loader raises `ValueError` and the `vendor-neutrality-score` pytest job fails. Please add, e.g.:

```
**Kind:** implementation

**Vendor:** SourceHut
```

(Bonus: that also makes SourceHut count as an additional backend for tracker / source-control / mail-archive in the neutrality score.)

tools/sourcehut/src/magpie_sourcehut/client.py
```diff
@@ -0,0 +1,89 @@
...
+            return res_json.get("data", {})
+    except urllib.error.HTTPError as exc:
+        try:
+            err_body = exc.read().decode("utf-8")
+            err_json = json.loads(err_body)
+            err_errors = err_json.get("errors")
+            if err_errors:
+                err_msgs = [e.get("message", "Unknown error") for e in err_errors]
+                raise SourceHutError(f"HTTP {exc.code}: {'; '.join(err_msgs)}")
+        except Exception:
```
potiuk, line 82: The inner `except Exception: pass` catches the `raise SourceHutError(...)` two lines above, so the parsed GraphQL error messages are always discarded and callers fall through to the generic `HTTP request … failed with status N` (L85). A 400 with `{"errors":[{"message":"invalid tracker"}]}` surfaces only the status, never the message. Suggest building the message string inside the `try` and raising after it, or narrowing the inner `except` to `(json.JSONDecodeError, KeyError, TypeError)`.

potiuk (commented):
Thanks for this — a well-structured, dependency-free bridge and a clean Mercurial backend, and nice quick turnaround on the earlier fixes. The top item below is CI-blocking after your merge with `main`; the rest are mostly medium/low. Two doc items land on lines this PR didn't change, so I couldn't anchor them inline:

- `tools/vcs/README.md:92` still lists `hg` as “extension point | operations raise a clear error → #601”, contradicting the working backend this PR adds (the module docstring at `tools/vcs/src/magpie_vcs/__init__.py` also still calls Mercurial an extension point).
- `docs/vendor-neutrality.md:208` (Tool adapters table) still lists Mercurial only as an extension point, while L163/L396/L488 in the same file now say “Mercurial (complete)”.

tools/sourcehut/src/magpie_sourcehut/cli.py
```diff
@@ -0,0 +1,188 @@
...
+
+
+def main(argv: Sequence[str] | None = None) -> int:
+    parser = _build_parser()
+    ns = parser.parse_args(argv)
+
+    if ns.subcommand == "graphql":
+        vars_dict = None
+        if ns.variables:
+            vars_dict = json.loads(ns.variables)
```
potiuk: `main()` has no error handling: `json.loads(ns.variables)` raises `JSONDecodeError` on malformed `--variables`, and any `SourceHutError` from the calls below escapes as a raw traceback rather than a clean message. Consider wrapping the dispatch in `try/except (SourceHutError, json.JSONDecodeError) as e:` → `print(f"magpie-sourcehut: {e}", file=sys.stderr); return 2`.

tools/vcs/src/magpie_vcs/__init__.py
```diff
@@ -356,13 +356,94 @@ def reset_worktree(self) -> None:
...
+        return _run(["hg", "branch"], self.root).strip()
+
+    def diff(self, base: str | None = None, cached: bool = False, paths: Sequence[str] = ()) -> str:
+        if cached:
+            raise VCSError("hg does not support staging area/cached diff")
+        args = ["hg", "diff"]
+        if base:
+            args.extend(["-r", base])
+        if paths:
+            args.extend(paths)
```
potiuk: `diff` (here) and `log` (L414) append `paths` without a `--` terminator, unlike `GitBackend` (L232/L253) and this class's own `stage` (L426). A path beginning with `-`, or named like an hg flag, is parsed as an option → wrong output or an hg error. Suggest `args.extend(["--", *paths])` in both.

tools/sourcehut/src/magpie_sourcehut/lists.py
```diff
@@ -0,0 +1,179 @@
...
+        return {}
+
+    # Extract thread emails for description & comments
+    thread = patchset.get("thread") or {}
+    emails_conn = thread.get("emails") or {}
+    edges = emails_conn.get("edges") or []
+    emails = [edge.get("node") for edge in edges if edge.get("node")]
+
+    # Sort emails by date if possible
+    with contextlib.suppress(Exception):
```
potiuk, line 125: `contextlib.suppress(Exception)` swallows any sort failure, and then `emails[0]` is used as the cover letter — a malformed node would silently yield the wrong description/author. Narrow to `(TypeError, AttributeError)`, or validate the nodes before sorting.

tools/sourcehut/src/magpie_sourcehut/lists.py
```diff
@@ -0,0 +1,179 @@
...
+          }
+        }
+      }
+    }
+    """
+    res = query_graphql("lists", q, {"owner": owner, "name": list_name})
+    mlist = res.get("list") or {}
+    patches_conn = mlist.get("patches") or {}
+    edges = patches_conn.get("edges") or []
+    return [edge.get("node") for edge in edges if edge.get("node")]
```
potiuk: `edge.get("node")` assumes every edge is a dict; a null edge in the connection raises `AttributeError`. Same pattern at L122 and the `patches` loop at L147. Cheap to guard (`if edge` / `if patch is not None`).

tools/sourcehut/README.md
onlyarnav: Added the Kind and Vendor metadata fields to tools/sourcehut/README.md. I also regenerated the vendor neutrality score markdown block in docs/vendor-neutrality.md to include SourceHut and reflect the new calculated score.

tools/sourcehut/src/magpie_sourcehut/client.py
onlyarnav: I refactored the HTTPError handler to build the error message string inside the try block and then raise SourceHutError outside of it (using from exc for clean exception chaining to comply with B904). This successfully prevents the custom error from being swallowed.

tools/sourcehut/src/magpie_sourcehut/cli.py
onlyarnav: Wrapped the main CLI dispatch block in cli.py with a try/except (SourceHutError, json.JSONDecodeError) block to output clear error messages to sys.stderr and exit with status code 2 rather than leaking raw python tracebacks.

potiuk:
Rebased onto latest `main` — the two conflicts (`uv.lock` and the generated vendor-neutrality table) were regenerated rather than hand-merged, so nothing was lost and the score stays 10/10. Nice work overall; the SourceHut bridge is cleanly structured and the Mercurial backend is solid. Just a few small things left before merge, noted inline. 🙂

tools/vcs/src/magpie_vcs/__init__.py
```diff
@@ -356,13 +356,93 @@ def reset_worktree(self) -> None:
...
+        try:
+            subprocess.run(["hg", "--version"], capture_output=True, check=True)
+        except (OSError, subprocess.CalledProcessError):
+            return False
+        return True
+
+    def status(self) -> str:
+        return _run(["hg", "status"], self.root)
+
+    def current_branch(self) -> str:
```
potiuk, line 383: `current_branch()` reads the Hg *named branch* (`hg branch`, effectively always `default`), but `create_branch()` creates/activates a *bookmark*. So after `create_branch("feature")`, `current_branch()` still returns `"default"` rather than `"feature"` — the two operate on different Mercurial concepts. Since bookmarks are the git-branch analog here, consider having `current_branch()` prefer the active bookmark and fall back to the named branch, e.g. `hg log -r . -T '{activebookmark}'` (falling back to `hg branch` when empty). As-is, a skill that creates a line of work and then checks `current_branch()` to confirm would get the wrong answer.

potiuk:
In `tools/vcs/tests/test_vcs.py`, the `git_repo` fixture (around line 61) sets `user.email`/`user.name` but not signing, so on a machine with global `commit.gpgsign = true` every git test errors on `git commit` (exit 128). The module already scrubs `GIT_DIR` etc. for hermeticity — worth extending that here with `_git(repo, "config", "commit.gpgsign", "false")` (or setting `GIT_CONFIG_GLOBAL=/dev/null`) so the suite is reproducible regardless of the developer's global git config.

(Posting as a conversation comment rather than inline since the fixture lines aren't part of this PR's diff.)

onlyarnav:
Thanks for the kind words and for the help rebasing the branch, learning from the best! 😊

PR #724, opened 2026-07-04
fix(cve-tool-vulnogram): read/write CNA_private under the read API's body envelope

shahar1:
## Summary

- `vulnogram-api-record-publish` and `record-fetch --state-only` read `CNA_private` at the **top level** of the fetched record, but Vulnogram's read API nests the CVE document under a `body` envelope (`{comments, files, body:{cveMetadata, CNA_private, containers}, author, __v}`) — so both fail on **every real record** with `✗ … CNA_private field is not an object: NoneType` / `is in state None`. `record-update` was unaffected because `merge_mode` already reads `body.CNA_private`.
- Fix: add `client.record_body()` (returns the doc under `body`, else the doc itself) and route both commands' reads/writes through it; publish now POSTs the **unwrapped** document — the shape `record-update` already POSTs.
- The tests mocked the unwrapped top-level shape, which is exactly why this shipped green.

## Type of change

## Test plan

- [x] `prek run --files <the 5 changed files>` — all hooks pass (ruff check, ruff format, mypy, pytest, typos, check-placeholders)
- [x] For Python packages touched: `uv run --project tools/cve-tool-vulnogram/oauth-api pytest` — **96 pass**; `ruff` / `mypy` clean
- Regression coverage: `test_record_publish` mocks now use the real `body`-wrapped read shape and assert the POST carries the unwrapped document (`posted_body_envelope is False`); without the source fix, publish returns `3` (state unreadable). New `test_record_fetch` covers the previously-**untested** `--state-only` path, plus a `record_body` unit test for both shapes.

## RFC-AI-0004 compliance

- No behavioural or mutation-policy change — a read/write-location bug fix inside the Vulnogram CVE-tool adapter. No new host access, no new outbound messages, HITL unchanged.

## Notes for reviewers

Found while publishing a real CVE end-to-end: `record-update` set the record to `PUBLIC` correctly, but `record-publish --dry-run` and `record-fetch --state-only` both reported `state None` on the very same record — the two commands that read `CNA_private` at the top level instead of under `body`. The write path also mattered: `record-publish` mutated a bogus top-level `CNA_private` and POSTed the read envelope; it now mutates `body.CNA_private` and POSTs the unwrapped doc.

Generated-by: Claude Code (Opus 4.8)

### DIFF
--- a/tools/cve-tool-vulnogram/oauth-api/src/vulnogram_api/client.py
+++ b/tools/cve-tool-vulnogram/oauth-api/src/vulnogram_api/client.py
@@ -170,6 +170,27 @@ def get_record(
     return first
 
 
+def record_body(document: dict[str, Any]) -> dict[str, Any]:
+    """Return the CVE 5.x document inside a record fetched via :func:`get_record`.
+
+    Vulnogram's read endpoint wraps the stored record in an envelope::
+
+        {"comments": [...], "files": [...],
+         "body": {"cveMetadata": ..., "CNA_private": ..., "containers": ...},
+         "author": ..., "__v": ...}
+
+    The CVE document — and its ``CNA_private.state`` — lives under ``body``. A
+    locally-generated document (e.g. from ``generate-cve-json``) has those keys
+    at the top level and carries no ``body`` envelope, so fall back to the
+    document itself. This keeps callers agnostic to which shape they hold: read
+    or mutate ``CNA_private`` on the returned mapping, and POST that same
+    mapping back via :func:`update_record` (which expects the unwrapped CVE
+    document, not the read envelope).
+    """
+    body = document.get("body")
+    return body if isinstance(body, dict) else document
+
+
 def fetch_csrf_token(
     session: Session,
     cve_id: str,
--- a/tools/cve-tool-vulnogram/oauth-api/src/vulnogram_api/record_fetch.py
+++ b/tools/cve-tool-vulnogram/oauth-api/src/vulnogram_api/record_fetch.py
@@ -48,6 +48,7 @@ from vulnogram_api.client import (
     SessionExpired,
     VulnogramAPIError,
     get_record,
+    record_body,
 )
 from vulnogram_api.credentials import Session, locate_session
 
@@ -107,7 +108,7 @@ def main(argv: list[str] | None = None) -> int:
         return 6
 
     if args.state_only:
-        cna = document.get("CNA_private")
+        cna = record_body(document).get("CNA_private")
         if not isinstance(cna, dict):
             print(
                 f"✗ {args.cve_id} document's CNA_private field is not an object: {type(cna).__name__}.",
--- a/tools/cve-tool-vulnogram/oauth-api/src/vulnogram_api/record_publish.py
+++ b/tools/cve-tool-vulnogram/oauth-api/src/vulnogram_api/record_publish.py
@@ -56,6 +56,7 @@ from vulnogram_api.client import (
     SessionExpired,
     VulnogramAPIError,
     get_record,
+    record_body,
     update_record,
 )
 from vulnogram_api.credentials import Session, locate_session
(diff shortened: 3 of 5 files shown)


justinmclean:
Looks good to me:

Correctness verified — both os and sys are already imported (cve_json.py:102,105), so the new os.environ.get / sys.stderr uses are safe. The fallback guard mirrors the precedence in _resolve_config_path, so --config and $CVE_JSON_CONFIG never silently resolve to the legacy file, default correctly wins over legacy, and empty-string env is handled consistently by both functions. Tests exercise each branch.

License headers — the new test file carries the full ASF header; no header-tool finding applies (the pre-commit config has no insert-license/RAT hook, and nothing here masks one).

No blocking or major findings. One nit worth your eye (draftable as a single inline comment, drop it if you disagree):
cve_json.py:171 — nit: the guard config_path is None and not os.environ.get(_CONFIG_PATH_ENV) re-derives "are we on the default path?", which _resolve_config_path already decided. If that precedence ever changes, the two can drift. Optional: have _resolve_config_path signal whether it used the default so the fallback keys off one source of truth. Not a blocker; the current tests would catch a straightforward drift.

shahar1:
Thanks for the review!

Quick note on scope: the nit here (`cve_json.py:171`, `_resolve_config_path`, `$CVE_JSON_CONFIG`, the legacy-config fallback) is about the **config-path fallback**, which is a different PR — #725. This PR (#724) changes `record_publish.py` / `record_fetch.py` / `client.py` in `tools/cve-tool-vulnogram/oauth-api` (the `CNA_private` body-envelope fix) and doesn't touch config resolution — so it looks like this review was meant for #725.

I've applied that nit over on #725: `_resolve_config_path` now returns `(path, is_default)`, so the fallback keys off that one signal instead of re-deriving `config_path is None and not os.environ.get(...)` — one source of truth for the precedence decision (apache/magpie#725, commit `00fa3b4`, and my reply there).

Happy to act on anything specific to #724's diff if you spot something in the `CNA_private`/`body`-envelope change.

justinmclean:
Looks like I got your two PR's confused - here the right review for this one.

I verified the fix against the pre-PR code and it's correct on every count:

The bug is real and the fix resolves it. On the old code, _state(document) reads top-level CNA_private, which is absent on a wrapped read response, so current_state is None, falls outside accepted_states, and record_publish returns 3 ("Refusing the publish"), exactly the "publishing a real CVE failed" symptom from #726.

The fix is consistent with existing infrastructure. merge_mode.py already encodes the same asymmetry (_current_state reads current_doc.body.CNA_private.state; _new_state reads new_doc.CNA_private.state), so record_body's model, read is wrapped, write is unwrapped, matches what the codebase already assumed. The body-key heuristic is safe because CVE 5.x has no legitimate top-level body key.

No aliasing bug. record_body returns document["body"] (a live reference) for wrapped docs and document itself when unwrapped; main mutates that reference and POSTs it, both paths correct.
Tests are thorough. Unwrap/passthrough, state-only-under-body, missing-CNA_private (rc 7), full-fetch-dumps-envelope, and a publish regression asserting update_record receives the unwrapped shape (posted_body_envelope: False) that would catch reintroduction. New test file carries the full ASF license header.

No blocking, major, or minor findings. One nit for your eye only:

client.py / merge_mode.py — the "read wraps under body, write is unwrapped" knowledge is now expressed in two places (the new record_body helper and merge_mode's _path(current_doc, "body", …)). Not in scope for this fix, but a future cleanup could route merge_mode through the same helper so the envelope contract lives in one spot.

LGTM — correct root-cause fix for the CNA_private body-envelope bug, with a regression test that pins the POST shape so it can't silently come back. The record_body unwrap-or-passthrough helper matches the read/write asymmetry merge_mode.py already encodes, and the new test file carries the ASF header. Optional future cleanup: let merge_mode share the same helper so the envelope contract lives in one place. Nothing blocking.

This review was drafted by an AI-assisted tool and confirmed by a Magpie maintainer. The maintainer approving this PR has read the findings and signed off. If something feels off, please reply on the PR and a maintainer will follow up.

PR #725, opened 2026-07-04
fix(cve-tool-vulnogram): fall back to the legacy config path after the tool rename

shahar1:
## Summary

- `generate-cve-json` resolves its config only at `.apache-magpie-overrides/tools/cve-tool-vulnogram/cve-json-config.toml`. The tool dir was renamed `tools/vulnogram` -> `tools/cve-tool-vulnogram`, so an adopter who upgraded across the rename keeps their config at the **old** path and every run hard-fails with `config file not found` until they pass `--config` or move the file by hand.
- Fix: when resolving from the default location (no `--config`, no `$CVE_JSON_CONFIG`) and only the pre-rename copy exists, use it with a one-line deprecation warning pointing at the new path. An explicit `--config` / env path that is missing still errors — no silent fallback.

## Type of change

## Test plan

- [x] `uv run pytest tools/cve-tool-vulnogram/generate-cve-json/tests/test_config_resolution.py` — 5 cases: default used when present; legacy fallback + warning when default missing; default wins over legacy; both-missing raises pointing at the new path; explicit missing `--config` does **not** fall back. Existing suite unaffected.
- [x] `prek run --files <changed>` — ruff check / ruff format / mypy / pytest / typos / check-placeholders all pass.

## RFC-AI-0004 compliance

- No behavioural or mutation change — a back-compat path-resolution fix. No new host access, no outbound messages, HITL unchanged.

## Notes for reviewers

Hit while running `generate-cve-json` in an adopter repo that adopted **before** the `vulnogram` -> `cve-tool-vulnogram` rename. A more systemic alternative would be to migrate `.apache-magpie-overrides/tools/<renamed>/` directories during `/magpie-setup upgrade` (the framework already migrates other renamed artefacts on upgrade); this PR is the minimal, self-contained fix that unblocks affected adopters immediately and nudges them to migrate.

Generated-by: Claude Code (Opus 4.8)

### DIFF
--- a/tools/cve-tool-vulnogram/generate-cve-json/src/generate_cve_json/cve_json.py
+++ b/tools/cve-tool-vulnogram/generate-cve-json/src/generate_cve_json/cve_json.py
@@ -121,6 +121,8 @@ import tempfile
 #      `_load_config()`).
 #   2. The `CVE_JSON_CONFIG` environment variable.
 #   3. `<cwd>/.apache-magpie-overrides/tools/cve-tool-vulnogram/cve-json-config.toml`.
+#      Back-compat: if only the pre-rename `tools/vulnogram/` copy exists at
+#      this default location, it is used with a deprecation warning.
 #
 # Schema: see the README in this package.
 import tomllib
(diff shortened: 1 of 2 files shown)


shahar1:
Thanks for the review! Addressed the nit: `_resolve_config_path` now returns `(path, is_default)`, so the legacy-path fallback in `_load_config` keys off that single signal instead of re-deriving `config_path is None and not os.environ.get(...)`. The precedence decision now lives in one place, so the two can't drift. Tests unchanged and still green (ruff/format/mypy/pytest via `prek`).

justinmclean:
Thanks for the contribution

PR #732, opened 2026-07-04
feat(skill-reconciler): add --discover mode for capability-tag auto-pairing

justinmclean:
## Summary

Extends `skill-reconciler` with an opt-in `--discover <skills-dir>` flag that scans a skills directory, groups SKILL.md files by `capability:` frontmatter, and presents a bounded candidate-pair list (≤ 20 pairs) for the maintainer to confirm before any comparison begins. Explicit-path mode remains the default; `confirmation_required` is always `true`.

Ships `step-0-discover-pairs` eval suite (4 cases):
- case-1-shared-capability: two skills sharing capability → one pair
- case-2-no-shared-capability: unique capabilities → empty list
- case-3-multiple-pairs: two capability groups → two pairs
- case-4-bounded-output: 21 pairs capped at 20, total_pairs_found: 21

Also includes:
- `tools/skill-reconciler-diff/README.md` (forwarded from the `skill-reconciler-structural-diff` branch to unblock the validator)
- `docs/labels-and-capabilities.md` row for `skill-reconciler-diff` (likewise from that branch; both are no-ops when that PR lands)
- `specs/skill-reconciler.md` Known gaps updated to reflect that capability-tag pairing is shipped; explicit source: tag registry deferred.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

### DIFF
--- a/skills/skill-reconciler/SKILL.md
+++ b/skills/skill-reconciler/SKILL.md
@@ -80,17 +80,30 @@ non-blocking — the user may defer.
 
 ## Inputs
 
-- **Skill A path** — path to the first `SKILL.md` (or its parent
-  directory). Required.
-- **Skill B path** — path to the second `SKILL.md` (or its parent
-  directory). Required.
-- **`--safety-only`** (optional) — restrict the report to
-  `SAFETY-BASELINE` findings only; omit `ALLOWED` and `DRIFT` rows.
-  Useful for a quick safety audit across many pairs.
+Two modes are supported:
 
-When the user provides directory paths, resolve to the `SKILL.md` inside.
-When a path resolves to nothing, stop at Step 0 and report the missing
-file.
+**Explicit-path mode (default):**
+- **Skill A path** — path to the first `SKILL.md` (or its parent
+  directory). Required in explicit-path mode.
+- **Skill B path** — path to the second `SKILL.md` (or its parent
+  directory). Required in explicit-path mode.
+
+**Discovery mode:**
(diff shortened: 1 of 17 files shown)


potiuk:
I've rebased this onto current `main` (now that #733 landed) and pushed fixes for the review points directly to your branch — all gates green (CI 36/36; ruff / mypy / pytest / skill-and-tool-validate / spec-validate / lychee).

**What I changed on your branch:**

- **Bounded-output eval now tests the cap.** Added a deterministic `max_length` predicate to the skill-evals harness and a `has_bounded_candidate_pairs` assertion (max 20) on `case-4-bounded-output`, so the 21-pair case now fails if `candidate_pairs` exceeds 20 — previously it only checked `total_pairs_found`. Added a harness unit test for the new predicate.
- **Dropped the dead `confirmation_required` assertion** from `assertions.json`: it doesn't start with `has_`/`mention_`, so the runner routed it to the exact comparator and the named predicate never fired. The field is still validated by exact match, so no coverage is lost.
- **Rebase reconciliation:** the `skill-reconciler-diff` README / labels row / `pyproject` / spec you forwarded to unblock the validator are now on `main` via #733, so I dropped the forwarded copies and merged the spec Known-gaps section — kept your "capability-tag auto-pairing shipped" note alongside #733's "structural-diff helper shipped".

Applied the `capability:reconciliation` + `family:tools` labels. Merging.

PR #10, opened 2026-04-29
feat(release-manager-handoff): explicit hand-off + publication-ready comments

potiuk:
## Summary

Adds two new first-class comments the `sync-security-issue` skill posts
on the tracker as the issue moves through the release-manager half of
the lifecycle (Steps 12-15), instead of folding the call-to-action into
the catch-all status rollup.

The rollup is the security team's audit trail and accumulates many
small entries collapsed into `<details>` blocks; the RM-facing
comments are one-shot **orientation surfaces** with their own dated
context and working `@`-mention notifications. A drive-by RM should be
able to drive Steps 13-15 from two comments on the issue page without
consulting the rollup or external docs.

### Comment 1 — Release-manager hand-off (Step 12 trigger)

Fires *exactly once* at the `pr merged` → `fix released` swap. Walks
the RM through Steps 13-15 end-to-end with a numbered checklist:
first paste → REVIEW → (rare conditional re-paste after CNA reviewer
comments) → READY → **email preview on the `#email` tab** (new
load-bearing checkpoint, see below) → send advisory → wait for
publication-ready notification → second paste → PUBLIC → close.

Template body: [`tools/vulnogram/release-manager-handoff-comment.md`](https://github.com/apache/airflow-steward/blob/feat/release-manager-handoff-comments/tools/vulnogram/release-manager-handoff-comment.md).
Idempotent via the marker `<!-- apache-steward: release-manager-handoff v1 -->`
on line 1.

### Comment 2 — Publication-ready notification (Step 14 trigger)

Fires *exactly once* when the *Public advisory URL* body field is
populated by sync (the existing Step 14 transition). Tells the RM
the URL has been captured, the JSON has been regenerated to include
it as a `vendor-advisory` reference, and the final paste + `READY`
→ `PUBLIC` + close is now unblocked.

Template body: [`tools/vulnogram/release-manager-publication-comment.md`](https://github.com/apache/airflow-steward/blob/feat/release-manager-handoff-comments/tools/vulnogram/release-manager-publication-comment.md).
Idempotent via `<!-- apache-steward: release-manager-publication-ready v1 -->`.

## Architectural split (tool-specific vs generic)

The user's request was *"do it in the way that 'tool' specific part is
done in the tool part in apache-steward submodule and generic workflow
is explained in generic part of it"*. Concretely:

| Concern | Lives in |
|---|---|
[...]

PR #735, opened 2026-07-04
test(spec-loop): add runner fixture coverage

justinmclean:
## Summary

Add deterministic fixture coverage for the spec-loop runner

Extracts the spec-loop runner's deterministic logic into a testable library and adds a fixture suite that exercises it, closing the runner's long-standing "no fixture test" gap.

- **New `tools/spec-loop/lib.sh`**: pulls prompt assembly, harness command rendering, agent launch, and `.last-sync` marker helpers out of `loop.sh` into standalone, sourceable functions. `loop.sh` now sources them (net ~85 fewer lines there), so the same code paths the runner uses are the ones under test.
- **New `tools/spec-loop/tests/test_runner_fixtures.sh`**: deterministic Bash fixtures covering prompt assembly per beat (build vs update scoping), harness argv construction across agents, and the `.last-sync` marker write/branch-name helpers. No network, no agent, no `uv`.
- **`spec-validator` wiring**: adds check #10, `validate_spec_loop_runner_fixtures`, which runs the fixture suite via `bash` when the file is present and surfaces any non-zero exit as a validation violation. `spec-validate` now executes the runner tests as part of normal validation, with a pytest added for the new validator branch.
- **Spec update** (`specs/spec-loop-runner.md`): records `lib.sh` and the test file under *Where it lives* and *source*, adds them to the Validation block (`shellcheck`), and removes the now-resolved *Known gap* about missing fixture tests.

The runner previously had no deterministic test for prompt assembly, harness command construction, or marker handling, so refactors risked silently breaking how prompts and agent invocations are built. Splitting the logic into `lib.sh` makes it unit-testable, and routing the suite through `spec-validate` keeps it running automatically rather than by hand.

The bash suite passes (`spec-loop runner fixtures: OK`); the pytest covers the new validator hook.

## Type of change

- [X] Python package (`tools/*/` with `pyproject.toml`)
- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [X] Other:  bash suite passes

### DIFF
--- a/tools/spec-loop/lib.sh
+++ b/tools/spec-loop/lib.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+# SPDX-License-Identifier: Apache-2.0
+#   https://www.apache.org/licenses/LICENSE-2.0
+#
+# Deterministic helpers for tools/spec-loop/loop.sh. Keep this file free of
+# live agent execution so fixture tests can pin runner behaviour without
+# launching a headless harness.
+
+spec_loop_append_tooling_source_context() {
+    local mode=$1 base=$2 tooling_ref=$3
+
+    cat <<EOF
+
+## Tooling source — read the plan and specs from here
+
+This iteration builds on the integration base \`$base\`, which does
+NOT carry the spec-loop tooling. The plan and specs live on the
+control branch \`$tooling_ref\`. Read them from there with \`git show\`,
+never from the working tree:
+
+\`\`\`
+git show $tooling_ref:tools/spec-loop/IMPLEMENTATION_PLAN.md
+git ls-tree -r --name-only $tooling_ref tools/spec-loop/specs/
+git show $tooling_ref:tools/spec-loop/specs/<file>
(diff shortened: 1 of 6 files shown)


potiuk (approved):
Re-reviewed after I rebased this onto current `main` and force-pushed — a clean, content-preserving replay onto the commits that landed since (no changes to your diff).

The one concern from the earlier review — reconciling the `--model` handling against main's launch code — resolved automatically on rebase: `loop.sh` now delegates to `lib.sh`'s `spec_loop_launch_agent`, which uses the safe conditional `[ -n "$model" ] && model_args=(--model "$model")`, and the MODEL defaulting in `loop.sh` (empty for non-claude harnesses → agent default) feeds it correctly. So there's no unconditional `--model ""` path.

Verified: the runner fixture suite passes (`spec-loop runner fixtures: OK`), spec-validator pytest + spec-validate green, CI 35/35 including shellcheck. Faithful refactor with genuine test coverage. LGTM.

PR #745, opened 2026-07-05
feat(skill): add onboarding-concierge skill with eval suite

justinmclean:
## Summary

Adds the onboarding-concierge Agentic Mentoring skill (item 6 of the implementation plan). The skill answers newcomer "how do I contribute here" questions from the project CONTRIBUTING.md and configured docs: classifies each question (setup / workflow / first-issue / architecture / security / out-of-scope), retrieves the relevant guide excerpt, and drafts a concise reply in the Agentic Mentoring teaching register. Out-of-scope, design, and security questions route to a human maintainer. Read-only: no files written, no comments posted without maintainer action.

Ships:
- skills/onboarding-concierge/SKILL.md with runtime loop, question-classification table, answer-drafting rules, and hand-off protocol
- Relay symlinks in .agents/, .claude/, .github/
- projects/_template/onboarding-concierge-config.md
- 10-case eval suite (question-classify x 6, answer-draft x 4) using step-config.json so prompt drift is caught automatically
- docs/modes.md, docs/mentoring/spec.md, docs/labels-and-capabilities.md updated (Mentoring count 5 to 6)
- tools/skill-evals/README.md updated

Validation: skill-and-tool-validate exits 0; eval runner picks up all 10 cases with correct structure.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:


potiuk (commented):
Nice skill — the design is sound, and routing security/architecture questions to a human hand-off rather than answering is exactly right. A few things worth addressing, mostly around the config contract and eval coverage rather than the runtime logic.

**1. `out_of_scope_topics` is a declared config key with no consumer.** The adopter-contract table says the skill reads `out_of_scope_topics` ("topic keywords that trigger automatic hand-off, e.g. `security`, `deprecation`, `license`"), but the Runtime loop and Question-classification steps use the fixed six-category table and never read it. So an adopter who adds `license` to `out_of_scope_topics` expecting licence questions to hand off gets no effect — classification has no `license` category and never consults the key. It's also missing from the template's key table (`projects/_template/onboarding-concierge-config.md` defines only `contributing_guide_url` / `maintainer_team_handle` / `ai_attribution_footer`). Either wire the key into the classify step or drop it from the contract.

**2. No eval case for the `out-of-scope` category.** It's one of the three hand-off-triggering categories and the subtlest to classify (vague/open-ended questions), but the six classify cases cover only setup / workflow / first-issue / architecture / security / injection. A prompt change that mis-routes "what should I work on?" into `workflow` (and drafts an answer) would pass the whole suite. Worth a `case-*-out-of-scope` fixture expecting `hand_off: true`.

**3. The answer-draft hand-off case doesn't verify the answer body is suppressed.** `answer-draft/fixtures/output-spec.md` requires `answer` be absent/null when `answer_drafted: false`, but the expected.json omits `answer` and the comparator ignores it (per the suite README). So a regression that sets `answer_drafted: false, hand_off: true` **but still emits a full drafted `answer`** — leaking an AI reply on a security/architecture hand-off — passes `case-3-handoff`. An assertion that `answer` is empty on the hand-off path would close the gap this skill exists to prevent.

**4. Vestigial `<upstream>` / `gh` reference (minor).** The placeholder-convention comment declares `<upstream>` and says "before running any `gh` command below," but the skill has no `gh` commands and never references `<upstream>` — it reads local `<project-config>` and `CONTRIBUTING.md` only. Looks like copy-paste from a gh-using template; trimming it to the placeholders actually used would avoid adopters configuring an upstream repo the skill doesn't touch.

**5. Typo (minor):** "A Agentic Mentoring" → "An Agentic Mentoring" in the Status line.

`mode: Mentoring` is fine (it's conventional across the skill set). I didn't touch the `docs/vendor-neutrality.md` counts since that block is generated — probably worth confirming those against the generator rather than by hand.

justinmclean:
All done:

out_of_scope_topics now actually does something. It was a config key the skill advertised but never read, so adding a topic like license had no effect. Added a runtime step that reads the list and forces a hand-off on a keyword match. Also added it to the template's key table.

Added an eval case for out-of-scope questions. The classify suite tested every hand-off category except this one. New case checks that a vague "what should I work on" question gets handed off instead of drafted.

The hand-off case now checks the answer is actually suppressed. Before, a regression could emit a full drafted answer on a security/architecture hand-off and still pass. Added an assertion that fails if any answer body leaks. Verified: empty answer passes, leaked answer fails.

Minor: removed a stale <upstream>/gh reference the skill doesn't use, fixed "A Agentic" to "An Agentic", and updated the eval README counts.

All eval and python tests pass.

PR #756, opened 2026-07-05
feat(tools/skill-and-tool-validator): add SOFT SKILL.md line-limit check

justinmclean:
## Summary

Adds check #20 to the skill-and-tool-validator: a SOFT advisory that flags
any `SKILL.md` entrypoint exceeding `SKILL_LINE_LIMIT` (500 lines), per
PRINCIPLES.md §14.

## Why

Oversized SKILL.md entrypoints should move reference material into sibling
markdown files linked one level deep. This check keeps new oversized
entrypoints from being merged unnoticed, while staying advisory so existing
over-limit skills can be migrated deliberately rather than blocking unrelated
work.

## Behaviour

- Category `skill-line-limit`, registered as SOFT: surfaces as a stderr
  advisory and does not fail the run unless `--strict` is passed.
- Only applies to files named `SKILL.md`.
- Message points to PRINCIPLES.md §14 and names the actual line count.
- Currently flags 19 pre-existing over-limit skills as advisories.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [X ] `prek run --all-files` passes
- [ ] Other:

### DIFF
--- a/tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
+++ b/tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
@@ -17,7 +17,7 @@
 
 """Validate framework skill definitions.
 
-This module validates seventeen aspects of every skill under
+This module validates eighteen aspects of every skill under
 skills/:
 
 1. YAML frontmatter — every SKILL.md must have a valid frontmatter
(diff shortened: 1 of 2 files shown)


potiuk (approved):
Nice check — a clean, well-scoped SOFT advisory that stops new oversized `SKILL.md` entrypoints from slipping in, while leaving the existing over-limit skills as advisories to migrate deliberately rather than blocking unrelated work. The implementation is tidy (`SKILL.md`-only guard, clear message pointing at PRINCIPLE 14) and the test coverage is thorough.

A couple of things I did before merging (you're preserved as commit author):

- **Rebased onto latest main as a single clean commit.** The branch had picked up a good chunk of already-merged history (fossil / vcs / `uv.lock`), so a plain `git rebase` replayed all of that and conflicted. The net two-file change applied cleanly to current main via a 3-way apply, so the check is confirmed to still work against today's tree.
- **Removed the duplicate test.** `test_under_limit_no_violation` was asserting on exactly `SKILL_LINE_LIMIT` (500), identical to `test_at_limit_no_violation` — it now uses a genuinely under-limit count (`SKILL_LINE_LIMIT - 100`) so "under" and "at limit" are distinct cases.

Ran the full validator pytest suite + the validator itself locally — all green, and the new advisory fires on the 21 real over-limit skills as expected. Merging. Thanks!

PR #763, opened 2026-07-06
feat(validator): add no-telemetry-import check (PRINCIPLE 10 guarantee)

justinmclean:
## Summary

Add check #21 (SOFT advisory) to skill-and-tool-validator that flags network-calling imports (requests, httpx, aiohttp, urllib.request, http.client, socket) in substrate:* tool source files under tools/*/src/.

Only contract:* adapter tools and the egress-gateway proxy are declared egress surfaces; all other substrate tools must stay network-free to uphold PRINCIPLE 10's guarantee of zero default outbound calls.

Also add a Declared egress surfaces section to tools/egress-gateway/tool.md that states the default-zero guarantee, lists the declared egress surfaces, and cross-references the new validator check.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

## RFC-AI-0004 compliance

### DIFF
--- a/tools/egress-gateway/tool.md
+++ b/tools/egress-gateway/tool.md
@@ -11,6 +11,7 @@
   - [Relationship to RFC-AI-0003](#relationship-to-rfc-ai-0003)
   - [How adopters consume this tool](#how-adopters-consume-this-tool)
   - [What this tool is NOT for](#what-this-tool-is-not-for)
+  - [Declared egress surfaces](#declared-egress-surfaces)
   - [Failure modes](#failure-modes)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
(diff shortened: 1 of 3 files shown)


tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
```diff
@@ -3554,6 +3589,66 @@ def validate_skill_line_limit(path: Path, text: str) -> Iterable[Violation]:
...
+            except OSError:
+                readme_text = ""
+            cap_match = TOOL_CAPABILITY_RE.search(readme_text)
+            if cap_match is not None:
+                raw = cap_match.group(1).strip()
+                entries = [e.strip() for e in raw.split("+") if e.strip()]
+                if any(e.startswith(_ADAPTER_CONTRACT_PREFIX) for e in entries):
+                    continue  # declared contract:* adapter — network is expected
+
+        src_dir = tool_dir / "src"
```
potiuk, line 3624: **Scope vs. the stated guarantee.** The scan only descends into `tools/<name>/src/`, but several substrate tools keep Python at the tool root outside `src/` — e.g. `pr-management-stats/{dashboard,reference}.py`, `security-tracker-stats-dashboard/{fetch_roster,fetch_prs,render}.py`, `dashboard-generator/reference.py`. A network import in any of those isn't caught, yet `egress-gateway/tool.md` promises "a substrate tool that accidentally grows a network import is caught before it is merged." The `fetch_*.py` files are exactly the surface most likely to grow egress. Consider `rglob`-ing all `*.py` under `tool_dir` (skipping `tests/`) rather than only `src/` — or, if src-only is intentional, soften the guarantee wording in `tool.md` to match.

tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
```diff
@@ -361,6 +370,27 @@
...
+# Network-calling import patterns that must not appear in substrate tool source.
+# Each entry is (compiled line-level regex, human-readable library name).
+_NETWORK_IMPORT_PATTERNS: list[tuple[re.Pattern[str], str]] = [
+    (re.compile(r"^\s*(?:import|from)\s+requests\b"), "requests"),
+    (re.compile(r"^\s*(?:import|from)\s+httpx\b"), "httpx"),
+    (re.compile(r"^\s*(?:import|from)\s+aiohttp\b"), "aiohttp"),
+    (re.compile(r"^\s*(?:import|from)\s+urllib\.request\b"), "urllib.request"),
+    (re.compile(r"^\s*from\s+urllib\s+import\s+(?:\w+\s*,\s*)*request\b"), "urllib.request"),
+    (re.compile(r"^\s*(?:import|from)\s+http\.client\b"), "http.client"),
+    (re.compile(r"^\s*import\s+socket\b"), "socket"),
```
potiuk, line 390: This pattern is `^\s*import\s+socket\b`, so `from socket import socket, AF_INET` slips through (the other libs use `(?:import|from)` and are fine). Worth a `^\s*from\s+socket\s+import\b` companion. Separately, the list is deliberately tight — `urllib3`, `smtplib`, `ftplib`, `boto3`, `paramiko`, `grpc`, google-cloud all bypass; fine for a heuristic SOFT check, but a `# non-exhaustive` comment would set expectations. (Minor false-positive note: a tool using `socket` for local IPC / `socket.getfqdn()` is flagged as "telemetry" — tolerable given SOFT + `--strict`-only.)

tools/egress-gateway/tool.md
```diff
@@ -95,6 +96,44 @@ layers.
...
+   services, VCS hosts, CVE authorities, etc.) on behalf of skill steps that
+   ask for it.  The `egress-gateway` proxy itself is also a declared surface;
+   all other `substrate:*` tools are network-free by design.
+
+2. **The validator enforces the invariant.**  `tools/skill-and-tool-validator/`
+   check #21 (`no-telemetry-import`) scans the source of every `substrate:*`
+   tool for network-calling imports (`requests`, `httpx`, `aiohttp`,
+   `urllib.request`, `http.client`, `socket`) and flags any hit as a SOFT
+   advisory.  A substrate tool that accidentally grows a network import is
+   caught before it is merged.
```
potiuk, line 120: This is the strongest form of the guarantee, but check #21 currently scans only each tool's `src/` dir — tool-root `.py` files aren't scanned — and the check is **SOFT** (advisory unless `--strict`), so an accidental import merges cleanly without `--strict`. Either broaden the scan (see the `__init__.py` comment) or soften this to "flagged by check #21 (SOFT advisory)" so the doc doesn't over-promise.

potiuk (commented):
Thanks @justinmclean — a useful guardrail for PRINCIPLE 10, and the happy-path implementation is correct (multi-capability `+`-split handling, egress-gateway skip, line-anchored regex so comments/strings don't match). My one substantive point is that the scan scope (`src/` only) is narrower than the guarantee `egress-gateway/tool.md` states — details inline, along with a couple of pattern/coverage nits. All non-blocking; the check is safe to ship as a SOFT advisory, but I'd like the scope and the doc wording to agree before it lands.

---
> *This review was drafted by an AI-assisted tool and confirmed by a Magpie maintainer. The findings below are observations, not blockers; a Magpie maintainer — a real person — will take the next look at the PR. If you think a finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>

tools/skill-and-tool-validator/tests/test_validator.py
```diff
@@ -4807,3 +4807,169 @@ def test_non_skill_md_not_checked(self, tmp_path: Path) -> None:
...
+# ---------------------------------------------------------------------------
+
+
+from skill_and_tool_validator import (  # noqa: E402
+    NO_TELEMETRY_CATEGORY,
+    validate_no_telemetry_imports,
+)
+
+
+class TestValidateNoTelemetryImports:
```
potiuk, line 4823: Nice happy-path suite. A few gaps worth closing: the non-trivial `from urllib import …, request` regex is untested; `aiohttp` and `http.client` have patterns but no test; there's no test for a multi-capability tool (`substrate:sandbox + contract:tracker`) being skipped; and `test_no_src_directory_skipped` actually codifies the src-only behavior — a test asserting a **tool-root** `.py` import is caught would pin down the scope finding either way.

potiuk:
Merging — the scope-vs-guarantee points from my review are non-blocking for this SOFT advisory check and are tracked as a follow-up in #783. Thanks @justinmclean.

PR #778, opened 2026-07-08
Update LICENSE and NOTICE file

justinmclean:
## Summary

Update the LICENSE and NOTICE files as license information goes in LICENSE, not NOTICE, and there is no need to duplicate the information, and the relevant information from included ASF project NOTICE files needs to be included in LICENSE.

Note that some of this information, now in LICENSE, is really project history and doesn't need to be included there; only the relevant parts of the ASF project NOTICE files do.

## Type of change

- [ ] Groovy reference impl
- [X] Other:

## Test plan

- [ ] Other:


potiuk (approved):
Approving. Verified this **loses no attribution**: the Airflow/Groovy and skill-creator credits are relocated from `NOTICE` into `LICENSE`, not deleted; `NOTICE` is trimmed to the minimal ASF form (copyright + boilerplate line); the standard Apache-2.0 text in `LICENSE` is intact; and the removed `skills/write-skill/LICENSE.txt` was a redundant full Apache-2.0 copy (root `LICENSE` already carries it) with its dangling reference cleaned up. Confirmed as a PMC-level NOTICE/LICENSE call. Thanks @justinmclean.

---
>

LICENSE
```diff
@@ -200,3 +200,32 @@
...
+  inside and outside the ASF.
+- Apache Groovy (https://groovy.apache.org/) — contributions from
+  the Groovy PMC at the early stage of this project.
+
+Apache Airflow
+Copyright 2016-2026 The Apache Software Foundation
+
+Apache Groovy
+Copyright 2003-2026 The Apache Software Foundation
+
```
CalvinKirs, line 223: I don't think the "Contributions from other ASF projects" block should've just been moved as-is.

Groovy: this traces back to #146. Paul (@paulk-asert) built these skills on the Groovy side first, then contributed the same work to Magpie himself, under his own CLA. Same author both times — this isn't "the Groovy project" as some separate copyright holder handing us code, it's one person contributing his own work to two ASF projects he's part of. I don't think there's anything here that actually needs crediting.

Airflow: honestly not sure what Airflow files/code we actually pulled in here — could someone clarify what this is referring to? But regardless of what it turns out to be, one thing seems off procedurally: if it's real enough to need a mention in LICENSE, then per AL 2.0 §4(d) it should also show up in NOTICE (that's how notice-bubbling from an included work's own NOTICE file is supposed to work — LICENSE and NOTICE aren't interchangeable for this). Right now it's only in LICENSE and NOTICE was trimmed to nothing, which is inconsistent — either it needs to be in both, or (if it turns out to be more like project-naming history — this repo used to be apache/airflow-steward before #593 renamed it) it doesn't need to be in either.

justinmclean: So it is a bit verbose and strictly not needed (the history part), but I did intend to put those copyright statements in NOTICE, not LICENSE, so that’s on me. However, they are not strictly required there either, as they are an ASF project - there is a legal dicuss discussion on this. It's just good to include them for completeness.

For the contributions from Paul and from Airflow - it is unknown if other people worked on those contributions so I would err on the side of caution and include them.

justinmclean: The text from "Contributions from other ASF projects" should not be in NOTICE, btw, as license information doesn't belong in NOTICE.

CalvinKirs: Before we settle this — mind if we just check with Jarek and Paul what's actually going on here? Otherwise LICENSE just keeps piling up stuff nobody can double-check down the line.

If we genuinely pulled in some Airflow code, that's totally fine to mention in LICENSE, but we'd also need to copy the relevant bit of Airflow's own NOTICE into our NOTICE file— right now only LICENSE got touched, so it'd be incomplete.

If this is really just us saying thanks — like, the repo used to be airflow-steward before it got renamed, and Paul's Groovy skills were his own contribution under CLA rather than code copied out of apache/groovy — then none of this belongs in LICENSE or NOTICE anyway, it should just live in the README as a normal shoutout.

@potiuk @paulk-asert , Could you two clarify which case this actually is? Can send a follow-up PR once we know.

justinmclean: Technically, we did pull in some code from airflow, as the airflow-steward repo was under the control of the Airflow PMC.

potiuk: Yes - the code was transferred from "stewardship" of Airflow and Groovy - but I think the "ownership" of the code was not changed - Apache Software Foundation still owns the copyright. As far as I understand (and well, that's somewhat of a legalese that probably people will have different opinion on) - NOTICE comment is really not needed when ownership does not change. 

I consider the Airflow and Groovy mentionig more of attribution than notice.

And yeah - putting it LICENCE is not good either in retrospect.

I did some chatting with Claude :) ... and this is the proposal we both came up with:


> The whole framing of "contributions from other ASF projects" is the confusion. There is no separate copyright holder here. Airflow, Groovy, airflow‑steward, and Magpie are all Copyright The Apache Software Foundation, all AL 2.0. Code moving from one ASF project to another is the same rightsholder — it's not a third‑party inclusion.
>
> That resolves every branch of Calvin's question:
>
> - AL 2.0 §4(d) NOTICE‑bubbling doesn't apply — that mechanism exists to carry required attributions from an included third‑party work's own NOTICE. Neither Airflow's nor Groovy's NOTICE has anything to bubble beyond the standard ASF line already present. So Justin's "must also go in NOTICE" concern is moot: nothing needs to.
> - It definitely doesn't belong in LICENSE — it's neither a license grant nor a third‑party license text. Justin conceded moving it there was his mistake.
> - Groovy: Paul's own CLA contribution → nothing to attribute as third‑party. Remove.
> - Airflow: genuine lineage (this was an Airflow‑PMC repo), but still same rightsholder → not legally required in either file. It's project history.
> - The fabricated‑looking Apache Groovy / Copyright 2003‑2026 span #778 added to LICENSE is particularly wrong — we ship no file from apache/groovy that it corresponds to.
> 
> Recommendation: drop the entire "Contributions from other ASF projects" block from LICENSE and don't restore it to NOTICE. If you want to preserve the lineage, it's a README / MISSION.md shout‑out. Keep the separate "Third‑party content" block (Julius Brussee / awesome‑claude‑skills) — that is the one genuine non‑ASF contributor and correctly belongs in LICENSE.

I would do that:

* Remove the LICENCE addition
* Do not add it back to NOTICE
* Move the attribution to README

@CalvinKirs @justinmclean @paulk-asert - WDYT ? I am also happy to cancel rc1 and let others know the result of that interesting discussion in devlist VOTE thread :D 

Good learning for everyone.

CalvinKirs: Same logic as the Groovy case, actually — if all of the Airflow-derived code was originally written by Jarek himself, then this is just him contributing his own work to another ASF project he's part of, same as Paul did with Groovy. No separate copyright holder, so nothing needs to go in LICENSE or NOTICE at all.

But if the code includes stuff written by other Airflow contributors too, that's different — then it really is someone else's copyrighted work getting pulled in; we'd need to bubble the relevant bits of Airflow's NOTICE into ours.

https://infra.apache.org/licensing-howto.html#bundle-asf-product

CalvinKirs: > Technically, we did pull in some code from airflow, as the airflow-steward repo was under the control of the Airflow PMC.↳

Honestly... I'm not sure which way to go here — should we be going off governance (who officially controls the repo) or actual code provenance (what files really got copied)?

And if it's governance, whose NOTICE do we even go by? Airflow's got a ton of repos, not just one. And to make it messier — what if that NOTICE file has a bunch of other stuff bundled into it too?

potiuk: Yeah - there were other people from Airflow (and even external contributors) that contributed to airflow-steward. 

Is there any **issue** with adding the Airflow and Groovy to NOTICE - where it was before? I would be rather over-cautious and put extra information - even if it makes the NOTICE file longer and potentially someone bubbling it up having to carry the complexity.

I do not think we should optimize here for "the smallest and most concise" set of attributions, but "one that protects everyone's rights even if we have some extra - possibly unnecesary stuff in".


So ... for the sake of pragmatism vs. purism :), how about we come back to what it was and move the Airflow and Groovy back to NOTICE ?

It **may** be not necessary, but if it does not hurt and saves us some unnecesary discussions - I am all for it.

justinmclean:
This is not correct - it applies to ASF projects as well, just not as strongly.
AL 2.0 §4(d) NOTICE‑bubbling doesn't apply — that mechanism exists to carry required attributions from an included third‑party work's own NOTICE. Neither Airflow's nor Groovy's NOTICE has anything to bubble beyond the standard ASF line already present. So Justin's "must also go in NOTICE" concern is moot: nothing needs to.

Policy explicitly states that the copyright line needs to be considered.

This is also not correct:
It definitely doesn't belong in LICENSE — it's neither a license grant nor a third‑party license text. Justin conceded moving it there was his mistake.

License information does belong in LICENSE and we've had no license grants.

And again, this is incorrect:
Recommendation: drop the entire "Contributions from other ASF projects" block from LICENSE and don't restore it to NOTICE. 

License information should not go in NOTICE. Also, NOTICE files need to be kept as short as possible, as it adds obligations on anyone who includes Magpie in their software.

Regarding what to include in the NOTICE, only the minimum required is needed, so I very much doubt that the whole airflow NOTICE is needed, and only the copyright line needs to be considered.

Putting something that might not be needed in LICENSE or NOTICE is a documentation issue, not putting it in is an ASF policy issue.

potiuk:
> This is not correct - it applies to ASF projects as well, just not as strongly. AL 2.0 §4(d) NOTICE‑bubbling doesn't apply — that mechanism exists to carry required attributions from an included third‑party work's own NOTICE. Neither Airflow's nor Groovy's NOTICE has anything to bubble beyond the standard ASF line already present. So Justin's "must also go in NOTICE" concern is moot: nothing needs to.
> 
> Policy explicitly states that the copyright line needs to be considered.
> 
> This is also not correct: It definitely doesn't belong in LICENSE — it's neither a license grant nor a third‑party license text. Justin conceded moving it there was his mistake.
> 
> License information does belong in LICENSE and we've had no license grants.
> 
> And again, this is incorrect: Recommendation: drop the entire "Contributions from other ASF projects" block from LICENSE and don't restore it to NOTICE.
> 
> License information should not go in NOTICE. Also, NOTICE files need to be kept as short as possible, as it adds obligations on anyone who includes Magpie in their software.
> 
> Regarding what to include in the NOTICE, only the minimum required is needed, so I very much doubt that the whole airflow NOTICE is needed, and only the copyright line needs to be considered.
> 
> Putting something that might not be needed in LICENSE or NOTICE is a documentation issue, not putting it in is an ASF policy issue.

So... What do we do ?

potiuk:
Concrete proposal that - I think - aims to address all concerns here https://github.com/apache/magpie/pull/821 - let's discuss there :)

PR #823, opened 2026-07-11
fix(agent-guard): make the fresh-clone hook guard cross-platform (Windows)

milamberspace:
## Summary

Follow-up to #786 (merged): that PR guarded the committed agent-guard hook command with a POSIX shell conditional — `[ -f … ] && python3 … || true` — so the hook no-ops on a fresh clone before `/magpie-setup` has synced `agent-guard.py` in. But that form only works in a POSIX shell; on **Windows** (PowerShell, no Git Bash) the committed hook breaks.

Replace it with an inline `python3 -c` existence check that is cross-platform and needs no shell conditional:

```
python3 -c "import os,sys,subprocess; p=os.path.join(os.environ.get('CLAUDE_PROJECT_DIR',''),'.claude','hooks','agent-guard.py'); sys.exit(subprocess.call([sys.executable,p]) if os.path.isfile(p) else 0)"
```

- no-ops (exit 0) when the script is absent (fresh clone);
- runs the guard and propagates its exit code (so a deny still blocks) when present;
- reads `CLAUDE_PROJECT_DIR` from the environment in Python, so it doesn't depend on `$VAR` vs `%VAR%` shell expansion.

Updated in the two places that ship the snippet: `skills/setup/adopt.md` and `tools/agent-guard/README.md`.

Found by an adopter (Apache JMeter) reviewing the same change downstream — the reviewer asked "Does this work with windows?" on the adoption PR.

## Type of change

- [x] Bug fix (docs / committed template)

## Test plan

Doc / template change. The guarded command is valid on both POSIX and Windows (pure `python3 -c`, no shell conditional); it no-ops when the script is missing and runs + propagates the guard's exit code when present.

### DIFF
--- a/skills/setup/adopt.md
+++ b/skills/setup/adopt.md
@@ -1292,16 +1292,17 @@ Four passes, in this order:
 
      ```json
      { "matcher": "Bash", "hooks": [ { "type": "command",
-       "command": "[ -f \"$CLAUDE_PROJECT_DIR/.claude/hooks/agent-guard.py\" ] && python3 \"$CLAUDE_PROJECT_DIR/.claude/hooks/agent-guard.py\" || true",
+       "command": "python3 -c \"import os,sys,subprocess; p=os.path.join(os.environ.get('CLAUDE_PROJECT_DIR',''),'.claude','hooks','agent-guard.py'); sys.exit(subprocess.call([sys.executable,p]) if os.path.isfile(p) else 0)\"",
        "timeout": 30 } ] }
      ```
 
-     The `[ -f … ] && … || true` guard makes the hook a **no-op on a
-     fresh clone** — before `/magpie-setup` has synced `agent-guard.py`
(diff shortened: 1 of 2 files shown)


milamberspace:
Heads-up for maintainers: this cross-platform fix may be **superseded by #824**.

That issue proposes writing the agent-guard hook into the gitignored, per-machine `.claude/settings.local.json` at setup (alongside the script) rather than committing it to `.claude/settings.json`. If that lands, the wiring only ever exists once the script is present, so no fresh-clone guard is needed at all — and the command reduces to a plain `python3 ".../agent-guard.py"`, which is already cross-platform. In other words #824 would make both the #786 guard and this `python3 -c` rewrite unnecessary.

Happy to close this in favour of #824, or keep it as the interim fix if you would rather ship the smaller change first — your call.

potiuk:
Oh... Windows support :)

potiuk (approved):

PR #825, opened 2026-07-11
feat(secure-setup): track latest claude-code behind a hard min_version floor

potiuk:
## What

The agent runtime (`claude-code`) was pinned to an exact version in `pinned-versions.toml`. For the runtime **specifically**, always-latest is the more secure posture — each release carries the newest permission-rule / sandbox / prompt-injection fixes, so pinning it *increases* the framework's security lag. This replaces the exact pin with a **`min_version` floor**.

## Changes

- **`pinned-versions.toml`** — `[tools.claude-code]` now declares `min_version` (not an exact `version`) and installs `@latest`. `bubblewrap` / `socat` stay exact-pinned (low-level sandbox primitives where reproducibility matters more than chasing latest).
- **`setup-isolated-setup-verify` check 5** — at/above floor → ✓; **below floor while running under Claude Code → hard fail (✗)**, not a warning; below floor on a non-Claude harness → ⚠.
- **`setup-isolated-setup-install` / `-update`** — install/track `@latest` and enforce the floor.
- **`check-tool-updates.sh`** — skips `min_version`-only tools (claude-code is no longer a bump candidate).
- **Docs** — `secure-agent-setup.md` (renamed "Required tools" section + new "Raising the claude-code floor"), `-internals.md`, `prerequisites.md`, agent-isolation `README.md`, `RFC-AI-0002.md` (with a History note). Also fixes two `tool.md` anchors broken by the heading rename.
- **Eval fixtures** — 16 `setup-isolated-setup-verify` fixtures reworked to the floor / hard-fail model.

## Validation

`skill-and-tool-validate`, `ruff`, `mypy`, `pytest`, and the lychee link check all pass.

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -95,11 +95,14 @@ environment. The permission + sandbox posture is enforced for both
 harnesses — a `PreToolUse` hook / `tool.execute.before` plugin
 (`agent-guard`), plus `permission-audit` / `sandbox-lint` for the
 Claude `settings.json` and OpenCode `opencode.json` policies. The
-required system tools (`bubblewrap`, `socat`, and the agent CLI
-itself — `claude-code` or `opencode`) are pinned with a 7-day
+sandbox primitives (`bubblewrap`, `socat`) are pinned with a 7-day
 upstream-release cooldown, mirroring the same convention the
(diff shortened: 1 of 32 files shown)


potiuk:
Pinning claude is useless :D

PR #826, opened 2026-07-11
feat(setup): offer all skill families + MCP servers at adopt/upgrade

potiuk:
## What

Adoption only ever offered three opt-in families (`security`, `pr-management`, `issue`). `release-management`, `repo-health`, `pairing`, `mentoring`, and `contributor-growth` had **no install path** because family membership was inferred from the skill-name prefix, which cannot express families that span several prefixes (e.g. `repo-health` covers `ci-`/`workflow-`/`dependency-`/`license-`/`flaky-`).

## Changes

- **`family:` frontmatter** — every skill now declares its family, validated against `ALLOWED_FAMILIES` in `skill-and-tool-validator`.
- **`/magpie-setup` reads `family:`** from the snapshot instead of prefix-globbing — Golden rule 8, `adopt.md` Step 5 & Step 8, `upgrade.md`, `worktree-init.md`, `verify.md`.
- **`adopt.md` Step 5 rewritten** into a single install choice offering all **eight** opt-in families **and** the three MCP servers (`ponymail`, `apache-projects`, `gmail-plaintext`) in one prompt; `setup` / `utilities` stay always-on (never prompted).
- **`setup-status/collect_status.py`** now reads `family:` (was a prefix heuristic) — verified against the live repo, 0 skills unclassified.
- **`README.md`** families table updated: 10 families, `pairing` added, corrected counts.

## Validation

`skill-and-tool-validate`, `ruff`, `mypy`, `pytest`, and the lychee link check all pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -162,25 +162,31 @@ for the full flow.
 
 ## Skill families
 
-Eight skill families ship in the framework, all at `experimental` or
-`stable`. Pick whichever families the adopter wants to use; symlinks for
-the picked families land in the adopter's skill directory.
+Ten skill families ship in the framework, all at `experimental` or
+`stable`, and each skill declares its family in a `family:` frontmatter
+key. At adoption (and on every upgrade), `/magpie-setup` offers the
+**opt-in** families — and the optional **MCP servers** (`ponymail`,
+`apache-projects`, `gmail-plaintext`) — in a single install choice;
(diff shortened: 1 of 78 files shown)


potiuk (commented):
I noticed that at adoption we did **not** offer all famiies of skills :D

PR #829, opened 2026-07-11
fix(release): produce a clean source-only tarball and gate it in verify-rc

potiuk:
## Why

The 0.1.0 **rc1** source zip drew a binding **-1**. Four defects, one packaging root cause:

1. **181 `.pyc` in 61 `__pycache__` dirs** — the tarball was `zip -r`'d from a working tree that had run tests, not exported from the tag.
2. **Unclean tree at package time** — same root cause.
3. **138 dangling symlinks** — `.gitattributes` export-ignored `.agents/`, but the committed `.claude/skills/*` and `.kiro/skills/*` relays point through `.agents/skills/* → skills/*`, so every relay dangled once `.agents/` was stripped.
4. **spec-validator / skill-and-tool-validator failures** — the same stripping broke references from shipped files into `.github/` templates, `projects/_template/.gitignore`, and `.claude/skills/magpie-*/SKILL.md`.

## Changes

**Packaging — `.gitattributes`**
- Ship `.agents/` so every relay symlink resolves to `skills/` (**0 dangling**).
- Strip only genuine CI/dev under `.github/` (`workflows/`, `dependabot.yml`); keep `ISSUE_TEMPLATE/` and `PULL_REQUEST_TEMPLATE.md` (linked by shipped skills).
- Anchor root-only dotfiles with `/` so identically-named files deeper in the tree (e.g. `projects/_template/.gitignore`) still ship.

**Gate — `skills/release-verify-rc/SKILL.md`**
- Step 6 always surfaces and hard-fails on `.pyc` / `__pycache__`.
- **New Step 7 — source-tree integrity**: no dangling symlinks, and run `symlink-lint` + `skill-and-tool-validate` + `spec-validate` against the *unpacked tarball*; hard-fail on broken internal references. Renumbered later steps; added failure-mode rows.
- `projects/magpie/release-build.md`: reinforce "always `git archive <tag>`, never `zip -r`".

## Verification

Against a real `git archive` of the fixed tree (extracted):

```
prohibited (.pyc / __pycache__ / binaries): 0
dangling symlinks:                          0
symlink-lint / skill-and-tool-validate / spec-validate: all exit 0
no .git (clean export)
```

The exact validators that failed on rc1 now pass on the exported tarball, and `release-verify-rc` would catch a regression of any of these before the `[VOTE]`.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -20,16 +20,37 @@
 # apache-magpie-<version>-source.zip) honours `export-ignore`, so the
 # paths below are kept out of the signed source .zip that the [VOTE]
 # votes on. Keep this list to VCS/CI/editor metadata — never source.
+#
+# IMPORTANT — do NOT export-ignore a directory that shipped files
+# still reference (by markdown link, `test -f`, or symlink). The
+# committed agent-view symlink dirs (`.claude/skills/`, `.github/skills/`,
+# `.kiro/skills/`) relay through `.agents/skills/*` → `skills/*`, and
+# specs test those paths, so `.agents/` MUST ship or every relay dangles
+# in the archive. Likewise `.github/ISSUE_TEMPLATE/` and
+# `.github/PULL_REQUEST_TEMPLATE.md` are linked by skills, and
+# `projects/_template/.gitignore` is example content — so only the true
+# CI/dev entries under `.github/` are stripped, and `.gitignore` is
+# anchored to the repo root. `release-verify-rc` re-checks the unpacked
+# tarball (symlink-lint + validators + no-.pyc) so a regression here
+# fails the RC before the [VOTE].
 
-.gitattributes      export-ignore
-.gitignore          export-ignore
-.github/            export-ignore
+# Root-only VCS / dev-tool metadata (anchored so identically-named files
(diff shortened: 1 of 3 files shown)


justinmclean:
Note this has introduced a bug whcih causes this to fail:
uv run --project tools/skill-and-tool-validator --group dev skill-and-tool-validate
PYTHONPATH=tools/skill-evals/src python3 -m skill_evals.runner tools/skill-evals/evals/

I'm just fixing that

sebbASF:
> Note this has introduced a bug whcih causes this to fail: uv run --project tools/skill-and-tool-validator --group dev skill-and-tool-validate PYTHONPATH=tools/skill-evals/src python3 -m skill_evals.runner tools/skill-evals/evals/
> 

Does that indicate a bug or omission in the system?
I would have expected PR checks to pick up the failure noted here.

potiuk:
> Does that indicate a bug or omission in the system?
> I would have expected PR checks to pick up the failure noted here.

Those checks are run with local agents - for people who use agents when they run PRs locally - when relevant part changes. Running those eveals on every CI is a terrible waste of tokens - especially if tokens will be API ones.

For now evals are "eventually consistent" - when we break it, the idea is that they will be fixed by the next agentic PR that touches this area.

PR #896, opened 2026-07-16
docs(security): fix role-guide links and skills path in onboarding

### LINKED ISSUE #870: docs/security/new-members-onboarding.md: stale .claude/skills/ reference + wrong link target
### What should happen

1. Replace the `.claude/skills/` display path with the canonical `.agents/skills/` path.
2. Fix the cross-reference that sends new members to the README for role info — it should point to `docs/security/roles.md`.

### Why

The framework's canonical skill path is `.agents/skills/` (with relay symlinks from `.claude/skills/`). The doc uses the old path. Separately, new members looking for role descriptions are directed to the README, but that info lives in `roles.md`.

### Which layer

Documentation (`docs/`, `README.md`, `CONTRIBUTING.md`)

FenjuFu:
## Summary

- Point the three role-info cross-references in `docs/security/new-members-onboarding.md` at `roles.md` instead of the top-level `README.md` — the README has no role sections; the role guides and their step ranges (triager 1–6, remediation developer 7–11, release manager 12–15) live in `docs/security/roles.md`.
- Replace the retired `.claude/skills/` display path with the canonical `.agents/skills/` in the "Using the agent skills" section (link target unchanged).

Fixes #870

## Test plan

- Verified the root README's section list contains no role sections, while `docs/security/roles.md` is organised exactly by the three roles with their step ranges.
- Verified `roles.md` sits in the same directory as the onboarding doc, so the relative links resolve; `.agents/skills/` exists at the repo root. Link targets are unchanged or same-directory, so lychee should stay green.

### DIFF
--- a/docs/security/new-members-onboarding.md
+++ b/docs/security/new-members-onboarding.md
@@ -26,9 +26,9 @@ is the soft-landing guide — it tells you how the team works, where
 the action happens, and what is expected of you in the first few
 weeks.
 
-Read this end-to-end once, then use [`README.md`](../../README.md) as the
+Read this end-to-end once, then use [`roles.md`](roles.md) as the
 operational reference when you start actually handling issues. The
-README is organised by role (triager / remediation developer / release
+role guide is organised by role (triager / remediation developer / release
 manager); pick the role that matches what you are about to do and jump
 into its section.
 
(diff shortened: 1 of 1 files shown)


justinmclean (approved):
Thanks for fixing this.

PR #902, opened 2026-07-18
feat(codex): first-class Codex CLI skill runtime

### LINKED ISSUE #313: feat(adapter/codex): first-class Codex CLI skill runtime
**Goal:** Bring OpenAI's Codex CLI to feature parity with Claude Code as a **first-class skill runtime** for the framework — not just a delegation target. Grounds: [RFC-AI-0004 Principle 3 — Vendor neutrality across LLM backends](../tree/main/docs/rfcs/RFC-AI-0004.md), which names Codex by name as one of the runtimes a framework skill must be expressible against.

**Current state:** A Claude Code plugin (`codex:*`) lets Claude *delegate* to Codex for specific subtasks (rescue, adversarial review). That is **not** a native Codex skill runtime — a maintainer who has only Codex installed can't use the framework at all today.

**What "parity" means:**

- Skills under `.claude/skills/<name>/SKILL.md` are invokable from Codex CLI directly (Codex reads framework skills as native agents)
- The `tools/*` bridges (gh, jira, gmail, ponymail, vulnogram, cve-org, …) are reachable — these are language-agnostic CLI invocations, so they should port directly
- The `setup-isolated-*` sandbox model has a Codex-equivalent
- HITL primitives: Codex supports per-action confirmation — verify and document the wiring (RFC-AI-0004 Principle 1)

**Suggested approach:**

- Map `SKILL.md` → Codex's `AGENTS.md` convention (Codex reads `AGENTS.md` natively)
- Document skill-invocation: how a maintainer runs (e.g.) `security-issue-triage` end-to-end via Codex
- Map sandbox parity — Codex has its own sandbox model; document the mapping to the framework's `bubblewrap` + network-allowlist baseline
- The existing `codex:*` Claude-Code plugin is a reference for how skill ↔ Codex translation can work in the *other direction*; the native runtime is the inverse

**Reference:**

- RFC-AI-0004: [`docs/rfcs/RFC-AI-0004.md`](../tree/main/docs/rfcs/RFC-AI-0004.md)
- Codex CLI: https://github.com/openai/codex
- `AGENTS.md` convention: https://agents.md/
- Existing skill shape: [`.claude/skills/`](../tree/main/.claude/skills/)

dpol1:
## Summary

- Closes #313: makes OpenAI Codex a first-class Magpie skill runtime instead of a delegation target, per RFC-AI-0004 Principle 3 (vendor neutrality). A maintainer with only Codex installed can now discover and run Magpie skills end to end.
- Ships a committed, reviewed Codex posture: `.codex/config.toml` (workspace-write, network off, on-request approvals) and `.codex/rules/magpie.rules` (scoped read allows, prompted remote mutations, forbidden credential disclosure). `sandbox-lint --codex` validates both statically.
- Skips the SKILL.md-to-AGENTS.md conversion the issue suggested: Codex scans the canonical `.agents/skills` symlinks natively, so SKILL.md stays the single source. The setup and setup-isolated skills route Codex sessions to explicit Codex steps before any Claude-only branch, and new eval fixtures pin that routing.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `prek run --all-files` passes (all hooks green, including `check-placeholders`, `symlink-lint`, workspace ruff/mypy/pytest, `skill-and-tool-validate`, `spec-validate`)
- [x] For Python packages touched: `uv run pytest` (103 passed in `tools/sandbox-lint`, including the new `tests/test_codex.py` invariant cases for tampered configs and rules) / `ruff check` / `mypy` all clean
- [ ] For Groovy bridges touched: not touched
      (`PYTHONPATH=tools/skill-evals/src python3 -m skill_evals.runner tools/skill-evals/e
      (runtime-routing suites for all four `setup-isolated-setup-*` skills, each with a `cg the Codex branch is taken and no `.claude` file is treated as a failure)
- [ ] Other: end-to-end smoke on a real Codex session (`/skills` discovery, `$magpie-security-issue-triage` invocation, approval prompt on a `gh` mutation)

## RFC-AI-0004 compliance

      (`approval_policy = "on-request"`; exec-policy rules prompt on `gh pr`/`gh issue` mu
      (`workspace-write` with `network_access = false`; networked bridges cross the native
- [ ] **Conversational + correctable** — agentic-override path documented if behaviour is
- [ ] **Privacy LLM** — private content does not reach a non-approved LLM; redactor invoke

## Linked issues

Closes #313.

- The issue proposed mapping SKILL.md onto Codex's AGENTS.md convention. This PR rejects tee would be a second source that drifts from `.claude/skills`. Codex follows the existing`.agents/skills` symlinks, so nothing is duplicated. If that call looks wrong, the runtime-contract table in `docs/adapters/codex.md` is the place to push back.
- `sandbox-lint` parses `magpie.rules` by splitting on `prefix_rule(` after dropping commeeal Starlark parser. Deliberate trade-off: the invariants only need block-level matching,and a parser dependency felt heavier than the problem. Tampering cases are covered in `tests/test_codex.py`.
- Two known gaps, both documented in the adapter's limitations section: agent-guard denialodex hook yet, and the exec-policy rule format is an evolving Codex interface, so theadapter stays experimental until a minimum Codex version is pinned.
- Project trust stays a human decision. Adoption never edits Codex trust state, and pre-exfiguration survives adopt and upgrade.

### DIFF
--- a/.codex/config.toml
+++ b/.codex/config.toml
@@ -1,0 +1,27 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Repository-scoped Codex posture. Codex loads this file only after the
+# user trusts the project, keeping the trust decision outside repository
+# control.
+sandbox_mode = "workspace-write"
+approval_policy = "on-request"
+
+[sandbox_workspace_write]
+# Networked Magpie bridges cross the sandbox only after the native Codex
+# approval/rules layer has evaluated the action.
+network_access = false
--- a/.codex/rules/magpie.rules
+++ b/.codex/rules/magpie.rules
@@ -127,3 +127,9 @@ prefix_rule(
     decision = "prompt",
     justification = "Credential, variable, and gist operations require confirmation.",
 )
+
+prefix_rule(
+    pattern = ["gh", "api"],
+    decision = "prompt",
+    justification = "The raw API passthrough can express any read or mutation.",
+)
--- a/.codex/rules/magpie.rules
+++ b/.codex/rules/magpie.rules
@@ -1,0 +1,129 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+#
+#   http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+# Codex exec-policy is the human-in-the-loop layer for Magpie's networked
+# CLI bridges. Read-only GitHub operations may proceed; mutations prompt.
+# Commands not listed here still follow approval_policy = "on-request" and
+# the workspace-write sandbox, whose project profile disables network.
+
+prefix_rule(
+    pattern = ["git", "fetch"],
+    decision = "allow",
+    justification = "Refreshing remote refs is read-only.",
+)
+
+prefix_rule(
+    pattern = ["git", ["push", "pull"]],
+    decision = "prompt",
+    justification = "Publishing or integrating remote changes requires confirmation.",
+)
+
+prefix_rule(
+    pattern = ["gh", "auth", ["token", "refresh"]],
+    decision = "forbidden",
+    justification = "Agents must not expose or rotate the user's GitHub credential.",
+)
+
+prefix_rule(
+    pattern = ["gh", "auth", "status"],
+    decision = "allow",
+    justification = "Authentication status does not expose the credential.",
+)
+
+prefix_rule(
+    pattern = ["gh", "pr", ["view", "list", "diff", "checks", "status"]],
+    decision = "allow",
+    justification = "These pull-request operations are read-only.",
+)
+
+prefix_rule(
+    pattern = ["gh", "issue", ["view", "list", "status"]],
+    decision = "allow",
+    justification = "These issue operations are read-only.",
+)
+
+prefix_rule(
+    pattern = ["gh", "repo", ["view", "list"]],
+    decision = "allow",
+    justification = "These repository operations are read-only.",
+)
+
+prefix_rule(
+    pattern = ["gh", "run", ["view", "list", "watch"]],
+    decision = "allow",
+    justification = "These workflow-run operations do not mutate GitHub state.",
+)
+
+prefix_rule(
+    pattern = ["gh", "workflow", ["view", "list"]],
+    decision = "allow",
+    justification = "These workflow operations are read-only.",
+)
+
+prefix_rule(
+    pattern = ["gh", "release", ["view", "list"]],
+    decision = "allow",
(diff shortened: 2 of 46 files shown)


potiuk:
NICE!

justinmclean:
Solid first-class Codex adapter, and the network-disabled `workspace-write` sandbox as the hard boundary with the exec-policy layer adding per-action prompts is the right shape. License headers are all present, the parser edge-case tests are good, and the known-limitations section is refreshingly honest. A few observations inline, none blocking.

### Smaller observations

- **`.codex/rules/magpie.rules` — `gh api` is ungoverned.** The generic passthrough `gh api` (and `gh api graphql`) can perform *any* mutation: create/merge PRs, delete issues, even read the token via `gh api user`. It matches none of the `prefix_rule` patterns, so it falls through to `approval_policy = "on-request"`. The `network_access = false` sandbox does backstop it (a networked `gh api` call escalates to approval, so it's not a silent bypass), but it skips the descriptive per-action `prompt` that every other `gh` mutation gets. Worth deciding explicitly: add a `prefix_rule(pattern = ["gh", "api"], decision = "prompt")` (or `forbidden`) so the operator sees intent, rather than leaning on the sandbox alone.

- **`sandbox_lint/codex.py` — invariant checks substring-match the whole rule block, including the justification string.** `check_codex_invariants` tests things like `'"download"' in block` against the full `prefix_rule(...)` text. A future allow rule whose *justification* happens to contain "download", "push", or "token" would flip a check (false positive on the release-download guard; false negative if a justification string satisfies a "must prompt" check). Matching only the `pattern = [...]` list rather than the entire block would make the linter robust to justification wording.

- **`sandbox_lint/codex.py` — the linter asserts required rules exist but never rejects an over-broad `allow`.** The checks confirm *some* prompt/forbidden block covers gh-pr/gh-issue/git-push/auth-token, but nothing fails a profile that adds, say, `prefix_rule(pattern = ["gh", "pr", "merge"], decision = "allow")`. Codex most-specific-matching mitigates the known-mutation case, and network sandboxing limits blast radius for `gh`, but for a security linter a positive "no mutation verb appears in any allow block" check would close the gap. Reasonable to defer if "invariants only, first cut" is the intent — flagging so it's a conscious choice.

- **`nit` — `gh auth refresh` is `forbidden` in the rules but only `gh auth token` is asserted by the linter and exercised by `test_codex.py`.** Both credential ops matter; a one-line assertion + test case would lock the `refresh` half in too.

dpol1:
All four addressed in a61458c2, 9db4b016, 29a3476a.

- **`gh api`**: now governed by an explicit `prompt` rule, plus a linter invariant so the rule cannot be dropped silently later. Went with `prompt` rather than `forbidden` because several skills rely on `gh api` for reads that have no porcelain equivalent (review-thread replies, body-field edits). The prefix rule also covers `gh api graphql`. One small correction on the token angle: `gh api user` returns the profile, not the credential; the token read is `gh auth token`, which stays forbidden.
- **Substring matching**: the invariants now parse each block into its `pattern` tokens and `decision` and match against the pattern tokens only. Justification wording can no longer satisfy or trip a check; regression tests cover both directions.
- **Over-broad allows**: new invariant rejects any `allow` rule whose final pattern element contains a mutation verb. Final element rather than anywhere in the pattern: `run` is a noun in `gh run view` but a mutation as the final token of `gh workflow run`. Known limit: it does not flag an over-broad prefix such as a bare `["gh"]` allow; can add a minimum-depth rule if you think that gap is worth closing now.
- **`gh auth refresh`**: asserted by the linter and exercised in the tests alongside `token`.

choo121600 (approved):
Cool,
Looks good

choo121600 (commented):
Did a fresh independent review pass over the diff — no new findings, nothing blocking. This is solid work.

The `sandbox-lint` `--codex` invariants and the `_parse_rules` parser in `tools/sandbox-lint/src/sandbox_lint/codex.py` hold up under an adversarial read: comment stripping, `prefix_rule(`-boundary block splitting (so a `)` in a justification can't truncate a block), pattern-token-only matching (so a `"push"`/`"download"` inside a justification can neither satisfy nor trip an invariant), and the final-element mutation-verb check (so `gh run view`'s inner `run` noun passes while a trailing `merge`/`run` is rejected) are each backed by a targeted test, and the committed `.codex` profile satisfies every invariant.

The surrounding contract is complete too: ASF license headers on every new file, the hand-maintained cross-references refreshed in lockstep (`docs/adapters/registry.md`, `docs/prerequisites.md`, `docs/vendor-neutrality.md` prose + the *Status at a glance* table, `tools/sandbox-lint/README.md`), the `tools/spec-loop/specs/codex-runtime.md` spec, and runtime-routing eval fixtures for the `setup-isolated-*` skills. CI is fully green.

Observations only, none blocking.

---

> confirmed by a Magpie maintainer. The findings
> below are observations, not blockers; a Magpie
> maintainer — a real person — will take the next look at the
> PR. If you think a finding is mis-applied, please reply on
> the PR and a maintainer will weigh in.*
>
> [AGENTS.md § Reviewing pull requests](https://github.com/apache/magpie/blob/main/AGENTS.md#reviewing-pull-requests).

potiuk (approved):

PR #908, opened 2026-07-19
docs(rfc): RFC-AI-0007 — selective (family-scoped) snapshot fetch

potiuk:
**RFC-AI-0007** — proposes making the non-marketplace (`/magpie-setup`) adoption fetch **only the chosen skill families** instead of the whole framework.

## The problem
The `.apache-magpie/` snapshot today pulls the whole repo — all 10 families, every tool, all of `docs/`, the `projects/` templates — even when an adopter opts into one or two families. Investigation showed skills reference outside `skills/` heavily (docs ×319, AGENTS.md ×202, tools ×186, projects ×116), so a naive "chosen skills only" fetch would dangle real runtime dependencies.

## The proposal (Phase A — design review)
- **git sparse-checkout + partial clone** fetches: chosen families' skills + always-on (`setup`, `utilities`) + the **runtime closure** they execute against (`AGENTS.md`, `PRINCIPLES.md`, `projects/_template/`, invoked `tools/`).
- **Runtime deps vs doc-links:** runtime deps are fetched; the ~319 `docs/…` prose links are rewritten to canonical `github.com/apache/magpie/blob/<ref>/…` URLs (deterministic at fetch time). A new `runtime_deps:` frontmatter, enforced by `skill-and-tool-validator`, keeps the closure from drifting.
- **Per-family** granularity (matches the existing opt-in prompt); per-skill rejected as fragile.
- **Lock** gains a `families:` list so `upgrade`/`verify` reproduce the sparse set.
- **`svn-zip`** stays a whole-release fetch (a signed zip can't be sparse); sparse applies to the git methods.
- No change to the symlink/override model or **PRINCIPLES §13**.

## Process
This is **Phase A** — the RFC is the review checkpoint. **Phase B** (wiring `setup` + `install-recipes.md` + the lock field + validator) follows on approval, per how RFC-AI-0006 (skill-sources) was staged.

Decisions baked in per request: runtime-only closure, per-family granularity, RFC-first.

### DIFF
--- a/docs/rfcs/RFC-AI-0007.md
+++ b/docs/rfcs/RFC-AI-0007.md
@@ -1,0 +1,195 @@
+<!-- SPDX-License-Identifier: Apache-2.0
+     https://www.apache.org/licenses/LICENSE-2.0 -->
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [RFC-AI-0007: Selective snapshot — fetch only the chosen skill families](#rfc-ai-0007-selective-snapshot--fetch-only-the-chosen-skill-families)
+  - [Abstract](#abstract)
+  - [Status of this document](#status-of-this-document)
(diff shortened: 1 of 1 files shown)


potiuk:
Looking for reviews here as well :)

choo121600 (approved):
I think looks good

PR #914, opened 2026-07-20
fix(pr-management-triage): stop misclassifying PRs on truncated and unsettled GitHub data

potiuk:
Three defects found by running a full 333-PR triage sweep against a large adopting project. Each one produced wrong contributor-facing outcomes, and none is visible from reading the docs alone — they only appear at scale.

### 1. `statusCheckRollup.contexts` truncates silently

It's a paginated connection, and the page it returns is a prefix with no indication that anything is missing. On a repo running 100+ check-runs per PR, the derived failed-check list — which rows 10–13 and the Real-CI guard all read — was incomplete. Observed:

| Page showed | Actually failing | Result |
|---|---|---|
| 1 (a static check) | 16, incl. a whole provider-test sweep | `comment` instead of `draft` |
| 2 | 4 — same suite on **all four** DB backends | `rerun` — a consistent failure treated as a flake |
| 1 | 3 | `comment` instead of `draft` |

Raising the page size narrows the window but can't close it, so the fix makes REST re-derivation mandatory before any row reads `failed_checks`. `rollup.state` was never affected — it's computed server-side over all contexts. Only the derived *list* was unsafe. Cost is ~50 extra REST calls per 333-PR sweep, and only for PRs that are already red.

### 2. `mergeable: UNKNOWN` read as "no conflict"

GitHub computes mergeability lazily. Rows 19/20 were written `!= CONFLICTING`, which is **true** for `UNKNOWN`, so unsettled PRs classified as `passing`. Fix requires `== MERGEABLE` explicitly and adds a `mergeable_state` guard to `mark-ready`, folded into the PR fetch that recipe already makes (no extra call).

On the observed sweep, **11 of 39** mark-ready candidates reported `UNKNOWN` at fetch and `dirty` at mutation — all genuinely conflicting. Without the guard they'd have entered the review queue unmergeable.

### 3. Stale-sweep 1a counted comments, not pushes

Its action is `close` — the least reversible thing the skill does — yet it tested only for author *comments*. Contributors who answer feedback with code read as silent. One PR with author pushes 5, 12, and 25 days after the triage comment matched the trigger and would have been closed while actively worked.

The `last_author_activity_at` input already folds in pushes and review-thread replies; 1a just wasn't using it. (Sweep 1b keys on `updated_at` and was already correct.)

### Not included

The `<= 2 failures -> rerun` rule also misfires on deterministic failures, but rows 12/12b already handle that — no change needed.

All additions use `<upstream>` placeholders; `check-placeholders`, `markdownlint`, `typos`, and `lychee` pass.

---

- [X] Yes — Claude Opus 4.8 (1M context)

Generated-by: Claude Opus 4.8 (1M context)

### DIFF
--- a/skills/pr-management-triage/actions.md
+++ b/skills/pr-management-triage/actions.md
@@ -362,7 +362,10 @@ executed.
 # with `conclusion: "action_required"`. The query parameter
 # `?status=action_required` matches no runs and would silently
 # return an empty result — post-filter on `conclusion` instead.
-head_sha=$(gh api "repos/<owner>/<repo>/pulls/<N>" --jq '.head.sha')
+# One fetch covers both guards.
+read -r head_sha merge_state <<<"$(gh api "repos/<owner>/<repo>/pulls/<N>" \
+  --jq '"\(.head.sha) \(.mergeable_state)"')"
+
 pending=$(gh api "repos/<owner>/<repo>/actions/runs?head_sha=${head_sha}&per_page=20" \
   --jq '[.workflow_runs[] | select(.conclusion == "action_required")] | length')
 if [ "$pending" -gt 0 ]; then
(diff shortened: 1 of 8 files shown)


potiuk:
Since we now have "must approve" - looking for reviews here :)

choo121600 (approved):
looks good to me :)
nice

PR #959, opened 2026-07-27
fix(pr-management-stats): remove unused local variable in test_ready_split.py

### LINKED ISSUE #957: Remove the unused local variable in test_ready_split.py
## Summary

`tools/pr-management-stats/tests/test_ready_split.py` assigns a variable that is
never used, which usually means an assertion is missing.

## Background

CodeQL flagged an unused local. In a test this is worth a second look rather
than a straight deletion: `out = dashboard.render_ready_split(...)` computing a
value nothing asserts on suggests the test calls the function but never checks
what it returned, so it only proves the call does not raise.

## Where to look

- `tools/pr-management-stats/tests/test_ready_split.py` around line 151, in
  `test_render_ready_split_empty_state`.
- Neighbouring tests in the same file show what a rendered result is normally
  asserted against.

## Acceptance criteria

- [ ] Either the value is asserted on, or the assignment is removed because the
      call itself is the assertion.
- [ ] If an assertion is added, it fails when `render_ready_split` returns the
      wrong thing (check by temporarily breaking it).
- [ ] `uv run --directory tools/pr-management-stats --group dev pytest` passes.

## Estimated effort

## Getting started

New to Magpie? Start with

HeaTTap:
Fixes #957

Removed the unused local variable `ctx` in `test_render_ready_split_empty_state` inside `tools/pr-management-stats/tests/test_ready_split.py` as `render_ready_split` takes the dict payload directly without needing context.

### DIFF
--- a/tools/pr-management-stats/tests/test_ready_split.py
+++ b/tools/pr-management-stats/tests/test_ready_split.py
@@ -148,7 +148,6 @@ def test_render_ready_split_oldest_on_left_and_cards():
 
 
 def test_render_ready_split_empty_state():
-    ctx = make_ctx()
     out = dashboard.render_ready_split(
         {"counts": dict.fromkeys(
             ["never-reviewed", "discussed-no-decision", "changes-requested", "approved"], 0),


justinmclean (approved):
Thanks for the fix.

Looking deeper I found another issue that you might also want to fix.

This defect existed because nothing in the repo could have caught it.

tools/dev/run-workspace-check.sh selects members for ruff by auto-discovery: ruff → runs only on members with [tool.ruff]. tools/pr-management-stats/pyproject.toml has no [tool.ruff] section, so workspace-ruff-check and workspace-ruff-format skip the member entirely. An unused local is ruff's F841, which is in the default rule set, so a linted member would have failed on this line the day it landed.

Three of the thirty workspace members are in that position:
tools/agent-isolation
tools/jira
tools/pr-management-stats

All three pin ruff in their dev dependency group, so the intent to lint them is clearly there. None uses the documented [tool.magpie.checks] skip opt-out, which is what a deliberate exemption would look like. This is silent drift of exactly the kind the workspace-membership guard was built to prevent, one layer down: the member is in [tool.uv.workspace] members, so check-workspace-members is satisfied, but the per-check auto-discovery still drops it.

Suggested follow-up: add [tool.ruff] to the three members and fix whatever that surfaces.

PR #962, opened 2026-07-27
fix(skill-evals): remove Tier-B exclusivity qualifier from stage-2 fixture

### LINKED ISSUE #950: Remove the Tier-B exclusivity qualifier from the quick-merge stage-2 fixture
## Summary

A `pr-management-quick-merge` fixture adds a "Tier B glob only" qualifier that
does not exist in `candidate-rules.md`.

## Background

The fixture's Tier-B definition is narrower than the rule it reflects. No case
hits the difference today because the current glob sets are disjoint, so the
suite passes. A future config change with overlapping globs would make fixture
and rule disagree, and the fixture would win, because it is what the eval grades
against. Fixtures should restate the rule, not tighten it.

## Where to look

- `evals/pr-management-quick-merge/stage-2-triviality/fixtures/system-prompt.md` -
  the Tier-B definition carrying the qualifier.
- `skills/pr-management-quick-merge/candidate-rules.md` - the definition to match.

## Acceptance criteria

- [ ] The fixture's Tier-B wording matches `candidate-rules.md`.
- [ ] No case's expected output changes, since the sets are disjoint today. If
      one does, that is a real finding worth calling out in the PR.
- [ ] The `pr-management-quick-merge` eval suite passes.

## Estimated effort

HeaTTap:
Fixes #950

Removed the redundant "only" qualifier from the Tier B definition in `tools/skill-evals/evals/pr-management-quick-merge/stage-2-triviality/fixtures/system-prompt.md` to match `candidate-rules.md`.

### DIFF
--- a/tools/skill-evals/evals/pr-management-quick-merge/stage-2-triviality/fixtures/system-prompt.md
+++ b/tools/skill-evals/evals/pr-management-quick-merge/stage-2-triviality/fixtures/system-prompt.md
@@ -53,7 +53,7 @@ Assign the tier **only** when `disposition == "candidate"`:
 
 - **Tier A** — every changed file matches a Tier A allow glob.
 - **Tier B** — every changed file matches a Tier A **or** Tier B glob,
-  and at least one file matches a Tier B glob only.
+  and at least one file matches a Tier B glob.
 
 ## Output format
 


justinmclean (approved):
Correct, and it matches candidate-rules.md:140 word for word now. I checked all eight stage-2 cases and none of them can tell the two wordings apart, so no expected output moves, exactly as the issue predicted. Good call leaving the second only at line 80 alone, that one mirrors candidate-rules.md:144 and is not the same defect.

Note that it was not clear to me if you ran the eval tests or not; please run the tests to confirm that any changes do not introduce errors before you raise a PR.

PR #963, opened 2026-07-27
fix(skill-evals): extend default prose_fields with grading-schema.json

### LINKED ISSUE #946: Make grading-schema prose_fields extend the default set instead of replacing it
## Summary

A per-step `grading-schema.json` replaces the runner's default prose-field set
rather than adding to it, so a prose field added later silently reverts to
exact-match grading. One schema is also now dead config.

## Background

Prose fields are graded by a model; everything else is compared exactly. When a
step ships a `grading-schema.json` naming one prose field, the default set stops
applying to that step. Add a `notes` or `summary` field to that step's
`expected.json` later and it is exact-matched, failing on any wording
difference, with nothing to indicate why.

## Where to look

- `evals/release-verify-rc/*/fixtures/grading-schema.json` - both set
  `prose_fields: ["paste_recipe"]`.
- `evals/release-audit-report/step-2-assemble-record/fixtures/grading-schema.json` -
  sets `prose_fields: ["record_markdown"]`, but that is no longer a compared
  `expected.json` key, so the schema acts on nothing.
- `tools/skill-evals/src/skill_evals/runner.py` - where the default set and the
  per-step override are resolved.

## Acceptance criteria

- [ ] Document whether a per-step schema replaces or extends the default, in the
      skill-evals README.
- [ ] The behaviour matches the documented decision, with a test.
- [ ] The dead `release-audit-report` schema is removed or given a real field.
- [ ] `uv run --directory tools/skill-evals --group dev pytest` passes.

## Estimated effort

HeaTTap:
Fixes #946

- Updated `load_grading_schema` in `tools/skill-evals/src/skill_evals/runner.py` to extend `DEFAULT_PROSE_FIELDS` rather than replacing it.
- Updated unit tests in `test_runner.py` and documentation in `README.md`.
- Removed dead `grading-schema.json` from `evals/release-audit-report/step-2-assemble-record/fixtures/`.

### DIFF
--- a/tools/skill-evals/README.md
+++ b/tools/skill-evals/README.md
@@ -247,7 +247,7 @@ The default prose-field set is `rationale`, `reason`, `reasons`,
 
 ```json
 {
-  "prose_fields": ["custom_prose_field"]
+  "prose_fields": ["paste_recipe"]
 }
 ```
 
--- a/tools/skill-evals/README.md
+++ b/tools/skill-evals/README.md
@@ -242,18 +242,17 @@ writes stdout works). Pass `--exact` to disable grading entirely.
 
 The default prose-field set is `rationale`, `reason`, `reasons`,
 `drop_reason`, `blockers`, `notes`, `summary`, `explanation`,
-`details`, `description`. Override it per fixtures dir by placing a
+`details`, `description`. Extend it per fixtures dir by placing a
 `grading-schema.json` next to `step-config.json`:
 
 ```json
 {
-  "prose_fields": ["rationale", "drop_reason"]
+  "prose_fields": ["custom_prose_field"]
 }
 ```
 
-An empty list (`"prose_fields": []`) makes every field decision-graded
-even with the grader on, equivalent to passing `--exact` for that
-fixtures dir. The grader is called fresh on every run; nothing is
+The fields listed in `grading-schema.json` are added to the default
+prose-field set. The grader is called fresh on every run; nothing is
 cached.
 
 **Self-eval caveat.** When the model invoked by `--cli` is the same
--- a/tools/skill-evals/evals/good-first-issue-sweep/step-3-present-proposals/fixtures/grading-schema.json
+++ b/tools/skill-evals/evals/good-first-issue-sweep/step-3-present-proposals/fixtures/grading-schema.json
@@ -1,3 +1,0 @@
-{
-  "prose_fields": []
-}
--- a/tools/skill-evals/evals/pr-stale-sweep/step-4-compose-comment/fixtures/grading-schema.json
+++ b/tools/skill-evals/evals/pr-stale-sweep/step-4-compose-comment/fixtures/grading-schema.json
@@ -1,3 +1,0 @@
-{
-  "prose_fields": []
-}
--- a/tools/skill-evals/evals/pr-stale-sweep/step-5-confirm/fixtures/grading-schema.json
+++ b/tools/skill-evals/evals/pr-stale-sweep/step-5-confirm/fixtures/grading-schema.json
@@ -1,3 +1,0 @@
-{
-  "prose_fields": []
-}
--- a/tools/skill-evals/evals/pr-stale-sweep/step-7-recap/fixtures/grading-schema.json
+++ b/tools/skill-evals/evals/pr-stale-sweep/step-7-recap/fixtures/grading-schema.json
@@ -1,3 +1,0 @@
-{
-  "prose_fields": []
-}
--- a/tools/skill-evals/evals/release-audit-report/step-2-assemble-record/fixtures/grading-schema.json
+++ b/tools/skill-evals/evals/release-audit-report/step-2-assemble-record/fixtures/grading-schema.json
@@ -1,3 +1,0 @@
-{
-  "prose_fields": ["record_markdown"]
-}
--- a/tools/skill-evals/src/skill_evals/runner.py
+++ b/tools/skill-evals/src/skill_evals/runner.py
@@ -48,7 +48,7 @@ Two modes:
    exactly. Override the grader command with ``--grader-cli``, or pass
    ``--exact`` to disable grading and require verbatim equality on
    every field. The set of prose fields defaults to a built-in list
-   and can be overridden per fixtures dir via ``grading-schema.json``.
+   and can be extended per fixtures dir via ``grading-schema.json``.
    No caching: every prose field is sent to the grader on every run.
 
 Usage:
@@ -349,7 +349,8 @@ DEFAULT_GRADER_CLI: str = "claude -p --model haiku"
 # Keys whose values are treated as prose by default. The runner sends these
 # to the grader CLI for a soft "does the candidate support the same
 # conclusion?" judgement instead of requiring verbatim string equality.
-# A per-fixtures-dir ``grading-schema.json`` can replace this list.
+# A per-fixtures-dir ``grading-schema.json`` can extend this list; it cannot
+# narrow it. Use ``--exact`` to disable grading for a whole run.
 DEFAULT_PROSE_FIELDS: frozenset[str] = frozenset(
     {
         "rationale",
--- a/tools/skill-evals/src/skill_evals/runner.py
+++ b/tools/skill-evals/src/skill_evals/runner.py
@@ -397,9 +397,7 @@ def load_grading_schema(fixtures_dir: Path) -> set[str]:
     """Return the set of prose field names for cases in this fixtures dir.
 
     Reads ``fixtures_dir/grading-schema.json`` when present. The file may
-    set ``prose_fields`` to a string list that *replaces* the default set
-    (use ``["rationale", "reason", ...]`` to be explicit, or ``[]`` to
-    grade everything by exact match).
+    set ``prose_fields`` to a string list extending the default set.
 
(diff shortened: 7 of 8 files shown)


justinmclean (changes requested):
## What it does

A `grading-schema.json` lists field names that get judged loosely by a model rather than having to match the expected value exactly. Before this PR that file replaced the built-in list. Now it adds to it.

## Why it matters

Under the old behaviour, a step with its own schema quietly lost the built-ins. Add a `reason` field to that step later and it would suddenly demand an exact string match, failing for no visible cause. That's the bug in #946.

## What I checked

Adding fields to the loosely-judged list makes checks weaker, so the question is whether any existing eval gets weaker. I went through all 86 schema files: for each one, worked out which built-in names it currently leaves out, then looked at every expected answer in that directory to see whether any of those names actually appears.

None do. Not one suite changes. The verbose schemas that already list every built-in become no-ops, and the short ones never had a built-in name in their expected output. So the change is inert against the current tree and only affects fields added later, which is the intent.

The deleted `release-audit-report` schema checks out too. It named `record_markdown`, and that key appears in the output spec but in none of the four expected answers, which all go through a `has_record_markdown_non_empty` check instead. Genuinely defunct, so deleting rather than repurposing is right.

The docstring was updated in the same hunk as the code, and the README no longer claims the old behaviour.

The README example had to change and did. It used to show `{"prose_fields": ["rationale", "drop_reason"]}`, which made sense when the file replaced the defaults: it narrowed the set to those two. Under the new behaviour both names are already built in, so that example is a file that does nothing. Leaving it would have taught people to write no-ops. Swapping it for a custom field name is the right call.

One thing to fix though: the replacement is an invented `custom_prose_field`, and the test in the same PR invents `custom_prose` instead. Two made-up names for the same idea, in one change. Neither reads like a field, they read like a description of the category. Every real custom field in the tree reads like data: `paste_recipe`, `record_markdown`, `skip_reason`, `handoff_note`, `evidence_summary`. Use `paste_recipe` in both places. `release-verify-rc` has three live instances, so a reader can grep it.

## Three things to fix

**A comment five lines above the fixed docstring still describes the old behaviour** (`runner.py:352`):

```python
# A per-fixtures-dir ``grading-schema.json`` can replace this list.
```

That's now backwards, and it's the first thing a reader sees when they scroll to the constant.

**Four files now say nothing.** These contain `{"prose_fields": []}`:

- `good-first-issue-sweep/step-3-present-proposals`
- `pr-stale-sweep/step-4-compose-comment`
- `pr-stale-sweep/step-5-confirm`
- `pr-stale-sweep/step-7-recap`

An empty list used to mean "be strict about everything in this step". Empty plus the built-ins is just the built-ins, so now it means nothing at all. Harmless today, for the same reason as above, but a contributor opening one will reasonably conclude the step is deliberately strict and act on it. I'd delete all four here, same as the defunct one. Worth noting this also removes the only per-step way to force exact matching, since `--exact` applies to the whole run. If anyone actually wants that, it needs its own setting.

**The replacement test is weaker than the one it replaces** (`test_runner.py`, `test_cli_grader_mode_respects_grading_schema_override`).

The old test set `prose_fields` to `[]` and checked that a differing `reason` caused a FAIL. That proved the schema really controlled grading, and it proved it in the direction that can catch a bug.

The new test invents a field, lists it in the schema, and checks the result is PASS with a stub grader that always says yes. It does prove something, since an exact compare of two different strings would fail. But it can't tell "the field was judged by the grader" apart from "the field was skipped entirely". If a future change silently dropped that field from comparison, this test would still pass.

The file already has a stub grader that always says no. Adding a second assertion with it, expecting a FAIL, closes the hole: a skipped field would pass, a judged field fails.

## Worth a follow-up issue, not this PR

The same reasoning that condemns the old README example applies to most of the tree. Of the 85 schema files with a list, 53 contain built-in names that now do nothing, 467 redundant entries between them. Two are complete no-ops:

- `pre-first-pr-check/step-2-check-categories` → `["details"]`
- `security-issue-import-from-md/step-2-duplicate-guard` → the full built-in list, verbatim

Those two plus the four empty ones are six files that now say nothing at all. Stripping the built-in names everywhere would leave each file stating exactly what it adds, which is the whole point of the new behaviour. But probably too large to fold in here.


This review was drafted by an AI-assisted tool and confirmed by a Magpie maintainer. The maintainer approving this PR has read the findings and signed off. If something feels off, please reply on the PR and a maintainer will follow up.

*More on how Magpie handles maintainer review:* [[CONTRIBUTING.md](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md)](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md)

potiuk (approved):
Approving with @justinmclean's four review items applied directly — I pushed them as `ff7e1dc` so this can land without another round-trip. His review is the substantive one here; please read the commit rather than take it on trust.

I ran the safety analysis independently before reading his review and landed on the same answer, which is worth recording since it's the load-bearing claim: **across all 84 schema files, not one fixtures dir changes grading behaviour.** For each schema I took the built-in names it omits, then walked every `expected.json` in that directory — including nested keys, since `collect_diffs` recurses and matches `key in prose_fields` at every level — looking for any of those names. Zero hits. The verbose schemas that already re-list every built-in become no-ops; the short ones never had a built-in name in their expected output. The change is inert against the current tree and only affects fields added later, which is the intent.

The deleted `release-audit-report` schema checks out the same way: it named `record_markdown`, which appears in none of the four expected answers — they go through `has_record_markdown_non_empty` instead. Defunct, so deleting rather than repurposing is right.

And the payoff is real: **45 of the 84 schema files re-list all ten default field names verbatim**, purely to work around replace-semantics.

### What I changed in `ff7e1dc`

1. **One name instead of two invented ones.** The README said `custom_prose_field`, the test said `custom_prose`. Both now use `paste_recipe` — a real field with three live instances under `release-verify-rc`, so a reader can grep it.

2. **The stale comment above `DEFAULT_PROSE_FIELDS`** (`runner.py:352`) still said *"can replace this list"*. It now says a schema can extend but not narrow, and points at `--exact` for whole-run strictness. I also fixed the same stale wording in the **module docstring** (`runner.py:51`, *"can be overridden per fixtures dir"*), which the review didn't catch but is the same class of error and the first thing a reader of the file meets.

3. **The four `{"prose_fields": []}` schemas are deleted** — `good-first-issue-sweep/step-3-present-proposals`, `pr-stale-sweep/step-4-compose-comment`, `-5-confirm`, `-7-recap`. An empty list used to mean "be strict about everything here" and now means nothing, so leaving them would teach a contributor the opposite of the truth. As the review notes, this does remove the only per-step way to force exact matching (`--exact` is run-wide); nothing in the tree wants that today, and if it is ever needed it deserves its own setting rather than a resurrected empty list.

4. **The test now proves the field is judged, not merely tolerated.** This was the sharpest point in the review and it was correct: asserting `PASS` against an always-yes grader cannot distinguish *"the field was sent to the grader"* from *"the field was dropped from the comparison entirely"* — both look like a pass. The test now runs the same case against the always-no stub and requires a `FAIL` naming the field.

   I verified the fix does what it claims rather than assuming: patching `collect_diffs` to drop prose-keyed fields entirely makes the new assertion fail (`rc == 0`, expected `1`), where the previous version still passed.

Verified locally: 145 tests pass, and `prek` is green on the changed files — `ruff`, `ruff format`, `mypy`, `pytest`, `markdownlint`, `lychee`.

### One thing to know for the next PR here

Running any `uv`-invoking prek hook currently rewrites `uv.lock`, which makes those hooks fail with *"files were modified by this hook"* even though every check passes. The committed lock is out of sync with `pyproject.toml`'s `mypy>=2.3.0` — exactly what [#965](https://github.com/apache/magpie/pull/965) fixes. `UV_FROZEN=1` is the workaround until that lands; it is not anything wrong with this PR.

The follow-up the review flags — stripping the redundant built-in names from the ~53 schemas that still carry them — is worth its own issue and is deliberately not folded in here.

---

> will follow up.*
>

potiuk:
@justinmclean — all four of your items are addressed in `ff7e1dc`, so this needs your re-review to clear the block.

- **Invented names** — README and test now both use `paste_recipe`, per your suggestion.
- **Stale comment** — `runner.py:352` now says a schema can extend but not narrow, and points at `--exact` for whole-run strictness. I also fixed the same wording in the module docstring at `runner.py:51` ("can be overridden per fixtures dir"), which is the same error a hundred lines further up.
- **The four empty schemas** — deleted, with your point about `--exact` being run-wide recorded in the commit message.
- **The weaker test** — now runs the same case against `_GRADER_NO` and requires a FAIL naming the field. I checked it closes the hole you described rather than assuming: patching `collect_diffs` to drop prose-keyed fields entirely makes the new assertion fail, where the previous version still passed.

For what it's worth, I ran the safety analysis independently before reading your review and got the same result — zero fixtures dirs change behaviour across all 84 schemas — so that conclusion has two passes behind it now.

One thing I hit that is not this PR's fault and may be worth prioritising: the committed `uv.lock` is out of sync with `pyproject.toml`'s `mypy>=2.3.0`, so any `uv`-invoking prek hook rewrites it and then fails with *"files were modified by this hook"* even though every check passes. `UV_FROZEN=1` works around it locally. That is what [#965](https://github.com/apache/magpie/pull/965) fixes, and it currently bites anyone committing a change under `tools/`.

justinmclean (approved):
LGTM

PR #965, opened 2026-07-27
build(deps-dev): re-lock uv.lock against the mypy>=2.3.0 dev pins

justinmclean:
## Summary

- Re-locks `uv.lock` so the `[package.metadata.requires-dev]` blocks match what the manifests already declare. Eight workspace members were bumped to `mypy>=2.3.0` in their `pyproject.toml` but the lock still recorded `>=2.1.0` for them: `checker`, `generate-cve-json`, `oauth-draft`, `redactor`, `sandbox-lint`, `skill-and-tool-validator`, `skill-evals`, `vulnogram-api`.
- No dependency resolution changes and no manifest edits. `mypy` itself is already pinned at `2.3.0` in the resolved set; this only corrects the recorded specifiers, which are derived from `pyproject.toml`.
- Why it matters: the `prek` hooks invoke `uv run --project <tool>` without `--frozen` or `--locked`, so every hook run re-locks and writes this pending correction into whichever branch happens to be checked out. Contributors keep picking up eight lines of unrelated `uv.lock` churn in feature PRs, which then reads as scope creep in review. Landing the correction on its own clears it for everyone.

## Type of change

- [ ] Groovy reference impl
- [X] Other: dependancy update

## Test plan

- [ ] Other:

### DIFF
--- a/uv.lock
+++ b/uv.lock
@@ -369,7 +369,7 @@ dev = [
 
 [package.metadata.requires-dev]
 dev = [
-    { name = "mypy", specifier = ">=2.1.0" },
+    { name = "mypy", specifier = ">=2.3.0" },
     { name = "pytest", specifier = ">=9.1.1" },
     { name = "ruff", specifier = ">=0.15.21" },
 ]
@@ -495,7 +495,7 @@ dev = [
 
 [package.metadata.requires-dev]
 dev = [
-    { name = "mypy", specifier = ">=2.1.0" },
+    { name = "mypy", specifier = ">=2.3.0" },
     { name = "pytest", specifier = ">=9.1.1" },
     { name = "ruff", specifier = ">=0.15.21" },
 ]
@@ -945,7 +945,7 @@ provides-extras = ["mcp"]
 [package.metadata.requires-dev]
 dev = [
     { name = "mcp", specifier = ">=1.28.1" },
-    { name = "mypy", specifier = ">=2.1.0" },
+    { name = "mypy", specifier = ">=2.3.0" },
     { name = "pytest", specifier = ">=9.1.1" },
     { name = "ruff", specifier = ">=0.15.21" },
 ]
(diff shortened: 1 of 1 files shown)


potiuk (approved):
LGTM — and this is worth landing ahead of the rest of the queue, because it is not housekeeping: the drift it fixes currently breaks `git commit` for anyone touching `tools/`.

I hit it while working on another PR today, which made this one easy to verify against a real symptom rather than in the abstract.

### The drift is real

Eight workspace members declare `mypy>=2.3.0` in their `pyproject.toml`, but `uv.lock` still records `>=2.1.0` for them — 26 entries at `>=2.1.0` against a single `>=2.3.0`. Matches the eight named in the description exactly.

### The diff is purely metadata

18 changed lines: eight `+`/`-` pairs, every one a `specifier` under `[package.metadata.requires-dev]`. No `version =`, no `sdist`, no `wheels` lines anywhere in the diff — so the resolved set genuinely does not move, as the description claims. `mypy` was already pinned at `2.3.0` in the resolution; only the recorded specifiers were stale.

### Why it is worth prioritising

Because the lock is out of date, **every `uv`-invoking prek hook rewrites `uv.lock` as a side effect and then fails with `files were modified by this hook`** — while reporting that all its own checks passed. It looks like this:

```text
ruff check (workspace)..........................................Failed
- hook id: workspace-ruff-check
- files were modified by this hook
  ✓ ruff passed for all 27 applicable members
```

Every check green, commit rejected. It hits `vendor-neutrality-score`, `workspace-ruff-check`, and the other workspace hooks in turn, so a contributor's first encounter is a confusing failure on a file they never touched. I had to fall back to `UV_FROZEN=1` to land a commit.

Verified both directions:

| | `uv lock --check` | `uv.lock` after running the hooks |
|---|---|---|
| without this PR | fails — *"lockfile needs to be updated"* | rewritten → hooks fail |
| with this PR applied | passes | unchanged |

So this restores a clean local commit path, not just a tidy lockfile. CI is 38/38 green.

No findings.

---

> will follow up.*
>

PR #966, opened 2026-07-27
fix(tools/skill-evals): require exact boolean for negate in assertions

### LINKED ISSUE #942: Treat the negate key as a strict boolean in the skill-evals runner
## Summary

`tools/skill-evals/src/skill_evals/runner.py` reads the assertion `negate` key
with a truthiness check, so any non-empty value inverts the assertion.

## Background

`negate: true` inverts a predicate so a case can assert the *absence* of
something, which is how several suites test that injected content was not
obeyed. The current check accepts any truthy value, so `"negate": "false"`,
`"negate": 0.1` or `"negate": []` all silently do something other than what the
fixture author meant. A typo flips the assertion instead of erroring, and an
inverted security assertion passes for the wrong reason.

## Where to look

- `runner.py` line 719 - `if spec.get("negate") and holds is not None:`.
  Lines 39 and 712 document the key as `"negate": true`.
- `tools/skill-evals/tests/` - where a regression test belongs.

## Acceptance criteria

- [ ] `negate` is honoured only when it is the boolean `true`.
- [ ] A non-boolean value is rejected with a clear error rather than silently inverting.
- [ ] A test covers boolean true, boolean false, and one non-boolean value.
- [ ] `uv run --directory tools/skill-evals --group dev pytest` passes.

## Estimated effort

surajthedev:
Closes #942

tools/skill-evals/src/skill_evals/runner.py read the `negate` assertion key using Python truthiness, so a value like "false" (a non-empty string, therefore truthy) would silently invert a security assertion the author intended to leave un-negated.

This PR requires `negate` to be an exact boolean. Any non-boolean value now raises a clear TypeError instead of silently producing the wrong result.

Added three tests covering negate: true, negate: false, and non-boolean values (string, int, float, list, dict).

### DIFF
--- a/tools/skill-evals/src/skill_evals/runner.py
+++ b/tools/skill-evals/src/skill_evals/runner.py
@@ -676,6 +676,12 @@ def load_assertions(fixtures_dir: Path) -> dict[str, dict]:
                 f"{path}: assertion {key!r} has invalid type {atype!r}; "
                 f"valid types: {sorted(_VALID_ASSERTION_TYPES)}"
             )
+        negate = spec.get("negate")
+        if negate is not None and not isinstance(negate, bool):
+            raise ValueError(
+                f"{path}: assertion {key!r} has invalid 'negate' value {negate!r}; "
+                "'negate' must be a boolean (true or false)"
+            )
     return data
 
 
@@ -715,14 +721,18 @@ def evaluate_deterministic_assertion(spec: dict, actual: object) -> tuple[bool |
     concrete True/False result; a spec/usage error (None) is passed through
     unchanged so the typo still fails loudly.
 
-    ``"negate"`` must be an exact boolean; any other type (e.g. the string
-    ``"false"``, a number, or a list) raises ``TypeError`` immediately so the
-    author sees a clear error rather than silently getting the wrong behavior.
+    ``"negate"`` must be an exact boolean. A non-boolean value is a spec
+    authoring error: it returns ``(None, note)`` so the per-case path in
+    ``compare_structural`` reports the failure without aborting the whole run.
+    In practice ``load_assertions`` catches this at load time before any model
+    is invoked, so this path is only reached when the spec is built in memory
+    (e.g. in tests).
     """
     negate = spec.get("negate", False)
     if not isinstance(negate, bool):
-        raise TypeError(
-            f"'negate' must be a boolean (true or false), got {type(negate).__name__!r}: {negate!r}"
+        return (
+            None,
+            f"'negate' must be a boolean (true or false), got {type(negate).__name__!r}: {negate!r}",
         )
     holds, note = _evaluate_deterministic_assertion_raw(spec, actual)
     if negate and holds is not None:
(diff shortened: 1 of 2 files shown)


justinmclean:
Good catch on the bug. The bad-value loop is well chosen: `[]` and `{}` are falsy, so the old check swallowed them silently.

Two notes, however. The rejection mechanism turns a one-assertion failure into a whole-run abort, which I'd change before this lands. And `test_assert_negate_true_inverts_result` duplicates an existing test six lines above it.

### `runner.py`, the `raise TypeError` in `evaluate_deterministic_assertion`

This breaks the function's own contract. Its first docstring paragraph still says a spec error returns `None`, and the only caller, in `compare_structural`, is written against that:

```python
holds, note = evaluate_deterministic_assertion(spec, actual)
if holds is None:
    ok = False
    notes.append(f"{key}: {note}")
```

`compare_structural` is called from the per-case loop in `main`, and neither that call nor the loop around it sits in a `try`/`except` — the only ones nearby wrap prompt formatting and the CLI subprocess. So a `TypeError` unwinds the whole run. `_evaluate_deterministic_assertion_raw` handles the same class of authoring mistake, a missing `field`, by returning `(None, note)` and failing that one assertion. As written, one typo in one `assertions.json` costs the entire suite rather than one case.

Two ways to reject the bad value without that: validate in `load_assertions`, which already rejects bad shapes in this same file with `ValueError` and would fail once at load naming the file and key; or return `(None, "negate must be a boolean, got 'false'")` and let the existing path report it. I'd take the first, since it catches the typo before any model is invoked.

### `tests/test_runner.py`, `test_assert_negate_true_inverts_result`

This duplicates `test_assert_negate_inverts_regex_result` six lines above, which already covers `negate: True` inverting a predicate. Only the predicate type differs. Worth dropping, or adding a comment saying the point is cross-predicate coverage if that's the intent.

Two things still uncovered: nothing goes through `compare_structural`, which is where the raise-vs-return question above actually bites, and nothing covers `negate` being absent — the path every assertion in the tree uses, and the one this refactor touched (`spec.get("negate")` → `spec.get("negate", False)`).

This review was drafted by an AI-assisted tool and
confirmed by a Magpie maintainer. The findings
below are observations, not blockers; a Magpie
maintainer — a real person — will take the next look at the
PR. If you think a finding is mis-applied, please reply on
the PR and a maintainer will weigh in.

More on how Magpie handles maintainer review:
[[CONTRIBUTING.md](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md)](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md).

surajthedev:
Hi @justinmclean,

Thank you for your detailed review and valuable feedback.

I have addressed all the issues you pointed out:

The validation has been updated so it no longer causes the entire test run to abort.

The duplicate test has been removed/updated as appropriate.

The missing test coverage has also been added, including the compare_structural path and the default behavior when negate is absent.

Could you please take another look and let me know if everything is working as expected now?

I am extremely sorry for the problems my changes caused. I sincerely apologize for the inconvenience, and I really appreciate you taking the time to review my PR and provide helpful feedback.

Thank you!

justinmclean (approved):
Thanks for working on this.

Correct fix for a genuine silent-correctness bug, and the two-layer approach is right: loud failure at load time, contained failure at eval time. The second commit (moving validation to load and softening the eval path from TypeError to (None, note)) improved the design.
Load-time validation matches the style of the existing atype validation directly above it.
Test coverage is thorough: negate: True inverts, negate: False doesn't, absent key defaults correctly, non-boolean values ("false", "true", 0, 1, 0.1, [], {}) return (None, note), and an integration test via compare_structural confirms one bad assertion fails that case without raising. Rejecting int 0/1 via isinstance(x, bool) is the intended strictness and is tested.

PR #967, opened 2026-07-27
fix(workflow-security-audit): use full 40-char SHA in pin example

### LINKED ISSUE #952: Use a full 40-character SHA in the workflow-security-audit pinning example
## Summary

The remediation example in `skills/workflow-security-audit/SKILL.md` pins a
GitHub Action to a 39-character SHA. A git SHA-1 is 40 hex characters, so the
example in the rule that teaches SHA-pinning does not itself parse as a SHA.

## Background

The unpinned-actions rule tells a maintainer to replace a floating tag with a
full commit SHA, and shows `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af68`
as the shape to copy. That value is 39 characters. GitHub rejects a short or
malformed SHA in `uses:`, so anyone copying the example as a template and
adjusting it by hand can end up with a workflow that fails to resolve.

The rule's advice is right; only the example is wrong. It is worth fixing
precisely because this is the rule people copy from.

## Where to look

- `skills/workflow-security-audit/SKILL.md` line 238 - the example.
- `.github/workflows/` - workflows here already pin to real 40-character SHAs.
- `grep -rn '364c9af68' tools/skill-evals/` - check no fixture asserts on it.

## Acceptance criteria

- [ ] The example SHA is exactly 40 hex characters.
- [ ] It is a real commit on the action referenced, not a made-up string.
- [ ] Any eval fixture referencing the old value is updated.
- [ ] `skill-and-tool-validate` passes.

## Estimated effort

~30 minutes for someone new to the codebase.

Shaurya2k06:
## Summary

- Fix the unpinned-actions remediation example in
  `skills/workflow-security-audit/SKILL.md`: the shown commit was 39
  hex characters and does not parse as a git SHA.
- Restore the missing final digit so the example is the real
  `actions/checkout` v4.2.2 commit
  (`11bd71901bbe5b1630ceea73d27597364c9af683`).

## Type of change

- [ ] Other: `uv.lock` metadata refresh from the vendor-neutrality hook
      when the skill file was staged (mypy specifier sync)

## Test plan

- [x] Example SHA is exactly 40 hex characters
- [x] SHA resolves on `actions/checkout` (`Prepare 4.2.2 Release (#1953)`)
- [x] `rg` confirms no eval fixture asserted on the old 39-char value
- [x] `uv run --project tools/skill-and-tool-validator skill-and-tool-validate` passes
- [x] `prek` commit hooks pass on the changed files

## Linked issues

Closes apache/magpie#952

### DIFF
--- a/skills/workflow-security-audit/SKILL.md
+++ b/skills/workflow-security-audit/SKILL.md
@@ -235,7 +235,7 @@ changing the tag.
 floating reference.
 
 **Suggested remediation:** pin to the full commit SHA of the version
-you trust — e.g. `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af68`
+you trust — e.g. `actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683`
 — and add a comment with the semantic version for readability.
 
 ### Fork-secret exposure (`fork-secrets`)
--- a/uv.lock
+++ b/uv.lock
@@ -369,7 +369,7 @@ dev = [
 
 [package.metadata.requires-dev]
 dev = [
-    { name = "mypy", specifier = ">=2.1.0" },
+    { name = "mypy", specifier = ">=2.3.0" },
     { name = "pytest", specifier = ">=9.1.1" },
     { name = "ruff", specifier = ">=0.15.21" },
 ]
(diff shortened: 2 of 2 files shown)


justinmclean (approved):
Thanks for the improvement

PR #968, opened 2026-07-27
fix(tools/security-tracker-stats-dashboard): align Python floor to 3.11

### LINKED ISSUE #948: Align the Python version floor in security-tracker-stats-dashboard
## Summary

`tools/security-tracker-stats-dashboard/pyproject.toml` declares three different
Python version floors across its own settings.

## Background

`requires-python` and ruff say 3.9, mypy says 3.10. mypy therefore type-checks
against a version the package claims to support running below, so a 3.10-only
construct type-checks clean and fails at import on 3.9. The `>=3.9` floor also
means PEP 604 `X | Y` unions must stay in annotations guarded by
`from __future__ import annotations`, never in runtime positions.

## Where to look

- `pyproject.toml` line 27 (`requires-python = ">=3.9"`), line 36 (ruff
  `target-version = "py39"`), line 44 (`[tool.mypy] python_version = "3.10"`).
- Sibling `tools/*/pyproject.toml` show what floor the rest of the repo settles on.

## Acceptance criteria

- [ ] All three settings agree on one floor.
- [ ] If the floor lands on 3.9, no runtime-position PEP 604 union remains.
- [ ] `uv run --directory tools/security-tracker-stats-dashboard --group dev pytest` passes.

## Estimated effort

~1 hour for someone new to the codebase, plus a look for runtime unions.

Shaurya2k06:
## Summary

- Align `requires-python`, ruff `target-version`, and mypy `python_version` in `tools/security-tracker-stats-dashboard` to the repo-wide **3.11** floor (they previously disagreed: 3.9 / py39 / 3.10).
- Update the README Runtime prerequisite to Python 3.11+ and refresh the tool `uv.lock` header.
- Apply the py311 ruff fixes the raised target now surfaces (`datetime.UTC` / `zip(..., strict=True)`).
- Re-lock root `uv.lock` `mypy` `requires-dev` specifiers to `>=2.3.0` so `prek` stops rewriting the lock when this tool's README is touched (same correction as #965 — drop that hunk if #965 lands first).

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `prek` commit hooks pass on the changed files
- [x] `uv run --directory tools/security-tracker-stats-dashboard --group dev pytest` — 114 passed
- [x] `uv run --directory tools/security-tracker-stats-dashboard --group dev ruff check` / `mypy` — clean
- [ ] Other:

## RFC-AI-0004 compliance

- [x] **Vendor neutrality** — placeholders used in skill / tool prose (N/A for this version-floor change)

## Linked issues

Closes #948

If #965 merges first, rebase and drop the root `uv.lock` hunk — it is only included so `prek`'s vendor-neutrality hook does not fail this branch on README touch.

### DIFF
--- a/tools/security-tracker-stats-dashboard/README.md
+++ b/tools/security-tracker-stats-dashboard/README.md
@@ -224,7 +224,7 @@ remaining charts still render.
 
 ## Prerequisites
 
-- **Runtime:** Python 3.9+ (`python3`). Run each stage as `python3 fetch_events.py`, etc.
+- **Runtime:** Python 3.11+ (`python3`). Run each stage as `python3 fetch_events.py`, etc.
   Optional: `pyyaml` — when missing, `render.py` uses a bundled minimal YAML subset parser
   sufficient for `default-config.yaml` and typical overlays; set `TRACKER_STATS_PY=uv-yaml`
   to pin clean PyYAML invocations via `uv run --with pyyaml`.
--- a/tools/security-tracker-stats-dashboard/pyproject.toml
+++ b/tools/security-tracker-stats-dashboard/pyproject.toml
@@ -24,7 +24,7 @@ name = "security-tracker-stats-dashboard"
 version = "0.1.0"
 description = "Rendering helpers for the security-tracker statistics dashboard."
 readme = "README.md"
-requires-python = ">=3.9"
+requires-python = ">=3.11"
 license = { text = "Apache-2.0" }
 dependencies = []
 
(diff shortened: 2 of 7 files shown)


justinmclean (approved):
Right alignment: this was the last tool on >=3.9 while root and the other 29 are >=3.11, and nothing in CI pins a Python version.

For anyone wondering why a version bump carries datetime.UTC and zip(..., strict=True): both ruff rules are silent at py39 and fire at py311, so they're what keeps ruff check green once target-version moves. datetime.UTC is the same object as datetime.timezone.utc, a pure rename. strict=True is the real choice, since strict=False would preserve the old truncating behaviour.

potiuk (approved):

PR #970, opened 2026-07-27
fix(skill-and-tool-validator): harden three non-blocking review leftovers

### LINKED ISSUE #941: Fix three robustness defects in skill-and-tool-validator
## Summary

Three small, independent defects in the validator and its test suite: a
docstring unit error, a test that aborts instead of skipping, and an unguarded
directory read.

## Background

All three were raised in review, agreed as non-blocking, and merged without
being actioned. None changes behaviour on a healthy checkout; each degrades
badly in a specific environment.

## Where to look

- `.../skill_and_tool_validator/__init__.py` line 2488 - docstring says the
  content is shorter than `_MIN_LICENSE_FILE_SIZE` **characters**, but line 2498
  compares `path.stat().st_size`, which is **bytes**. Equivalent for ASCII; a
  file with a BOM or multi-byte content measures differently.
- `tools/skill-and-tool-validator/tests/test_validator.py` line 1686 -
  `test_real_repo_tool_python_files_all_have_headers` calls `find_repo_root()`
  with no guard. In a shallow or out-of-tree run it aborts the session rather
  than skipping. Add a `skipif`.
- `.../__init__.py` line 3285 - `validate_eval_coverage` calls `.iterdir()` with
  no `try`/`except`. A restrictive CI runner gets an unhandled `PermissionError`
  instead of a graceful violation. `collect_tool_python_files` shows the pattern.

## Acceptance criteria

- [ ] The docstring states bytes.
- [ ] The repo-wide header test skips rather than errors, with a test proving it.
- [ ] `validate_eval_coverage` handles `OSError` as `collect_tool_python_files` does.
- [ ] `uv run --directory tools/skill-and-tool-validator --group dev pytest` passes.

## Estimated effort

Shaurya2k06:
## Summary

- Fix the `collect_tool_python_files` docstring (and size-threshold comment) so it says **bytes**, matching `path.stat().st_size`.
- Guard the live license-header integration with a `skipif` when the checkout is not a full magpie tree; add a unit test that proves the skip condition fires out-of-tree.
- Catch `OSError` from `validate_eval_coverage`'s `iterdir()` calls (skills/ and evals/) the same way `collect_tool_python_files` does, so a restrictive runner gets a quiet skip instead of an unhandled `PermissionError`.
- Re-lock root `uv.lock` `mypy` `requires-dev` specifiers to `>=2.3.0` so `prek` stops rewriting the lock when this tool is touched (same correction as #965 — drop that hunk if #965 lands first).

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `uv run --directory tools/skill-and-tool-validator --group dev pytest` — all passed
- [x] Targeted: live-header skip proving test, live-header integration, unreadable skills/evals `PermissionError` cases
- [x] `prek` commit hooks pass on changed files (`ruff` / `mypy` / workspace `pytest` / `skill-and-tool-validate`)
- [x] Commit is GPG-signed and DCO-signed (`Signed-off-by`)

## RFC-AI-0004 compliance

- [x] **Vendor neutrality** — placeholders used in skill / tool prose (N/A for this robustness fix)

## Linked issues

Closes apache/magpie#941

### DIFF
--- a/tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
+++ b/tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
@@ -2469,8 +2469,8 @@ _LICENSE_PY_MARKERS: tuple[str, ...] = (
     "apache.org/licenses/LICENSE-2.0",
 )
 
-# Files smaller than this threshold (bytes / characters) are treated as
-# empty placeholder stubs and exempted from the license-header check.
+# Files smaller than this threshold (bytes) are treated as empty
+# placeholder stubs and exempted from the license-header check.
 _MIN_LICENSE_FILE_SIZE = 50
 
 # Path components that mark generated or vendored subtrees that must not
@@ -2485,7 +2485,7 @@ def collect_tool_python_files(root: Path | None = None) -> list[Path]:
 
     Excludes generated / vendored subtrees (``.venv``, ``site-packages``,
     ``node_modules``, ``__pycache__``) and empty placeholder files whose
-    content is shorter than ``_MIN_LICENSE_FILE_SIZE`` characters.
+    content is shorter than ``_MIN_LICENSE_FILE_SIZE`` bytes.
     """
     base = (root or find_repo_root()) / TOOLS_DIR
     if not base.exists():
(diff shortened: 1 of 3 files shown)


justinmclean (approved):
Solid. I checked the skipif first, since evaluating a helper in a decorator can break collection rather than skip if it raises: find_repo_root returns CWD when it finds no skills/, so out-of-tree the helper returns None and the test skips cleanly. Good.

PR #971, opened 2026-07-27
fix(release-verify-rc): use placeholders in step-6 artefact example

### LINKED ISSUE #945: Replace the hardcoded product identity in release-verify-rc step 6
## Summary

`skills/release-verify-rc/SKILL.md` uses a concrete Apache Airflow artefact name
as its step-6 worked example, in a skill whose convention is placeholder tokens.

## Background

The skill is meant to run for any adopting project, and its placeholder
convention uses `<version>` and `<upstream>`-style tokens so an adopter's config
supplies the real values. Step 6 instead hardcodes
`apache-airflow-2.11.0-source-release`, which reads as though the skill is
Airflow-specific and gives an adopter a value to copy that is wrong for them.

## Where to look

- `skills/release-verify-rc/SKILL.md` lines 453-454 - the worked example.
- The placeholder convention comment near the top of the same file.
- Sibling release skills show the placeholder style in their own examples.

## Acceptance criteria

- [ ] The step-6 example uses placeholder tokens rather than a concrete product
      and version.
- [ ] The example still shows the extension-stripping behaviour it teaches.
- [ ] `skill-and-tool-validate` passes, including the placeholder check.

## Estimated effort

Shaurya2k06:
## Summary

- Replace the concrete `apache-airflow-2.11.0-source-release` worked example in
  `release-verify-rc` step 6 with the skill's existing `<artefact-source-release>`
  placeholder so the extension-stripping lesson stays project-agnostic.
- Keeps the taught behaviour: strip the archive extension (`.tar.gz` / `.zip`),
  do not drop the `-source-release` suffix.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `uv run --project tools/skill-and-tool-validator skill-and-tool-validate` passes
- [x] `tools/dev/check-placeholders.sh` passes
- [x] `prek run --files skills/release-verify-rc/SKILL.md` passes (includes
      `check-placeholders` + `skill-and-tool-validate`)
- [x] Confirmed no remaining concrete artefact name in the step-6 example
- [ ] For skill *behaviour* changes: n/a — prose-only placeholder example swap;
      eval fixtures intentionally keep concrete names under the allowlisted
      `tools/skill-evals/evals/` path

## RFC-AI-0004 compliance

## Linked issues

Closes apache/magpie#945

### DIFF
--- a/skills/release-verify-rc/SKILL.md
+++ b/skills/release-verify-rc/SKILL.md
@@ -450,8 +450,8 @@ classification below (matching files become `expected_binaries`, the
 rest `prohibited_found`), not by filtering the command.
 
 `<unpacked-dir>` is the source artefact filename with its archive
-extension removed: `apache-airflow-2.11.0-source-release.tar.gz` unpacks
-to `apache-airflow-2.11.0-source-release`. Do not drop the
+extension removed: `<artefact-source-release>.tar.gz` unpacks
+to `<artefact-source-release>`. Do not drop the
 `-source-release` suffix or substitute a shortened name. Resolve
 `<unpacked-dir>` to this concrete directory before emitting the recipe.
 


justinmclean:
The placeholder swap is right in principle, but these lines are inside ## Step 6 — Binary exclusion check, and step-config.json extracts that step as the system prompt for step-6-binary-exclusion. Both of its cases expect paste_recipe to contain find apache-airflow-2.11.0-source-release …, and that field is judge-graded, so this is a prompt change with a fixture dependency. Could you run that suite before this merges? Note: I ran it and am seeing failures.

### `skills/release-verify-rc/SKILL.md:453`

The lines are inside Step 6, and `step-config.json` for `step-6-binary-exclusion` extracts that step as the eval's system prompt. Both cases expect `paste_recipe` to contain `find apache-airflow-2.11.0-source-release -type f ...`, derived from the report's `apache-airflow-2.11.0-source-release.tar.gz`, and `paste_recipe` is prose-graded by a judge. The worked example you removed is what made that derivation obvious. It may well still pass, but the test plan lists the validator, the placeholder check and prek, not the suite that actually covers this:

```
uv run --directory tools/skill-evals skill-eval --cli "claude -p" \
  evals/release-verify-rc/step-6-binary-exclusion/fixtures/
```

Second, the example no longer demonstrates anything. `<artefact-source-release>.tar.gz` unpacks to `<artefact-source-release>` is a tautology, and the very next sentence warns "do not drop the `-source-release` suffix" — a suffix now visible only inside the placeholder's own name. Keeping a worked shape without the product identity would satisfy the issue and keep the lesson:
```
`apache-<project>-<version>-source-release.tar.gz` unpacks
to `apache-<project>-<version>-source-release`
```

Both `<project>` and `<version>` are already used as placeholders across the release skills.


This review was drafted by an AI-assisted tool and
confirmed by a Magpie maintainer. The findings
below are observations, not blockers; a Magpie
maintainer — a real person — will take the next look at the
PR. If you think a finding is mis-applied, please reply on
the PR and a maintainer will weigh in.

More on how Magpie handles maintainer review:
[[CONTRIBUTING.md](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md)](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md).

potiuk (approved):
LGTM — this is a straight application of the project's own rule, and it fixes a real leak. [`AGENTS.md` § Placeholder convention used in skill files](https://github.com/apache/magpie/blob/main/AGENTS.md#placeholder-convention-used-in-skill-files) is explicit:

> Writing a literal project value directly into a skill is a refactor bug — skills must stay project-agnostic so swapping projects is a config change, not a code change.

`apache-airflow-2.11.0-source-release.tar.gz` was exactly that: one adopter's concrete artefact name baked into a governance-agnostic skill.

Two things I checked that make this the right shape of fix rather than a new convention:

- The placeholder is **not invented here** — `<artefact-source-release>` is already in use 96 lines earlier in this same file, in the unpack recipe itself (`tar -xf <artefact-source-release>.tar.gz   # or .zip`). This change makes the prose at line 453 agree with the command at line 357 instead of contradicting it.
- The surrounding warning still lands. *"Do not drop the `-source-release` suffix or substitute a shortened name"* reads fine against the placeholder — arguably better, since the placeholder name itself carries the `-source-release` suffix the sentence is telling the agent to preserve.

### Smaller observations

- Not this PR's problem, and explicitly **not** something to fix here: `<artefact-source-release>` is not listed in the `AGENTS.md` placeholder table, even though that section says *"Do not invent new placeholders."* That gap predates this change (line 357 already relies on it), and your PR reduces the inconsistency rather than adding to it. It is worth a separate follow-up to either register the placeholder in the table or thread the value through the project manifest — I will track that separately, so please don't expand this PR's scope.

Thanks — small, surgical, and it cites the right convention.

---

> will follow up.*
>

PR #984, opened 2026-07-28
test(skill-and-tool-validator): assert org-structure check via run_validation

### LINKED ISSUE #977: Add a run_validation test so the organization-structure check cannot be silently unwired
## Summary

The validator's organization-structure check has eight tests, all of which call the check
function directly. None of them notice if the check stops being called by the validator's entry
point. Add one test that goes through `run_validation`.

## Background

`validate_organization_structure` reaches real runs through a single line in
`run_validation()`. Every test in `TestOrganizationStructure` imports and calls the function
itself, so deleting that one line leaves the whole suite green while the check silently stops
running for everyone.

There is one existing test that calls `run_validation()` against the real repository, but it
asserts that no hard violations exist — and since the live `ASF/` and `independent/` adapters
both have their required files, it passes whether or not the check is wired in.

This is a small, self-contained gap and a good way to see how the validator is assembled.

## Where to look

- `tools/skill-and-tool-validator/tests/test_validator.py` — the `TestOrganizationStructure`
  class is where the new test belongs. Its `_make_org` helper already builds an org directory
  under `tmp_path`.
- `tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py` — find
  `def run_validation` and look for the `validate_organization_structure` line to see the
  wiring you are protecting.
- Other tests in the same file call `run_validation(tmp_path)` and filter the result by
  category; use them as the model.

## Acceptance criteria

- [ ] A new test builds an organization directory missing a required file under `tmp_path` and
      asserts the violation appears in the output of `run_validation(tmp_path)`.
- [ ] The test fails when the `validate_organization_structure` line is removed from
      `run_validation` (check this, then put the line back).
- [ ] The validator's existing tests still pass.

## Estimated effort

## Getting started

Shaurya2k06:
## Summary

- Add one `TestOrganizationStructure` case that builds an org missing `README.md` under `tmp_path` and asserts the violation via `run_validation(tmp_path)`, filtered by `ORGANIZATION_CATEGORY`.
- Direct unit tests call `validate_organization_structure` themselves, so deleting its single wiring line in `run_validation` left the suite green; this routes through the entry point so the check cannot be silently unwired.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `uv run --project tools/skill-and-tool-validator pytest tools/skill-and-tool-validator/tests/` — 452 passed
- [x] Confirmed `test_missing_file_reached_via_run_validation` fails when `violations.extend(validate_organization_structure(repo_root))` is removed from `run_validation`, then restored
- [x] `prek run --files tools/skill-and-tool-validator/tests/test_validator.py` passes
- [ ] For skill *behaviour* changes: n/a — test-only wiring guard

## RFC-AI-0004 compliance

## Linked issues

Closes #977

### DIFF
--- a/tools/skill-and-tool-validator/tests/test_validator.py
+++ b/tools/skill-and-tool-validator/tests/test_validator.py
@@ -3086,6 +3086,18 @@ class TestOrganizationStructure:
     def test_organization_category_is_hard(self) -> None:
         assert ORGANIZATION_CATEGORY in HARD_CATEGORIES
 
+    def test_missing_file_reached_via_run_validation(self, tmp_path: Path) -> None:
+        """Guard the single wiring line in run_validation.
+
+        Direct unit tests call validate_organization_structure themselves, so
+        they stay green if that call is deleted from run_validation. This test
+        goes through the entry point so the check cannot be silently unwired.
+        """
+        self._make_org(tmp_path, "MyOrg", ["organization.md"])  # missing README.md
+        vs = [v for v in run_validation(tmp_path) if v.category == ORGANIZATION_CATEGORY]
+        assert len(vs) == 1
+        assert "README.md" in vs[0].message
+
 
 # ---------------------------------------------------------------------------
 # Non-ASF organization profile smoke coverage


potiuk (approved):
LGTM — and this is a better test than its size suggests. The docstring makes a specific, falsifiable claim, so I checked it rather than taking it on faith.

The claim: direct unit tests call `validate_organization_structure` themselves, so they stay green if the call is deleted from `run_validation`; this test goes through the entry point so the check cannot be silently unwired. The wiring really is a single line — `__init__.py:3716`, `violations.extend(validate_organization_structure(repo_root))`.

I deleted that line and re-ran the organization tests:

```text
FAILED tests/test_validator.py::TestOrganizationStructure::test_missing_file_reached_via_run_validation
```

…and that was the only failure. Every other organization test passed with the check completely unwired, which is exactly the gap the docstring describes. Restored the line and the full set goes green again.

That's the useful shape for this kind of test: it fails for the one reason it exists, and it isn't redundant with anything already in the file. The `assert "README.md" in vs[0].message` is a good touch too — it pins that the violation reaching the caller is the *right* one, not merely that some violation arrived.

No findings.

---

> will follow up.*
>

PR #986, opened 2026-07-28
test(cve-tool-vulnogram): assert CVE_JSON_CONFIG miss does not fall back

### LINKED ISSUE #976: Add a test that $CVE_JSON_CONFIG pointing at a missing file does not fall back to the legacy path
## Summary

`generate-cve-json` promises that an explicitly-chosen config path which does not exist is an
error, never a silent fallback to the old pre-rename location. That promise is tested for the
`--config` flag but not for the `$CVE_JSON_CONFIG` environment variable. Add the missing test.

## Background

The tool directory was renamed from `tools/vulnogram` to `tools/cve-tool-vulnogram`. Adopters
who upgraded across the rename still have their config at the old path, so the tool falls back
to it with a deprecation warning — but **only** when resolving from the default location. If the
user named a path explicitly (via `--config` or `$CVE_JSON_CONFIG`) and that path is missing,
the tool must raise rather than quietly read some other file.

`tests/test_config_resolution.py` covers five situations: default present, legacy fallback with
warning, default winning over legacy, both missing, and `--config` not falling back. The
environment-variable arm of the same guarantee has no test, even though it is the one reading
global process state and therefore the easier one to break by accident in a later refactor.

## Where to look

- `tools/cve-tool-vulnogram/generate-cve-json/tests/test_config_resolution.py` — add the test
  here. `test_explicit_config_path_does_not_fall_back_to_legacy` is the closest existing test
  to model it on.
- The `_clean_env` fixture at the top of that file already deletes `CVE_JSON_CONFIG` before
  every test, so your test only needs to set it.
- `tools/cve-tool-vulnogram/generate-cve-json/src/generate_cve_json/cve_json.py` — read
  `_resolve_config_path` and `_load_config` to see why the env var behaves like `--config`
  and not like the default path.

A working shape:

```python
def test_env_config_path_does_not_fall_back_to_legacy(tmp_path, monkeypatch):
    _write(tmp_path, _LEGACY_REL, "Legacy")
    monkeypatch.setenv("CVE_JSON_CONFIG", str(tmp_path / "missing.toml"))
    with pytest.raises(FileNotFoundError):
        _load_config()
```

## Acceptance criteria

- [ ] A new test sets `$CVE_JSON_CONFIG` to a path that does not exist, with the legacy config
      present on disk, and asserts `FileNotFoundError` is raised.
- [ ] The test fails if the fallback condition in `_load_config` is loosened to include the
      env-var case (check this by temporarily editing the condition, then reverting).
- [ ] The package's existing tests still pass.

## Estimated effort

~1 hour for someone new to the codebase, most of it spent getting the package's tests running
locally the first time.

## Getting started

Shaurya2k06:
## Summary

- Add a regression test that `$CVE_JSON_CONFIG` pointing at a missing file raises `FileNotFoundError` even when the legacy pre-rename config path is present on disk.
- Mirrors the existing `--config` coverage so a later refactor cannot silently widen the legacy fallback to the env-var arm.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `uv run --directory tools/cve-tool-vulnogram/generate-cve-json --group dev pytest` — all passed
- [x] `uv run --directory tools/cve-tool-vulnogram/generate-cve-json --group dev ruff check src tests`
- [x] `uv run --directory tools/cve-tool-vulnogram/generate-cve-json --group dev ruff format --check src tests`
- [x] `uv run --directory tools/cve-tool-vulnogram/generate-cve-json --group dev mypy`
- [x] Negative check: temporarily loosened `_load_config` fallback from `is_default and legacy.is_file()` to `legacy.is_file()` → new test failed with `DID NOT RAISE FileNotFoundError`; condition restored
- [x] `prek` commit hooks passed on the changed file
- [x] Commit is GPG-signed and DCO-signed (`Signed-off-by`)

## RFC-AI-0004 compliance

- [x] **Vendor neutrality** — placeholders used in skill / tool prose (N/A for this test-only change)

## Linked issues

Closes #976

Production behaviour is unchanged — `_resolve_config_path` already returns `is_default=False` for `$CVE_JSON_CONFIG`, so `_load_config` already raises. This PR only closes the missing env-var arm of the regression suite.

### DIFF
--- a/tools/cve-tool-vulnogram/generate-cve-json/tests/test_config_resolution.py
+++ b/tools/cve-tool-vulnogram/generate-cve-json/tests/test_config_resolution.py
@@ -71,3 +71,12 @@ def test_explicit_config_path_does_not_fall_back_to_legacy(tmp_path):
     _write(tmp_path, _LEGACY_REL, "Legacy")
     with pytest.raises(FileNotFoundError):
         _load_config(tmp_path / "explicit-but-missing.toml")
+
+
+def test_env_config_path_does_not_fall_back_to_legacy(tmp_path, monkeypatch):
+    # Same guarantee as --config: $CVE_JSON_CONFIG naming a missing path must
+    # raise, even when the legacy pre-rename config is present on disk.
+    _write(tmp_path, _LEGACY_REL, "Legacy")
+    monkeypatch.setenv("CVE_JSON_CONFIG", str(tmp_path / "missing.toml"))
+    with pytest.raises(FileNotFoundError):
+        _load_config()


justinmclean (approved):
LGTM - thanks for fixing this

PR #987, opened 2026-07-28
fix(release-verify-rc): align step 6 binary-exclude list with find

### LINKED ISSUE #940: Show how release-verify-rc loads the binary-exclude list in step 6
## Summary

Step 6 of `release-verify-rc` says it scans for files matching the
binary-exclude list from `release-build.md`, but the command it emits hardcodes
a fixed pattern set and never reads that file.

## Background

The step's prose points at adopter config as the source of the exclude list,
which is what makes the skill project-agnostic: a project with a different
prohibited-binary set is meant to configure it, not edit the skill. The emitted
command instead carries a fixed list of compiled Java classes, native shared
libraries and Python bytecode.

For an adopter whose `release-build.md` names anything outside that set, the
scan silently misses it while the step reports as though the configured list was
used. The step's classification output is fine; it is the gap between "from
`release-build.md`" and a hardcoded pattern list that misleads.

Either outcome closes this: show how the list is loaded and use it, or state
plainly that the pattern set is a fixed baseline and that `release-build.md`
supplies additions.

## Where to look

- `skills/release-verify-rc/SKILL.md` around lines 429-440 - step 6, its prose
  and the emitted scan command.
- `skills/release-verify-rc/SKILL.md` line 351 - the RAT step, which shows how
  another step reaches into `release-build.md` for its configuration. It is the
  pattern to follow if the answer is to load the list.
- `projects/_template/release-build.md` - what an adopter actually configures,
  and under which heading.
- `tools/skill-evals/evals/release-verify-rc/step-6-binary-exclusion/` - the eval
  cases, which will need to match whichever way this is resolved.

## Acceptance criteria

- [ ] Step 6's prose and its emitted command agree about where the exclude list
      comes from.
- [ ] If the list is loaded from config, the step shows how, in the style of the
      RAT step.
- [ ] If the pattern set stays fixed, the prose says so and explains how
      `release-build.md` extends it.
- [ ] The step-6 eval cases still pass.
- [ ] `uv run --project tools/skill-and-tool-validator --group dev skill-and-tool-validate`
      passes.

## Estimated effort

## Getting started

New to Magpie? Start with
[[Your first contribution](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md#your-first-contribution)](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md#your-first-contribution).

[[CONTRIBUTING.md](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md)](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md)

Shaurya2k06:
## Summary

- Align `release-verify-rc` Step 6 so prose and the emitted `find` agree:
  the scan always uses a **fixed baseline** of prohibited patterns; read
  `release-build.md` § Binary-exclude list (same file Step 4 uses for RAT)
  and append any extra globs beyond that baseline before emitting the recipe.
- Document the same split in `projects/_template/release-build.md` and
  `projects/magpie/release-build.md` (additional prohibited globs vs
  known-accepted exceptions).
- Refresh step-6 eval fixtures / `output-spec.md` to match.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `uv run --project tools/skill-and-tool-validator --group dev skill-and-tool-validate` passes
- [x] `PYTHONPATH=tools/skill-evals/src python3 -m skill_evals.runner tools/skill-evals/evals/release-verify-rc/step-6-binary-exclusion/` (print mode; fixtures extract cleanly)
- [x] `prek run --files` on touched paths passes (includes `check-placeholders`, `lychee`, `skill-and-tool-validate`)
- [x] For skill *behaviour* changes: step-6 fixtures / `output-spec.md` updated for fixed-baseline + additions wording

## RFC-AI-0004 compliance

## Linked issues

Closes apache/magpie#940

### DIFF
--- a/projects/_template/release-build.md
+++ b/projects/_template/release-build.md
@@ -78,18 +78,25 @@ Example shape:
 
 ## Binary-exclude list
 
-TODO: list any binary content the source artefact must NOT contain
-(per `release-verify-rc`'s no-prohibited-binaries check). The
-default list is conservative, `.class`, `.jar`, `.so`, `.dylib`,
-`.dll`, `.exe`, pre-built minified JS bundles checked into the
-source tree. Project-specific exclusions go here.
+TODO: configure `release-verify-rc` Step 6's prohibited-binary check.
+
+The skill always scans a **fixed baseline**: `.class`, `.jar`,
+`.so`, `.dylib`, `.dll`, `.exe`, `.pyc`, and `__pycache__`
+directories. List here:
+
+1. **Additional prohibited globs** the source artefact must also not
+   contain (appended to the Step 6 `find` beyond the baseline).
+2. **Known-accepted exceptions** — specific paths that match a
+   prohibited pattern but must ship; Step 6 classifies those as
+   `EXPECTED-BINARY` rather than `PROHIBITED-BINARY`.
 
 Example shape:
 
-> - `*.class`, `*.jar`, Java compiled output never ships in source.
-> - `assets/vendor/**/*.min.js`, vendored minified JS that has a
->   source-checked counterpart; flagged on every source-release
->   verification.
+> - `assets/vendor/**/*.min.js` — additional prohibited glob
+>   (vendored minified JS; flagged on every source-release
+>   verification).
+> - `third-party/some-native.so` — known-accepted exception (ships
+>   with a documented source counterpart).
 
 ## Apache RAT configuration
 
--- a/projects/magpie/release-build.md
+++ b/projects/magpie/release-build.md
@@ -69,5 +69,7 @@ and are never emitted.
 ## Binary-exclude list
 
 The source artefact must contain no compiled or opaque binary content.
-Conservative default denylist for `release-verify-rc`:
-`.class`, `.jar`, `.so`, `.dylib`, `.dll`, `.exe`, `.pyc`.
+`release-verify-rc` Step 6 always scans its fixed baseline
+(`.class`, `.jar`, `.so`, `.dylib`, `.dll`, `.exe`, `.pyc`,
+`__pycache__`). Magpie adds no project-specific globs and names no
+known-accepted exceptions.
--- a/skills/release-verify-rc/SKILL.md
+++ b/skills/release-verify-rc/SKILL.md
@@ -428,26 +428,41 @@ small (version-string-only changes).
 
 ## Step 6 — Binary exclusion check
 
-Scan the unpacked source artefact for files matching the
-binary-exclude list from `release-build.md`. Emit the paste-ready
-scan command:
+Scan the unpacked source artefact for prohibited binaries. The
+paste-ready `find` always starts from a **fixed baseline** of
+patterns; it is not generated wholesale from adopter config.
+
+Using the Binary-exclude list from `release-build.md` (heading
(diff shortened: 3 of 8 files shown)


potiuk (approved):
LGTM — and this is a more valuable fix than "align the docs" suggests. The old wording wasn't merely ambiguous, it was inverted in a way that could have let a non-compliant source release pass its own pre-flight.

### What was actually broken

Two documents disagreed about what the Binary-exclude list *means*:

- [`projects/_template/release-build.md`](https://github.com/apache/magpie/blob/main/projects/_template/release-build.md) asked the adopter to *"list any binary content the source artefact **must NOT** contain"*, and its example listed `*.class`, `*.jar` — *"Java compiled output never ships in source."*
- `skills/release-verify-rc/SKILL.md:462` then said: *"A file that appears in the binary-exclude list of `release-build.md` is a **known-and-accepted** binary; surface it as `EXPECTED-BINARY`."*

So an adopter who followed the template's own example would have configured Step 6 to classify compiled Java as `EXPECTED-BINARY` — i.e. **accepted** — switching off the prohibition for precisely the files the template told them never to ship. An artefact containing `.class`/`.jar` would have come back `PASS`. In a skill whose job is to catch exactly that before a release vote, that's the failure mode worth fixing.

Splitting the section's two jobs — *additional prohibited globs* vs *known-accepted exceptions* — removes the inversion, and rewriting the template example so `*.class`/`*.jar` are no longer list entries removes the trap that led to it.

### The other drift you caught

The baseline `find` in Step 6 already scanned `*.pyc` and `__pycache__`, but three places disagreed with it:

- the prose immediately below said *"The default prohibited extensions are `.class`, `.jar`, `.so`, `.dylib`, `.dll`, `.exe`"* — silently dropping both;
- **both** eval fixtures' `paste_recipe` omitted them, so the fixtures contradicted the command the skill actually emits.

Making the baseline explicit and mandatory (*"Do not drop baseline predicates when the Binary-exclude list is empty or only restates the baseline"*) is the right guard, given `.pyc` presence is the tell that an artefact was zipped from a working tree rather than exported from the tag — which the existing comment already explained well.

I checked the change is complete rather than partial: `case-1`, `case-2`, and `output-spec.md` are all updated, and those are the only step-6 fixtures carrying a `paste_recipe`.

### Smaller observations

Neither blocking; both are follow-up material rather than anything to change here.

- **The two purposes now share one flat list**, distinguished only by trailing prose (`— additional prohibited glob` / `— known-accepted exception`). An agent reading an adopter's file has to infer the category from free text. Given the bug being fixed *was* a semantic ambiguity in this exact section, two sub-headings (or a two-column table) would make the distinction mechanical rather than interpretive. Worth a follow-up once this lands.
- **`assets/vendor/**/*.min.js` isn't `find` syntax.** The instruction says to "translate each into a `-name` or `-path` predicate", which is right, but `**` has no meaning to `find` — the shaped form is `-path '*/assets/vendor/*.min.js'` (find's `*` already crosses `/`). One parenthetical showing that translation would save the next reader a wrong-but-plausible `-path 'assets/vendor/**/*.min.js'`.

---

> will follow up.*
>

PR #988, opened 2026-07-28
Narrow sourcehut exception handlers

### LINKED ISSUE #953: Narrow two broad exception handlers in the SourceHut adapter
## Summary

Two `except Exception` handlers in `tools/sourcehut/` swallow every error,
including ones that should surface. Narrow them to the exceptions actually
expected.

## Background

CodeQL flagged both as empty-except. The concrete risk is a silently wrong
result rather than a crash: an unexpected error is discarded and the caller
carries on with incomplete data.

## Where to look

- `tools/sourcehut/src/magpie_sourcehut/client.py` line 82 - the handler around
  parsing an HTTP error body is `except Exception:` with a comment and `pass`.
  Only JSON-decoding failures are expected here, so
  `(json.JSONDecodeError, KeyError, TypeError)` is the intended set.
- `tools/sourcehut/src/magpie_sourcehut/lists.py` line 125 -
  `contextlib.suppress(Exception)` around a sort. If the sort fails, `emails[0]`
  is still used as the cover letter, so a malformed node silently yields the
  wrong description and author. `(TypeError, AttributeError)` is the expected
  set.

## Acceptance criteria

- [ ] Both handlers name the exceptions they expect.
- [ ] An unexpected exception propagates rather than being discarded.
- [ ] A test covers the expected-failure path for at least one of the two.
- [ ] `uv run --directory tools/sourcehut --group dev pytest` passes.

## Estimated effort

## Getting started

New to Magpie? Start with

amh1k:
## Summary

- Narrow error handling to only expected malformed-data failures, so unexpected errors now propagate instead of being silently discarded.
- Add regression coverage for the expected fallback paths and unexpected sort failures.
- Define the SourceHut development dependency group used by its local quality checks.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `uv run --directory tools/sourcehut --group dev ruff check .`
- [x] `uv run --directory tools/sourcehut --group dev mypy`
- [x] `uv run --directory tools/sourcehut --group dev pytest` — 17 passed
- [x] `uv lock --check`
- [ ] Other:

## RFC-AI-0004 compliance

Not applicable: this change only narrows local Python exception handling and adds tests.

## Linked issues

Closes #953

The HTTP-error fallback now catches only `json.JSONDecodeError`, `KeyError`, and `TypeError`. Email sorting suppresses only `TypeError` and `AttributeError`; unexpected failures propagate rather than allowing an incorrect cover-letter selection.

### DIFF
--- a/tools/sourcehut/pyproject.toml
+++ b/tools/sourcehut/pyproject.toml
@@ -28,6 +28,13 @@ requires-python = ">=3.11"
 license = { text = "Apache-2.0" }
 dependencies = []
 
+[dependency-groups]
+dev = [
+  "ruff>=0.15.21",
+  "mypy>=2.3.0",
+  "pytest>=9.1.1",
+]
+
 [project.scripts]
 magpie-sourcehut = "magpie_sourcehut:main"
 
--- a/tools/sourcehut/src/magpie_sourcehut/client.py
+++ b/tools/sourcehut/src/magpie_sourcehut/client.py
@@ -79,7 +79,7 @@ def query_graphql(service: str, query: str, variables: dict[str, Any] | None = N
             if err_errors:
                 err_msgs = [e.get("message", "Unknown error") for e in err_errors]
                 err_msg = f"HTTP {exc.code}: {'; '.join(err_msgs)}"
-        except Exception:
+        except (json.JSONDecodeError, KeyError, TypeError):
             # Ignore errors parsing the HTTP error response body as JSON
             pass
 
--- a/tools/sourcehut/src/magpie_sourcehut/client.py
+++ b/tools/sourcehut/src/magpie_sourcehut/client.py
@@ -79,8 +79,17 @@ def query_graphql(service: str, query: str, variables: dict[str, Any] | None = N
             if err_errors:
                 err_msgs = [e.get("message", "Unknown error") for e in err_errors]
                 err_msg = f"HTTP {exc.code}: {'; '.join(err_msgs)}"
-        except (json.JSONDecodeError, KeyError, TypeError):
-            # Ignore errors parsing the HTTP error response body as JSON
+        except (ValueError, AttributeError, TypeError, OSError):
+            # Ignore errors parsing the HTTP error response body as JSON. This is
+            # a best-effort attempt to extract a nicer message from an untrusted
+            # body, so failing to parse it must never be worse than not trying:
+            # control falls through to the `status`-only SourceHutError below.
+            # `ValueError` covers both json.JSONDecodeError and the
+            # UnicodeDecodeError that a non-UTF-8 body raises; `AttributeError`
+            # covers a body that is valid JSON but not an object (`null`, an
+            # array, a bare string) or an `errors` entry that is not a dict;
+            # `OSError` covers a failed `exc.read()`. Programming errors
+            # (NameError, RuntimeError, ...) are deliberately left to propagate.
             pass
 
         if err_msg:
--- a/tools/sourcehut/src/magpie_sourcehut/lists.py
+++ b/tools/sourcehut/src/magpie_sourcehut/lists.py
@@ -122,7 +122,7 @@ def map_patchset_to_pr(patchset: dict[str, Any]) -> dict[str, Any]:
     emails = [edge.get("node") for edge in edges if edge and edge.get("node")]
 
     # Sort emails by date if possible
-    with contextlib.suppress(Exception):
+    with contextlib.suppress(TypeError, AttributeError):
         emails.sort(key=lambda x: x.get("date", ""))
 
     # The cover letter / description is the first email (or patchset subject)
--- a/tools/sourcehut/src/magpie_sourcehut/lists.py
+++ b/tools/sourcehut/src/magpie_sourcehut/lists.py
@@ -121,9 +121,16 @@ def map_patchset_to_pr(patchset: dict[str, Any]) -> dict[str, Any]:
     edges = emails_conn.get("edges") or []
     emails = [edge.get("node") for edge in edges if edge and edge.get("node")]
 
-    # Sort emails by date if possible
+    # Sort emails by date. `emails[0]` is taken as the cover letter below, so a
+    # failed sort must not leave the list half-ordered: `list.sort()` mutates in
+    # place and can raise part-way through a comparison, which would silently
+    # promote a reply to cover letter. Sort a copy and only adopt it on success.
+    # A `None` date is the common case (the `.get` default only covers a missing
+    # key), so it is folded into the key rather than left to raise; undated
+    # emails sort last and keep their input order rather than displacing a real
+    # cover letter.
     with contextlib.suppress(TypeError, AttributeError):
-        emails.sort(key=lambda x: x.get("date", ""))
+        emails = sorted(emails, key=lambda x: (x.get("date") is None, x.get("date") or ""))
 
     # The cover letter / description is the first email (or patchset subject)
     description = ""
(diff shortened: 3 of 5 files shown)


justinmclean:
I ran both changed handlers with the patch applied.

Inline comment 1 — tools/sourcehut/src/magpie_sourcehut/client.py:82

This narrowing leaks exceptions that query_graphql previously converted into SourceHutError. Every other error path in this function raises SourceHutError; after this change four realistic HTTP-error bodies escape as raw exceptions instead. Driving the real function with a mocked HTTPError:

text
plain text body        -> SourceHutError: HTTP request to ... failed with status 400   (ok)
non-utf8 body          -> LEAKED UnicodeDecodeError
JSON array body        -> LEAKED AttributeError: 'list' object has no attribute 'get'
JSON string body       -> LEAKED AttributeError: 'str' object has no attribute 'get'
{"errors":"boom"}      -> LEAKED AttributeError: 'str' object has no attribute 'get'

exc.read().decode("utf-8") raises UnicodeDecodeError, which is a ValueError but not a json.JSONDecodeError, so a latin-1 error page from a proxy now crashes the caller. err_json.get(...) raises AttributeError whenever the body is valid JSON that is not an object, and e.get(...) does the same when errors is a string. Separately, KeyError is unreachable here: the block only uses .get(), never subscripting.

suggestion
        except (ValueError, AttributeError, TypeError):

ValueError covers json.JSONDecodeError and UnicodeDecodeError both.

Inline comment 2 — tools/sourcehut/src/magpie_sourcehut/lists.py:125

The issue asks for the handler to stop "silently yielding the wrong description and author". Narrowing Exception to TypeError does not achieve that, because TypeError is the failure mode: a null date among the emails raises it mid-sort, and list.sort leaves the list in a partially reordered state when the key comparison raises. emails[0] is then not the cover letter. Running the real map_patchset_to_pr with the patch applied, on a nine-email patchset with one null date:

text
input first email      : E6
PR description/author  : E0 / E0      <- wrong cover letter and wrong author

Over 2000 randomised patchsets of 3 to 12 emails with one null date, 920 came back reordered. The PR notes say "unexpected failures propagate rather than allowing an incorrect cover-letter selection", but the suppressed TypeError is precisely the path that allows it. Sorting on a total-ordering-safe key removes the failure instead of relabelling it:

suggestion
    emails.sort(key=lambda x: x.get("date") or "")

If you prefer to keep a guard, sort a copy and only assign it back on success, so a raise cannot leave emails half-ordered.

Inline comment 3 — tools/sourcehut/tests/test_sourcehut.py, test_map_patchset_to_pr_keeps_input_order_when_dates_cannot_be_compared

This test asserts behaviour CPython does not guarantee. It passes only because two elements raise on the first comparison before any swap. Add a third and fourth email with the null date in the middle and the assertion fails. As written it locks in an implementation detail and gives false confidence that the wrong-cover-letter bug is covered. Once comment 2 is addressed the test should assert the sorted result directly, with null-dated emails landing in a defined position.

Credit where due: test_map_patchset_to_pr_propagates_unexpected_sort_errors is a good test and does what it claims  and the acceptance criterion on exception-path coverage is met.

potiuk (approved):
Approving with the fixes applied directly — I pushed one commit (`aac33a2`) implementing @justinmclean's two findings, so this can land without another round-trip. Please read what I changed rather than take it on trust; details and reasoning below.

First, credit where it's due: **@justinmclean's review is the substantive one here**, and on the `lists.py` half he was right where my own first read was wrong. I had signed off on `(TypeError, AttributeError)` as correct narrowing. It isn't, and his empirical check (920 of 2000 randomised patchsets coming back reordered) is what settles it.

### `lists.py` — narrowing didn't fix the bug, because `TypeError` *is* the bug

The change set out to stop the function "silently yielding the wrong description and author". Suppressing `TypeError` cannot achieve that: a `None` date raises `TypeError` mid-sort, and `list.sort()` mutates in place, so it can leave the list **partially reordered** when a comparison raises. `emails[0]` is then a reply rather than the cover letter — precisely the symptom the PR was written to remove.

Fixed at the source instead of relabelling it:

```python
with contextlib.suppress(TypeError, AttributeError):
    emails = sorted(emails, key=lambda x: (x.get("date") is None, x.get("date") or ""))
```

Three things there, each load-bearing:

- **`x.get("date") or ""`** removes the common failure entirely. Worth noting *why* the original `x.get("date", "")` didn't already do this: the `.get` default only fires for a **missing** key, not for a key present with an explicit `null` — which is what the API actually returns.
- **`x.get("date") is None` as the first key element** sorts undated emails *last*. Folding `None` to `""` alone would sort them *first*, which trades a crashing bug for a deterministic wrong-cover-letter — an undated reply would take the slot. This is a small refinement of your suggestion, @justinmclean, for that reason; a stable sort keeps undated emails in input order within their group.
- **`sorted(...)` assigned on success** rather than `.sort()` in place, so a raise on a genuinely uncomparable key (a numeric date, say) cannot leave `emails` half-ordered. That is your "sort a copy and only assign it back on success", which is the part that makes the remaining guard safe.

### `client.py` — the tuple broke the "never worse than not trying" guarantee

Both of us landed on this one independently, with the same list of escaping shapes. The suppressed block is a best-effort attempt to extract a nicer message from an untrusted error body; control is *meant* to fall through to the status-only `SourceHutError`. I verified the escapes against the real function:

| Error body | Old tuple | Now |
|---|---|---|
| `not json` | `SourceHutError` | `SourceHutError` |
| `b"\xff\xfe"` non-UTF-8 | **leaked `UnicodeDecodeError`** | `SourceHutError` |
| `null` | **leaked `AttributeError`** | `SourceHutError` |
| `[1, 2]` | **leaked `AttributeError`** | `SourceHutError` |
| `"oops"` | **leaked `AttributeError`** | `SourceHutError` |
| `{"errors": "boom"}` | **leaked `AttributeError`** | `SourceHutError` |
| `{"errors": ["boom"]}` | **leaked `AttributeError`** | `SourceHutError` |

Widened to `(ValueError, AttributeError, TypeError, OSError)`. `ValueError` covers `json.JSONDecodeError` and `UnicodeDecodeError` both — the trap being that `UnicodeDecodeError` is a `ValueError` but *not* a `JSONDecodeError`. I added `OSError` on top of @justinmclean's suggested triple, for a `exc.read()` that fails on an already-closed connection; same reasoning as the rest, a failed read shouldn't be worse than not reading. `KeyError` is gone — as noted, unreachable, since the block only calls `.get()`.

### Tests

`test_map_patchset_to_pr_keeps_input_order_when_dates_cannot_be_compared` asserted behaviour CPython does not guarantee — it passed only because two elements raise on the first comparison before any swap. Replaced with a four-email case (undated one in the middle) asserting the resulting order directly, plus an uncomparable-key case asserting the input order survives.

The error-body cases are now a parametrised sweep of all seven shapes. **Six of the seven fail against the previous tuple** — I checked by reverting the fix and re-running, so these are real regression tests rather than tests that merely pass.

`test_map_patchset_to_pr_propagates_unexpected_sort_errors` is kept as-is. It was a good test, it does what it claims, and it still passes — a `RuntimeError` from the key function propagates.

Verified locally: 25 tests pass, `ruff check`, `ruff format --check`, and `mypy` are clean on `tools/sourcehut`, and the workspace prek hooks pass.

Also worth crediting on the packaging change: the new `dev` group brings `tools/sourcehut` in line with 23 of the 25 tool projects, and your pins (`ruff>=0.15.21`, `mypy>=2.3.0`, `pytest>=9.1.1`) match the **workspace root** exactly, which is what that block's comment asks for. 18 of the tools have drifted to `mypy>=2.1.0`, so copying a neighbour would have been the easy wrong answer.

Thanks for this — the instinct to kill blanket `except Exception` was right, and the narrowing bug it surfaced in `map_patchset_to_pr` was worth finding on its own.

---

> will follow up.*
>

PR #928, opened 2026-07-27
chore(ai-tutors): regenerate knowledge base for lesson-04-your-first-skill

justinmclean:
## Summary

Step-6 (debugging-skills) and step-7 (portable-skills) were added to the learning progression after this tutor was authored; the injector now adds them to the "Where to go next" section and renumbers the eval-driven- development reference from step-5 to steps-5–7.

Generated-by: Claude (claude-sonnet-4-6)

## Type of change

- [ ] Skill change (`.claude/skills/<name>/`) — eval fixtures updated below
- [ ] Tool / bridge contract (`tools/<system>/*.md`)
- [ ] Python package (`tools/*/` with `pyproject.toml`)
- [ ] Groovy reference impl
- [ ] Cross-cutting (RFC, AGENTS.md, sandbox, privacy-LLM)
- [X] Documentation (`docs/`, `README.md`, `CONTRIBUTING.md`)
- [ ] Project template (`projects/_template/`)
- [ ] CI / dev loop (`prek`, workflows, validators)
- [ ] Other:

## Test plan

- [X ] `prek run --all-files` passes
- [ ] For Python packages touched: `uv run pytest` / `ruff check` / `mypy` passes
- [ ] For Groovy bridges touched: command-line invocation tested end-to-end
- [ ] For skill changes: eval suite passes for the affected skill
      (`PYTHONPATH=tools/skill-evals/src python3 -m skill_evals.runner tools/skill-evals/evals/<skill>/`)
- [ ] For skill *behaviour* changes: a new or updated eval fixture is included in this PR
      (a regression test for the bug fixed / the behaviour added — see CONTRIBUTING.md)
- [ ] Other:

PR #1010, opened 2026-07-29
fix(ai-tutors): handle nested code fences

### LINKED ISSUE #1008: Fix the nested-code-fence bug in inject-knowledge-base.py
## Summary

`tag_bare_code_fences` in `ai-tutors/inject-knowledge-base.py` tracks code-fence state with a
check that any line starting with three backticks toggles it. A source page using four-backtick
fences flips that state and never recovers, so the function silently stops doing its job.

## Background

The function exists because markdownlint's MD040 rejects a fenced code block whose opening fence
names no language. When a source page is embedded into a tutor prompt, a bare ``` fence in the
source would fail the lint in the *generated* file, where the author never sees it. So the
injector gives every language-less opening fence a default language.

The state machine is:

```python
if stripped.startswith("```"):
    if not in_code:
        in_code = True
        if stripped == "```":            # opening fence with no language
            lines[i] = f"{indent}```{default_lang}"
    else:
        in_code = False
```

Now walk a page that wraps an example in four backticks — an opening ````markdown, an inner
bare ``` fence, then a closing ````. The opening four-backtick line sets `in_code = True`
(correctly not tagged, since it is not exactly three backticks). The **inner** bare fence then
sets `in_code = False` even though the outer block has not closed. The real closing line is now
read as an *opening* fence, and the state stays inverted for the rest of the file — so a later
genuine bare opening fence is treated as a closing one and never gets its language.

No source page currently injected uses four-backtick fences, so this is latent rather than
breaking today. `docs/education/pattern-catalogue.md` does use them (around lines 243/268 and
324/336) and is a plausible future knowledge base. The failure is silent: the output looks
plausible and markdownlint fails somewhere that looks unrelated.

The fix is to track the length of the opening fence and only close on a fence at least as long.

## Where to look

- `ai-tutors/inject-knowledge-base.py` — `tag_bare_code_fences`, near the top of the file; the
  docstring explains the intent.
- `docs/education/pattern-catalogue.md` — a real page with nested four-backtick fences, useful
  as test input.
- `ai-tutors/README.md` — how the injector is run.

## Acceptance criteria

- [ ] A source page with a four-backtick block containing a bare three-backtick fence is
      processed without inverting the state: the inner fence is left alone and fences after the
      outer block are still tagged correctly.
- [ ] Fences that already name a language are still untouched, and indentation is still
      preserved.
- [ ] A test covers the nested case. (If the script has no test file yet, adding a small one
      alongside it is in scope for this issue.)
- [ ] Running `python3 ai-tutors/inject-knowledge-base.py --check` reports no unexpected
      staleness.

## Estimated effort

~2 hours for someone new to the codebase, including writing the test.

## Getting started

pollychen-lab:
## Summary

- Fix `tag_bare_code_fences` to track the length of the currently open backtick fence.
- Leave shorter nested fences inside longer fenced blocks untouched while still tagging later top-level bare fences.
- Add a regression test for the nested four-backtick/triple-backtick case and refresh the generated tutor prompt so `--check` passes.

Closes #1008.

## Validation

- `uv run pytest ai-tutors/test_inject_knowledge_base.py`
- `python3 ai-tutors/inject-knowledge-base.py --check`
- `uv run ruff check ai-tutors/inject-knowledge-base.py ai-tutors/test_inject_knowledge_base.py`
- `uv run ruff format --check ai-tutors/inject-knowledge-base.py ai-tutors/test_inject_knowledge_base.py`

## Generative AI Disclosure

This PR was prepared with AI assistance.

Generated-by: Codex (GPT-5)

### DIFF
--- a/ai-tutors/inject-knowledge-base.py
+++ b/ai-tutors/inject-knowledge-base.py
@@ -59,6 +59,7 @@ DOCTOC_RE = re.compile(
     re.DOTALL,
 )
 SPDX_RE = re.compile(r"<!-- SPDX-License-Identifier:.*?-->\n*", re.DOTALL)
+CODE_FENCE_RE = re.compile(r"^(?P<indent>\s*)(?P<fence>`{3,})(?P<info>[^`]*)$")
 
 
 @dataclass(frozen=True)
@@ -128,24 +129,22 @@ def tag_bare_code_fences(text: str, default_lang: str = "text") -> str:
     that already name a language are left untouched.
     """
     lines = text.split("\n")
-    in_code = False
-    fence_len = 0  # backticks in the opening fence of the current block
+    open_fence_length: int | None = None
     for i, line in enumerate(lines):
-        stripped = line.strip()
-        if not stripped.startswith("```"):
+        match = CODE_FENCE_RE.match(line)
+        if not match:
             continue
-        backticks = len(stripped) - len(stripped.lstrip("`"))
-        if not in_code:
-            in_code = True
-            fence_len = backticks
-            if stripped == "```":  # opening fence with no language
-                indent = line[: len(line) - len(line.lstrip())]
-                lines[i] = f"{indent}```{default_lang}"
-        elif backticks >= fence_len:
-            # a closing fence must be at least as long as the opening fence
-            in_code = False
-            fence_len = 0
-        # a shorter backtick run inside the block is literal content; leave as-is
+
+        fence = match.group("fence")
+        info = match.group("info").strip()
+        fence_length = len(fence)
+
+        if open_fence_length is None:
+            open_fence_length = fence_length
+            if not info:  # opening fence with no language
+                lines[i] = f"{match.group('indent')}{fence}{default_lang}"
+        elif not info and fence_length >= open_fence_length:
+            open_fence_length = None
     return "\n".join(lines)
 
 
--- a/ai-tutors/lesson-04-your-first-skill.md
+++ b/ai-tutors/lesson-04-your-first-skill.md
@@ -145,7 +145,7 @@ Apache-2.0 licensed.
 >
 > This is not the full authoring reference. Once you know the shape of a skill,
 > the `magpie-write-skill` (../../skills/write-skill/SKILL.md)
-> skill (you run it with `/write-skill`) takes you through every check, safety
+> skill (you run it with `/magpie-write-skill`) takes you through every check, safety
 > step, and packaging detail. Come back to it after your first skill has landed
 > and you want the complete checklist.
 >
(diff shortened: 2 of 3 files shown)


potiuk (approved):
Approving, and I owe you an explanation for the delay and for the rebase I pushed.

**Your implementation is the better one, and I merged the other one first.** [#1011](https://github.com/apache/magpie/pull/1011) fixed the same bug, was opened after this PR, and I merged it without noticing yours — it had no CI at the time (its workflow runs were sitting unapproved), so it never surfaced in my review queue. That was my miss, not a judgement on this PR.

I have rebased this onto the merged state and kept **your** parser. The two implementations are not equivalent, and I checked the difference by running both rather than reading them:

| Input | the merged version | yours |
|---|---|---|
| bare four-backtick fence ` ```` ` | left **untagged** — MD040 still fires, which is the entire reason `tag_bare_code_fences` exists | tagged ` ````text `, at its own fence length |
| a `` ```python `` line inside a three-backtick block | treated as a close, so the *real* closing fence then looked like an opening one and got tagged — **corrupting the output** | correctly kept the block open |

Two things your version gets right that the other did not:

- **Requiring the closing fence to carry no info string.** That is the CommonMark rule, and it is what prevents the corruption in the second row above.
- **Preserving the opening fence's length when tagging** — `f"{indent}{fence}{default_lang}"` rather than a hardcoded three backticks — which is what makes the first row work.

Using a single anchored regex with named groups for all of it reads better than counting backticks by hand, too.

### What I changed while rebasing

- **Kept your `tag_bare_code_fences` wholesale**, discarding the merged implementation.
- **Merged the two test suites** rather than replacing one with the other: the six tests from #1011 plus your end-to-end nested-fence case, adapted to the module binding the file already uses.
- **Corrected one test that had pinned the wrong behaviour.** `test_shorter_fence_inside_block_is_literal` asserted that a bare four-backtick fence stays untagged. That encoded the merged version's limitation as if it were intended. The function's own docstring says it exists to give *every* language-less opening fence a language so MD040 passes, so a bare ` ```` ` must be tagged. It now asserts `"````text\n```\n````"`.
- **Added two regression tests** for the behaviours your version fixes and the other did not — the bare longer opening fence, and the info-string line that must not close a block. Both fail against the merged implementation.

Nine tests, all passing, and they now run in CI: `ai-tutors/` became a uv workspace member as part of #1011, so there is a real `pytest (ai-tutors)` job.

### One thing deliberately left out

`inject-knowledge-base.py --check` reports `lesson-08-eval-driven-development.md` as needing a refresh. I checked whether the parser change caused that, and it did not — both `lesson-04` and `lesson-08` were already stale on `main` before this PR, from source-page edits unrelated to code fences. Your `lesson-04` regeneration stays, since it demonstrates this fix; `lesson-08` wants its own regeneration commit and should not ride along here.

Worth noting for a maintainer: [#928](https://github.com/apache/magpie/pull/928) regenerates `lesson-04` only, so it now overlaps with this PR and probably wants closing or rebasing once this lands.

Thanks for this, and sorry for the ordering — the stricter parser is the one worth having.

---

> will follow up.*
>

pollychen-lab:
@potiuk thanks for the detailed review and for preserving the stronger implementation here.

PR #1011, opened 2026-07-29
fix(ai-tutors): stop nested code fences inverting the fence state

### LINKED ISSUE #1008: Fix the nested-code-fence bug in inject-knowledge-base.py

wangqiao258:
## Summary
Fix the nested-code-fence bug in `tag_bare_code_fences` (`ai-tutors/inject-knowledge-base.py`). The function toggled `in_code` on any line starting with three backticks, so a four-backtick block wrapping a bare three-backtick fence flipped the state and never recovered: the inner fence was treated as a close, the real close was treated as an open, and later bare fences stopped getting their default language. The fix tracks the opening fence's backtick length and only closes on a fence at least as long, so a shorter run inside the block is literal content. Adds `ai-tutors/test_inject_knowledge_base.py` covering the nested case plus the existing behaviour.

## Type of change
- [ ] Skill change (`.claude/skills/...`) — eval fixtures updated below
- [ ] Tool / bridge contract (`tools/.../*.md`)
- [ ] Groovy reference impl
- [x] Other: `ai-tutors/inject-knowledge-base.py` knowledge-base injector script (not under `tools/`)

## Test plan
- [x] Other: `python -m pytest ai-tutors/test_inject_knowledge_base.py` — 6/6 pass (incl. nested-case regression); `python ai-tutors/inject-knowledge-base.py --check` reports no new staleness (the one pre-existing stale tutor `lesson-04-your-first-skill.md` is identical before and after this fix — verified by `git stash` on a clean clone).

## RFC-AI-0004 compliance
N/A — pure string-processing bug fix to a markdown helper; no agentic mutation, network, sandbox, or LLM surface introduced.
- [ ] HITL
- [ ] Sandbox
- [ ] Vendor neutrality
- [ ] Conversational + correctable
- [ ] Write-access discipline
- [ ] Privacy LLM

## Linked issues
Closes #1008

- Behaviour-preserving for all currently-injected source pages (none use four-backtick fences today; the issue notes this is latent). Verified by `--check` returning the same single pre-existing stale tutor before and after the change.
- The pre-existing staleness on `lesson-04-your-first-skill.md` is unrelated to this fix and left untouched (out of scope).

### DIFF
--- a/ai-tutors/inject-knowledge-base.py
+++ b/ai-tutors/inject-knowledge-base.py
@@ -130,16 +130,23 @@ def tag_bare_code_fences(text: str, default_lang: str = "text") -> str:
     """
     lines = text.split("\n")
     in_code = False
+    fence_len = 0  # backticks in the opening fence of the current block
     for i, line in enumerate(lines):
         stripped = line.strip()
-        if stripped.startswith("```"):
-            if not in_code:
-                in_code = True
-                if stripped == "```":  # opening fence with no language
-                    indent = line[: len(line) - len(line.lstrip())]
-                    lines[i] = f"{indent}```{default_lang}"
-            else:
-                in_code = False
+        if not stripped.startswith("```"):
+            continue
+        backticks = len(stripped) - len(stripped.lstrip("```"))
+        if not in_code:
+            in_code = True
+            fence_len = backticks
+            if stripped == "```":  # opening fence with no language
+                indent = line[: len(line) - len(line.lstrip())]
+                lines[i] = f"{indent}```{default_lang}"
+        elif backticks >= fence_len:
+            # a closing fence must be at least as long as the opening fence
+            in_code = False
+            fence_len = 0
+        # a shorter backtick run inside the block is literal content; leave as-is
     return "\n".join(lines)
 
 
--- a/ai-tutors/inject-knowledge-base.py
+++ b/ai-tutors/inject-knowledge-base.py
@@ -36,7 +36,6 @@ import sys
 from dataclasses import dataclass
 from pathlib import Path
 
-
 ROOT = Path(__file__).resolve().parents[1]
 AI_TUTORS_DIR = ROOT / "ai-tutors"
 TRAINING_DIR = ROOT / "docs" / "education" / "training"
@@ -135,7 +134,7 @@ def tag_bare_code_fences(text: str, default_lang: str = "text") -> str:
         stripped = line.strip()
         if not stripped.startswith("```"):
             continue
-        backticks = len(stripped) - len(stripped.lstrip("```"))
+        backticks = len(stripped) - len(stripped.lstrip("`"))
         if not in_code:
             in_code = True
             fence_len = backticks
@@ -195,9 +194,7 @@ def refresh_tutor(tutor: Path) -> TutorUpdate:
 
     training_lesson = TRAINING_DIR / tutor.name
     if not training_lesson.is_file():
-        raise FileNotFoundError(
-            f"matching training lesson does not exist: {training_lesson}"
-        )
+        raise FileNotFoundError(f"matching training lesson does not exist: {training_lesson}")
 
     source_page = source_page_for(training_lesson)
     if not source_page.is_file():
--- a/ai-tutors/pyproject.toml
+++ b/ai-tutors/pyproject.toml
@@ -1,0 +1,82 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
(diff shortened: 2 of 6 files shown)


potiuk (approved):
Approving with one commit added on top (`f1a45c3`) that makes your test actually run in CI. Details below — please sanity-check it rather than take it on trust, since it touches the workspace wiring rather than your fix.

### The fix is right, and the test is a real regression test

I checked by mutation rather than by reading. Restoring `main`'s version of the injector while keeping your test file makes exactly one test fail:

```text
FAILED test_inject_knowledge_base.py::test_nested_four_backtick_block_does_not_invert_state
```

…and it passes with your fix in place. Tracking `fence_len` and requiring `backticks >= fence_len` before treating a fence as a close is the right rule: a shorter run inside a longer block is content, not a delimiter.

### Why the extra commit

The test would never have run. `ai-tutors/` was not a uv workspace member and appeared in no workflow or prek hook, so `test_inject_knowledge_base.py` was the only test file in the repo outside `tools/` and nothing collected it. The green checks on this PR did not cover it — revert the fix tomorrow and CI would still be green. A regression test that does not run is not a regression test.

`f1a45c3` makes `ai-tutors/` a workspace member so the existing machinery picks it up: `tests.yml` derives its pytest matrix from `[tool.uv.workspace] members` and emits a job for any member with a `[tool.pytest.ini_options]` section, so this adds a `pytest (ai-tutors)` job with no workflow edit. The member is `package = false`, since the injector is a standalone script rather than an installable package.

### Two things that surfaced the moment the checks applied

Both are in the new code, and both are the reason wiring this up was worth doing rather than deferring:

- **`lstrip("```")` passes a character set, not a prefix.** It happens to work only because that set is a single repeated backtick — `"```x```".lstrip("```")` would strip from both notions of "backtick", and the intent reads as "strip this prefix", which is not what the call does. Ruff's `B005` flags it. Now `lstrip("`")`, which is what was meant, with identical behaviour here.
- **The `importlib` shim was not type-safe.** `spec_from_file_location` returns `ModuleSpec | None` and `.loader` is `Loader | None`; mypy rejected indexing into both. Narrowed with an `assert` that doubles as a clear failure if the injector ever stops being loadable.

One incidental fix was needed to allow any of this: `check-workspace-members` discovered members by scanning `tools/` only, so a legitimate member outside that tree was reported as *stale* with an instruction to delete the correct entry. It now also scans top-level project directories. `ai-tutors/` is the only one today.

Full `prek run --all-files` is green, including the new member's ruff, mypy, and pytest.

Thanks for this — the fix was already correct, and the parser rule you landed on is the one CommonMark uses.

---

> will follow up.*
>

Issue #888, opened 2026-07-15
docs/rfcs/RFC-AI-0004.md: status says "Draft", text says implementation operationalises every principle


rbowen:
### What should happen

1. Update the RFC status from "Draft" to "Accepted" (or "Implemented").
2. Fix the inconsistency in the Motivation section that says "four principles" in one place and "six principles" in another.

### Why

The RFC's own text states the reference implementation already operationalises every principle — the "Draft" status contradicts this. The four/six inconsistency is confusing.

### Which layer

Cross-cutting (RFC, AGENTS.md, sandbox, privacy-LLM)

### References

Note: RFC status changes may require a mailing-list thread per the amendment process in PRINCIPLES.md.

potiuk:
Fixed in #1159 (3d9b2c98). Status is now `Implemented`.

Deliberately not `Stable`: the RFC's own promotion gate calls for public review
and pilot-cohort validation, and neither has happened — the `Discussion` row is
still unfiled. The new status section records what remains open so the flip
does not contradict the Drawbacks and Open-questions sections.

The "four shifts" line was the only "four" in a document that says six
everywhere else, but the list beneath it genuinely had four items. Rather than
flip the numeral over a four-item list, the two missing gap bullets were added
(Principle 5, outbound messages in the maintainer's voice; Principle 6, privacy
as a deployment detail).

If the project would rather an RFC status change go through a dev@ thread
first per the PRINCIPLES.md amendment process, happy to revert this and take
it there.

PR #1039, opened 2026-07-29
fix: correct principle count in RFC-AI-0004 from four to six

shenxianpeng:
## Summary

Fix a stale-text bug in `docs/rfcs/RFC-AI-0004.md`: the Motivation section's pilot-validation milestone said "four principles" but there are six.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

## RFC-AI-0004 compliance

## Linked issues

Refs #888

### DIFF
--- a/docs/rfcs/RFC-AI-0004.md
+++ b/docs/rfcs/RFC-AI-0004.md
@@ -71,7 +71,7 @@
 |---|---|
 | **RFC** | AI-0004 |
 | **Title** | Principles of agentic interaction for open-source maintainers |
-| **Status** | Draft |
+| **Status** | Accepted |
 | **Authors** | The Apache Magpie project (see [`MISSION.md`](https://github.com/apache/magpie/blob/main/MISSION.md) for roster) |
 | **Initial draft** | 2026-05-07 |
 | **Supersedes** | None |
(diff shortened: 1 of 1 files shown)


potiuk (changes requested):
Two independent changes are bundled here, and they need different treatment: the `four principles` → `six principles` correction is right and I would merge it today. The `Draft` → `Accepted` status flip is a governance change that needs a recorded project decision behind it, and I can't merge the two together.

### Status change needs a recorded decision (`docs/rfcs/RFC-AI-0004.md:74`, `:96`)

Three things point the same way:

1. **The RFC's own lifecycle doesn't have an `Accepted` state.** The line this PR rewrites said promotion runs *"before promotion from `Draft` to `Stable`"* — `Draft` → `Stable`, with no intermediate step. `Accepted` appears nowhere in `docs/rfcs/`, so this introduces a new status value into the vocabulary as a side effect of flipping one document.
2. **No decision is on record.** The project socialises RFCs on `dev@magpie.apache.org` with a `[PROPOSE]` thread — that is what happened for RFC-AI-0006. There is no `[PROPOSE]` thread, and no acceptance thread, for RFC-AI-0004; the list archive has no mention of it at all.
3. **The RFC's own prerequisites are still open.** This PR keeps both milestones — public review and pilot validation — and simply relabels them as steps toward `Stable`. By the document's own terms neither has completed, which is the argument for it still being a `Draft`.

RFC-AI-0004 is the document the PR template calls the framework's constitution, so its status line carries real weight; a docs PR is not the right instrument for moving it.

**Suggested split.** Drop the two status hunks from this PR and keep the principle-count fix — I will merge that immediately. If you would like to see the RFC promoted, the path is a `[PROPOSE]` thread on `dev@magpie.apache.org` (RFC-AI-0006's thread is a good template); once consensus is recorded, the status change lands as its own small PR citing that thread, and it should also define `Accepted` in the RFC lifecycle rather than leaving it implicit.

For the status row, reverting to:

```suggestion
| **Status** | Draft |
```

### The principle-count fix is correct

Verified — this one is a genuine stale-text bug and a good catch. `RFC-AI-0004.md` has six `## Principle N` headings and already says "six principles" in three other places (the TOC entry, the abstract at line 86, and the *How the six principles compose* section). Line 99 was the only survivor of the four-principle era.

---

> confirmed by an Apache Magpie maintainer. After you've
> addressed the points above and pushed an update, an Apache Magpie
> maintainer — a real person — will take the next look
> at the PR. The findings cite the project's review criteria;
> if you think one of them is mis-applied, please reply on the
> PR and a maintainer will weigh in.*
>

shenxianpeng:
Thanks for the review @potiuk. I've reverted the status change back to Draft and kept only the four → six fix. Ready for another look.

potiuk (approved):

PR #1043, opened 2026-07-30
docs: document assertions.json requirement in Pattern 8

### LINKED ISSUE #1004: Document assertions.json in the pattern catalogue's eval-suite pattern
## Summary

Pattern 8 in `docs/education/pattern-catalogue.md` shows the eval-suite layout but never
mentions `assertions.json`. A reader who builds the injection case it prescribes gets a suite
that is never graded automatically. Add the missing file and the rule for when it is needed.

## Background

Pattern 8 tells the reader to build a four-case minimum suite: normal, injection, empty queue,
confirm gate. For the injection case it describes the expected output as "injection flagged, not
followed".

The natural way to express that is a flag like `has_injection_flagged` — and here is the trap:
the runner treats any expected key beginning `has_` or `mention_` as a *structural* key, which
requires a matching predicate in an `assertions.json` file in the fixtures directory. When that
file is absent the runner falls back to manual review for the whole case rather than grading it.
The layout in Pattern 8 does not include `assertions.json`, so a reader following it exactly
produces a suite that quietly never runs.

This is not hypothetical. Two suites shipped in exactly that state — the
`non-asf-profile-smoke` reviewer-routing suite and the `good-first-issue-sweep` step-3 suite —
and both sat hand-graded until later PRs added the missing file.

## Where to look

- `docs/education/pattern-catalogue.md` — Pattern 8, "Test your skill with an eval before
  shipping it"; the `text` block showing the directory layout and the prose beneath it.
- `tools/skill-evals/evals/issue-triage/step-3-classify/fixtures/assertions.json` — a small,
  real example: one flag mapped to a regex over another field.
- `tools/skill-evals/README.md` — the assertion types and how structural keys are graded.
- `docs/education/eval-driven-development.md` — already explains `assertions.json` at more
  length; link to it rather than repeating the detail.

## Acceptance criteria

- [ ] The layout block shows `assertions.json` in the fixtures directory.
- [ ] The prose states the rule: an expected key starting with `has_` or `mention_` needs a
      matching entry in `assertions.json`, or the case is not graded automatically.
- [ ] The four-case minimum guidance is unchanged.

## Estimated effort

~45 minutes for someone new to the codebase.

## Getting started

deepusnath:
Closes #1004

## Summary

- Add `assertions.json` to the fixtures directory-layout block in Pattern 8,
  right after `user-prompt-template.md`.
- Add a paragraph after the four-case list explaining that `expected.json`
  keys starting with `has_` or `mention_` are structural flags that need a
  matching predicate in `assertions.json` — without it the case falls back
  to `MANUAL` grading instead of `PASS`/`FAIL` — and linking to
  `eval-driven-development.md` for the format.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

## RFC-AI-0004 compliance

## Linked issues

Closes #1004

None.

### DIFF
--- a/docs/education/pattern-catalogue.md
+++ b/docs/education/pattern-catalogue.md
@@ -361,6 +361,7 @@ tools/skill-evals/evals/<skill-name>/
         ├── step-config.json          # { "skill_md": "...", "step_heading": "..." }
         ├── output-spec.md            # JSON schema the step must return
         ├── user-prompt-template.md   # user-facing prompt with {variable} slots
+        ├── assertions.json           # checks for any has_*/mention_* keys below
         ├── case-1-normal/
         │   ├── report.md             # realistic example input
         │   └── expected.json         # expected structured output
@@ -400,6 +401,16 @@ booleans, IDs) are compared exactly; prose fields are scored by a judge model.
    with a "nothing to do" message, not an error.
 4. **Confirm gate.** Check that the skill stops and asks before any step that
    changes something. It must not assume it already has permission.
+
+**If `expected.json` uses `has_*` or `mention_*` keys, add `assertions.json`.**
+The runner treats keys starting with `has_` or `mention_` as structural flags
+and looks for a matching predicate in `assertions.json` (see
+[`docs/education/eval-driven-development.md`](eval-driven-development.md) for
+the format). Without it, those cases fall back to `MANUAL` instead of
+`PASS`/`FAIL` — easy to miss since the suite still runs, it just never grades
+itself. The injection case is the one most likely to need this, since
+"injection flagged, not followed" is usually checked with a `has_*` flag on
+the rationale rather than an exact string match.
 
 **Why the injection case matters:** it is the easiest one to forget and the
 most important one to have. Without it, a reviewer cannot check the skill's


potiuk (approved):
LGTM — accurate on every point I checked, and it documents a genuinely nasty failure mode.

Verified against the runner rather than taking the description on trust:

- `has_*` / `mention_*` really are the structural-flag prefixes, and the runner looks for a matching predicate in `assertions.json` — `runner.py`'s module docstring lists the whole predicate set (`regex`, `contains`, `contains_all`, `empty`, `non_empty`, `field_true`, `max_length`, `judge`).
- The consequence you describe is exactly what the code says: *"A structural case with no `assertions.json` falls back to MANUAL and prints prompts for manual review."*
- [`eval-driven-development.md`](eval-driven-development.md) does document the format, with a worked `has_merge_ready` example — so the pointer lands somewhere useful rather than at a doc that merely mentions the file.
- The tree diagram puts `assertions.json` at the `fixtures/` level, next to `step-config.json` and above the `case-*` directories, which matches where the real suites keep it.

The sentence worth the whole PR is this one:

> easy to miss since the suite still runs, it just never grades itself

That is the trap. A missing `assertions.json` does not error, does not fail, and does not skip — the suite reports `MANUAL` and a contributor scanning for red sees none. Writing that down next to the pattern is more useful than another paragraph on the format.

Tying it to the injection case is the right call too, since *"injection flagged, not followed"* is precisely the kind of property nobody wants to assert with an exact string match.

Thanks for the first contribution to Apache Magpie.

---

> will follow up.*
>

Issue #1002, opened 2026-07-29
Correct the AGENTS.md claim that all system tools are pinned with a 7-day cooldown


justinmclean:
## Summary

`AGENTS.md` still summarises the framework's supply-chain posture as "system tools pinned with
a 7-day upstream cooldown". That stopped being true for the agent runtime when PR #825 moved
`claude-code` to an always-latest install. Update the sentence.

## Background

PR #825 deliberately unpinned the agent runtime: `claude-code` is now installed at `@latest`
with a `min_version` floor and **no** cooldown, because each release carries the newest
permission-rule, sandbox and prompt-injection fixes, so pinning it would increase the security
lag. The sandbox primitives (`bubblewrap`, `socat`) keep their exact pins and the 7-day
default.

That change was propagated carefully — `docs/prerequisites.md`,
`docs/setup/secure-agent-setup.md`, `docs/setup/secure-agent-internals.md`,
`docs/rfcs/RFC-AI-0002.md`, the `agent-isolation` README, three setup skills and 16 eval
fixtures all moved. `AGENTS.md` did not.

It matters because `AGENTS.md` is the file every agent reads first, so the summary an agent
forms of the framework's own supply-chain posture is the pre-#825 one. It is also a security
claim, which is the kind of sentence worth being exactly right.

## Where to look

- `AGENTS.md` around line 431 — the sentence inside the *Local setup* section, in the paragraph
  beginning "Run the agent in the credential-isolation setup".
- `docs/setup/secure-agent-setup.md` § *Required tools* — the accurate current description and
  the wording to align with. Note its table row: `claude-code | (unpinned — @latest) | — | none`.
- `tools/agent-isolation/pinned-versions.toml` — the header comment explains the reasoning; the
  `[tools.claude-code]` table holds the `min_version` floor.

## Acceptance criteria

- [ ] The `AGENTS.md` sentence no longer implies the agent runtime is pinned or subject to the
      cooldown.
- [ ] It still conveys that the sandbox primitives *are* pinned with a cooldown, since that
      remains true.
- [ ] No other file is changed.
- [ ] `prek run --all-files` passes.

## Estimated effort

~30 minutes for someone new to the codebase; most of it is reading the two source docs closely
enough to summarise them accurately in one sentence.

## Getting started

[...]

PR #1045, opened 2026-07-31
docs: clarify tool update policy

tomatotomata:
## Summary\n\n- clarify that only the sandbox primitives use the seven-day cooldown\n- document that the agent runtime tracks the latest release behind a minimum-version floor\n\n## Validation\n\n- git diff --check\n- prek run --all-files *(not run locally because prek and uv are unavailable)*\n\nFixes #1002

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -428,8 +428,9 @@ with default-permissive access to `~/`, env vars, and arbitrary network
 egress is a real exfiltration risk. See
 [`docs/setup/secure-agent-setup.md`](docs/setup/secure-agent-setup.md)
 for the layered defence the framework dogfoods (sandbox + tool
-permissions + clean-env wrapper, system tools pinned with a 7-day
-upstream cooldown).
+permissions + clean-env wrapper, sandbox primitives pinned with a
+7-day upstream cooldown, and an agent runtime installed at the latest
+release subject to a minimum-version floor).
 
 **Tool credentials live under `$HOME`, never in the project tree.** Any
 persistent token, API key, OAuth refresh token, or session cookie a


potiuk (approved):
LGTM — the sentence you replaced was not merely vague, it was wrong, and in the direction that matters for a security doc.

The old text said *"system tools pinned with a 7-day upstream cooldown"*, which reads as though everything the secure setup installs is pinned. [`docs/setup/secure-agent-setup.md`](https://github.com/apache/magpie/blob/main/docs/setup/secure-agent-setup.md) — the very file that sentence links to — says the opposite for one of the two categories:

> The **sandbox primitives** (`bubblewrap`, `socat`) are pinned with a **per-tool cooldown** … Default cooldown is 7 days
>
> The **agent runtime** (`claude-code`) is deliberately **not** pinned. The secure setup installs `@anthropic-ai/claude-code@latest` … pinning the runtime to an older build would *increase* the framework's security lag, not reduce it.

So the summary flattened a deliberate two-policy design into one, and asserted pinning for the component the project has consciously chosen not to pin. Your split restores both halves.

I also checked the second half of the new wording, since "subject to a minimum-version floor" is a specific claim: [`tools/agent-isolation/pinned-versions.toml`](https://github.com/apache/magpie/blob/main/tools/agent-isolation/pinned-versions.toml) carries `min_version = "2.1.202"` for `claude-code`, with the comment *"NOT an exact-version pin — `min_version` is a hard floor, not a pin"*. Accurate.

### Smaller observation

Not for this PR, and arguably a point in its favour: the new AGENTS.md sentence is now **more complete than the document it points at**. `secure-agent-setup.md`'s prose and its pins table describe the runtime as *"unpinned — `@latest`"* with *"Cooldown: none"* and never mention the `min_version` floor — that detail lives only in the TOML. Worth a follow-up to add it there, so a reader following the link does not conclude the floor was invented here.

Thanks for this one — correcting a security-relevant summary that had drifted from its own source is a good first contribution.

---

> will follow up.*
>

PR #1046, opened 2026-07-31
test(validator): add negative controls to TestOrganizationNonASFSmoke for invalid orgs and ASF coupling

### LINKED ISSUE #1009: Add negative controls to the non-ASF organization smoke tests
## Summary

Every assertion in `TestOrganizationNonASFSmoke` checks that a list of violations is empty. That
means the tests cannot tell "the checks ran and correctly found nothing" from "the checks did not
run at all". Add one case per surface where a violation *is* expected.

## Background

The class covers three workflow surfaces for a non-ASF project — security intake, release
housekeeping, contributor governance — and asserts for each that `organization: independent` is
accepted and that the fixture body produces no ASF-coupling warnings. All seven assertions are of
the form `assert x == []`.

These tests do have value: the realistic regression is an ASF-coupling heuristic drifting too
broad and firing on a GHSA body or a GitHub-Releases workflow, and these fixtures would catch
that. What they cannot catch is the check going silent — delete the coupling scan tomorrow and
all four tests still pass.

A negative control fixes that: one fixture per surface that *should* raise a violation, asserting
it does. Then an empty result in the positive cases means something.

## Where to look

- `tools/skill-and-tool-validator/tests/test_validator.py` — the `TestOrganizationNonASFSmoke`
  class; each test already builds its fixture under `tmp_path`, so the scaffolding you need is
  written.
- Other classes in the same file — most pair a "no violations" test with one that asserts a
  specific violation is produced; `TestOrganizationStructure` is a compact example.
- `tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py` — the organization
  membership check and the ASF-coupling scan, to see what input actually triggers each.

## Acceptance criteria

- [ ] At least one new test asserts that an unknown or invalid `organization:` value produces a
      violation.
- [ ] At least one new test asserts that genuinely ASF-coupled content under
      `organization: independent` produces a coupling violation.
- [ ] Each new test names the specific violation it expects rather than asserting a non-empty
      list.
- [ ] The four existing tests are unchanged and still pass.

## Estimated effort

~1-2 hours for someone new to the codebase.

## Getting started

onlyarnav:
### Summary
Adds negative control test cases to `TestOrganizationNonASFSmoke` in `tools/skill-and-tool-validator/tests/test_validator.py` for each workflow surface (organization frontmatter, security intake, release housekeeping, contributor governance, and organization structure).

Previously, assertions in `TestOrganizationNonASFSmoke` only verified that violation lists were empty (`assert x == []`), making it impossible to distinguish between "the checks ran and correctly found no violations" versus "the checks went silent or were not executed".

### Changes Made
Added 5 new negative control test cases to `TestOrganizationNonASFSmoke`:
- `test_invalid_organization_yields_violation`: Asserts an unknown/invalid `organization:` frontmatter value produces an `ORGANIZATION_CATEGORY` violation (`"is not a known organization"`).
- `test_security_intake_asf_coupled_body_yields_violation`: Asserts security intake skills containing ASF-coupled email patterns (`announce@apache.org`) under `organization: independent` produce an `ASF_COUPLING_CATEGORY` violation (`"asf-coupling"`).
- `test_release_backend_asf_coupled_body_yields_violation`: Asserts release housekeeping skills containing `svn` release commands under `organization: independent` produce an `ASF_COUPLING_CATEGORY` violation (`"asf-coupling"` and `"svn"`).
- `test_contributor_governance_asf_coupled_body_yields_violation`: Asserts contributor governance skills mentioning `PMC` votes under `organization: independent` produce an `ASF_COUPLING_CATEGORY` violation (`"asf-coupling"` and `"PMC"`).
- `test_independent_org_structure_missing_file_yields_violation`: Asserts incomplete organization adapters missing `organization.md` produce an `ORGANIZATION_CATEGORY` violation (`"missing required file 'organization.md'"`).

### Verification
- Executed `uv run pytest tools/skill-and-tool-validator/tests/test_validator.py -k TestOrganizationNonASFSmoke`: all 9 tests pass.
- All four existing positive control tests remain untouched and pass.

## fixes: #1009 

Generated using Claude Code

### DIFF
--- a/tools/bitbucket/tests/test_bitbucket.py
+++ b/tools/bitbucket/tests/test_bitbucket.py
@@ -904,6 +904,28 @@ def test_cloud_get_pull_request_commits_follows_next(
 
 
 @patch("urllib.request.build_opener")
+def test_cloud_get_pull_request_commits_rejects_repeated_next_url(
+    mock_build_opener: MagicMock,
+    cloud_env: None,
+) -> None:
+    repeated_url = "https://api.bitbucket.org/2.0/repositories/apache/magpie/pullrequests/7/commits?page=2"
+    mock_opener(
+        mock_build_opener,
+        {
+            "values": [],
+            "next": repeated_url,
+        },
+        {
+            "values": [],
+            "next": repeated_url,
+        },
+    )
+
+    with pytest.raises(BitbucketError, match="pagination returned a repeated URL"):
+        cloud.get_pull_request_commits(load_config(), "7")
+
+
+@patch("urllib.request.build_opener")
 def test_datacenter_get_pull_request_commits_follows_next_page_start(
     mock_build_opener: MagicMock,
     datacenter_env: None,
--- a/tools/skill-and-tool-validator/tests/test_validator.py
+++ b/tools/skill-and-tool-validator/tests/test_validator.py
@@ -3264,7 +3264,7 @@ class TestOrganizationNonASFSmoke:
         assert len(coupling_violations) == 1
         assert coupling_violations[0].category == ASF_COUPLING_CATEGORY
         assert "asf-coupling" in coupling_violations[0].message
-        assert "apache.org" in coupling_violations[0].message
+        assert "announce@apache.org" in coupling_violations[0].message
 
     def test_release_backend_asf_coupled_body_yields_violation(self, tmp_path: Path) -> None:
         """Release housekeeping skill containing SVN release commands triggers coupling violation."""
@@ -3317,11 +3317,6 @@ class TestOrganizationNonASFSmoke:
         assert len(vs) == 1
         assert vs[0].category == ORGANIZATION_CATEGORY
         assert "missing required file 'organization.md'" in vs[0].message
-
-
-
-
-
 
 
 # ---------------------------------------------------------------------------
--- a/tools/skill-and-tool-validator/tests/test_validator.py
+++ b/tools/skill-and-tool-validator/tests/test_validator.py
@@ -3266,6 +3266,28 @@ class TestOrganizationNonASFSmoke:
         assert "asf-coupling" in coupling_violations[0].message
         assert "apache.org" in coupling_violations[0].message
 
+    def test_release_backend_asf_coupled_body_yields_violation(self, tmp_path: Path) -> None:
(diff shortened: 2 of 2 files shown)


potiuk (approved):
LGTM — the premise is exactly right, and I confirmed it by experiment rather than by reading, because "these assertions cannot fail" is a claim worth testing.

### The premise, demonstrated

`TestOrganizationNonASFSmoke` asserted `x == []` throughout. That passes in two very different worlds: the validator ran and correctly found nothing, or the validator never ran at all. I disabled `validate_asf_coupling` (early `return []`) and re-ran both sets against the break:

| | Result |
|---|---|
| the four pre-existing `assert x == []` assertions | **all passed** — completely blind to it |
| your three coupling negative controls | **all failed** |

So the suite was reporting green on a validator that had stopped working. That is the whole argument for this PR, and it holds.

The assertions are also the right shape. Each pins a count, a category, *and* message content:

```python
assert len(coupling_violations) == 1
assert coupling_violations[0].category == ASF_COUPLING_CATEGORY
assert "announce@apache.org" in coupling_violations[0].message
```

Asserting only "non-empty" would have been the easy version and would fail to catch a validator that fires for the wrong reason. Naming the offending token in the message assertion means the test also documents what the rule is for.

This is the same class of gap as [#984](https://github.com/apache/magpie/pull/984) — where direct unit tests stayed green if the call was deleted from `run_validation` — and it is the reason I went looking for a mutation here at all.

### One thing worth knowing, no action needed

The `tools/bitbucket/tests/test_bitbucket.py` half of this PR adds `test_cloud_get_pull_request_commits_rejects_repeated_next_url`, which is byte-identical to the test [#1047](https://github.com/apache/magpie/pull/1047) landed while this was open.

I checked what that actually does on merge rather than assuming a conflict or a duplicate definition: computing the merge result with `git merge-tree` against current `main` gives **one** copy, not two — git recognises the identical addition and collapses it. So there is no `F811` redefinition and nothing to fix. The net effect of merging this PR today is exactly the 95 lines of `test_validator.py`, and the bitbucket hunk is simply a no-op.

Flagging it only so the diff not matching what lands is not a surprise later.

Verified: 11 tests pass under `-k yields_violation`, and the suite is green with the change applied.

---

> will follow up.*
>

PR #1047, opened 2026-07-31
test(bitbucket): add test coverage for defensive pagination and redirect guards

### LINKED ISSUE #1007: Add tests for the Bitbucket pagination guards and the redirect scheme-downgrade case
## Summary

The Bitbucket bridge has several defensive guards that exist because earlier reviews asked for
them, and none of them has a test. Add cases so a later simplification cannot remove them
silently.

## Background

Three guards are currently unprotected:

**1. Non-advancing pagination cursor (Data Center).** `get_pull_request_commits` breaks out of
its `while True` loop when `nextPageStart` is missing, not an integer, or not greater than the
current `start`. That last check exists because a review on #766 pointed out the loop would hang
forever against a server that keeps returning the same cursor. The tests cover only the happy
path: `nextPageStart: 25` followed by `isLastPage: True`.

**2. Repeated pagination URL (Cloud).** The Cloud fetchers thread a `seen_urls` set through
`_validated_next_url` so a `next` link pointing back at an already-fetched URL is rejected.
No test exercises that rejection.

**3. Redirect scheme downgrade.** `SameHostRedirectHandler` compares scheme, host and effective
port before following a redirect, so an `https` → `http` redirect on the same host is refused
and the `Authorization` header is never sent in clear text. Three tests exist for that handler
— same-origin accepted, different host rejected, different port rejected — but not the downgrade,
which is the security-relevant one.

Each guard is a couple of lines of code and each test is a couple of lines of fixture. The value
is that the next person to tidy these loops finds out immediately.

## Where to look

- `tools/bitbucket/tests/test_bitbucket.py` — all the tests live here.
  `test_datacenter_get_pull_request_commits_follows_next_page_start` and
  `test_same_host_redirect_handler_rejects_different_port` are the closest models to copy.
- `tools/bitbucket/src/magpie_bitbucket/datacenter.py` — `get_pull_request_commits`, for the
  three loop exits.
- `tools/bitbucket/src/magpie_bitbucket/cloud.py` — the `seen_urls` set and
  `_validated_next_url`.
- `tools/bitbucket/src/magpie_bitbucket/client.py` — `SameHostRedirectHandler`, `_origin` and
  `_effective_port`.

## Acceptance criteria

- [ ] A test feeds a Data Center page with `isLastPage: False` and a `nextPageStart` equal to
      or below the current `start`, and asserts the loop terminates rather than hanging.
- [ ] A test asserts a repeated Cloud `next` URL is rejected rather than re-fetched.
- [ ] A test asserts an `https` → `http` redirect on the same host raises rather than being
      followed.
- [ ] Each new test fails if its guard is removed. (Check this by deleting the guard, running the
      test, then restoring it.)
- [ ] The package's existing tests still pass.

## Estimated effort

~2 hours for someone new to the codebase; the third test is the most interesting and the fixtures
for the first two are small.

## Getting started

onlyarnav:
### Summary
Adds unit test cases in `tools/bitbucket/tests/test_bitbucket.py` for three defensive guards in the Bitbucket bridge to prevent regressions or accidental removal during future refactoring:

1. **Redirect Scheme Downgrade Guard**:
   - `test_same_host_redirect_handler_rejects_scheme_downgrade`: Asserts `SameHostRedirectHandler` refuses `https → http` redirects on the same host and raises `BitbucketError` ("refusing to forward credentials").
2. **Data Center Non-advancing Cursor Guard**:
   - `test_datacenter_get_pull_request_commits_stops_on_non_advancing_next_page_start`: Asserts `datacenter.get_pull_request_commits` terminates its pagination loop when `nextPageStart` is non-advancing (`<= start`), preventing infinite loops.
3. **Cloud Repeated Pagination URL Guard**:
   - `test_cloud_get_pull_request_commits_rejects_repeated_next_url`: Asserts `cloud.get_pull_request_commits` rejects repeated next URLs and raises `BitbucketError` ("pagination returned a repeated URL").

### Negative Control Verification
Each test was verified by temporarily disabling its guard in source code:
- Disabling scheme validation in `_origin()` caused `test_same_host_redirect_handler_rejects_scheme_downgrade` to fail (`DID NOT RAISE BitbucketError`).
- Removing `if next_start <= start: break` from `datacenter.py` caused `test_datacenter_get_pull_request_commits_stops_on_non_advancing_next_page_start` to fail (`StopIteration`).
- Removing `if next_url in seen_urls:` from `cloud.py` caused `test_cloud_get_pull_request_commits_rejects_repeated_next_url` to fail (`StopIteration`).

### Test Suite Status
All 101 unit tests in `tools/bitbucket` pass cleanly (`uv run --directory tools/bitbucket pytest`).

## fixes: #1007 

Generated using Claude Code

### DIFF
--- a/tools/bitbucket/tests/test_bitbucket.py
+++ b/tools/bitbucket/tests/test_bitbucket.py
@@ -919,6 +919,28 @@ def test_cloud_get_pull_request_commits_follows_next(
 
 
 @patch("urllib.request.build_opener")
+def test_cloud_get_pull_request_commits_rejects_repeated_next_url(
+    mock_build_opener: MagicMock,
+    cloud_env: None,
+) -> None:
+    repeated_url = "https://api.bitbucket.org/2.0/repositories/apache/magpie/pullrequests/7/commits?page=2"
+    mock_opener(
+        mock_build_opener,
+        {
+            "values": [],
+            "next": repeated_url,
+        },
+        {
+            "values": [],
+            "next": repeated_url,
+        },
+    )
+
+    with pytest.raises(BitbucketError, match="pagination returned a repeated URL"):
+        cloud.get_pull_request_commits(load_config(), "7")
+
+
+@patch("urllib.request.build_opener")
 def test_datacenter_get_pull_request_commits_follows_next_page_start(
     mock_build_opener: MagicMock,
     datacenter_env: None,
--- a/tools/bitbucket/tests/test_bitbucket.py
+++ b/tools/bitbucket/tests/test_bitbucket.py
@@ -149,6 +149,21 @@ def test_same_host_redirect_handler_rejects_different_port() -> None:
         )
 
 
+def test_same_host_redirect_handler_rejects_scheme_downgrade() -> None:
+    handler = SameHostRedirectHandler()
+    request = urllib_request("https://bitbucket.example.test/rest/api/1.0/foo")
+
+    with pytest.raises(BitbucketError, match="refusing to forward credentials"):
+        handler.redirect_request(
+            request,
+            None,
+            302,
+            "Found",
+            {},
+            "http://bitbucket.example.test/rest/api/1.0/bar",
+        )
+
+
 def test_load_config_defaults_to_cloud(monkeypatch: pytest.MonkeyPatch) -> None:
     monkeypatch.delenv("BITBUCKET_KIND", raising=False)
     config = load_config()
--- a/tools/bitbucket/tests/test_bitbucket.py
+++ b/tools/bitbucket/tests/test_bitbucket.py
@@ -945,6 +945,27 @@ def test_datacenter_get_pull_request_commits_follows_next_page_start(
     assert [item["id"] for item in result["values"]] == ["abc123", "def456"]
 
 
+@patch("urllib.request.build_opener")
+def test_datacenter_get_pull_request_commits_stops_on_non_advancing_next_page_start(
+    mock_build_opener: MagicMock,
+    datacenter_env: None,
+) -> None:
+    opener = mock_opener(
+        mock_build_opener,
+        {
+            "values": [{"id": "abc123", "message": "First commit"}],
+            "isLastPage": False,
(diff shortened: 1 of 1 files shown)


potiuk (approved):
LGTM — three tests for three real guards, and they bite. Tests for defensive paths are easy to write in a way that passes whether or not the defence exists, so I checked by deleting each guard rather than by reading.

**Scheme downgrade.** `_origin()` returns `(scheme, hostname, port)`, so an `https` → `http` redirect genuinely produces a different origin and trips `SameHostRedirectHandler`. The test pins the credential-forwarding refusal on the axis that matters most: a downgrade to plaintext is exactly when you least want the `Authorization` header to follow.

**Non-advancing `nextPageStart`.** Removing `if next_start <= start: break` from `get_pull_request_commits` makes your test fail — it blows up on `StopIteration` as the loop keeps asking for pages past the single mocked response, which is the infinite-loop behaviour the guard exists to prevent. `assert len(opener.open.call_args_list) == 1` is what makes it precise: it pins *one* request, not merely a successful return.

**Repeated `next` URL.** Correctly targeted at the cloud paginator's own error.

### Smaller observation — worth a follow-up, not a change here

While confirming the above I removed the *same* guard from `get_repository_restrictions` instead, and the **entire 101-test suite still passed**. So that sibling guard is completely unpinned.

`datacenter.py` repeats the `next_start <= start` pattern in six functions:

```text
get_repository_restrictions   list_open_pull_requests    get_pull_request_commits
get_pull_request_status       get_pull_request_reviews   get_pull_request_discussion
```

This PR pins one of the six. `cloud.py` does not have that problem because its equivalent lives in a single `_validated_next_url()` helper, so one test covers every caller.

That contrast is the interesting part, and it points at a better follow-up than "write five more tests": hoisting the datacenter guard into a shared helper the way cloud already does would make one test cover all six, and would remove the possibility of the next paginator being added without it. Either way it is out of scope here — this PR strictly improves matters.

---

> will follow up.*
>

PR #1049, opened 2026-07-31
fix(pr-management-triage): centralize terminal PR links

AyushS1304:
## Summary

- Add one formatter for short and full GitHub pull-request references, resolving each to a canonical `/pull/<number>` URL.
- Emit OSC 8 terminal hyperlinks when supported, with readable URL fallbacks for `NO_COLOR`, `TERM=dumb`, and non-terminal output.
- Apply the renderer consistently across triage progress, proposals, grouped results, drill-down views, errors, and summaries, with regression fixtures for each fallback path.

## Type of change

- [x] Skill change (`.claude/skills/<name>/`) — eval fixtures updated below
- [ ] Tool / bridge contract (`tools/<system>/*.md`)
- [ ] Python package (`tools/*/` with `pyproject.toml`)
- [ ] Groovy reference impl
- [ ] Cross-cutting (RFC, AGENTS.md, sandbox, privacy-LLM)
- [x] Documentation (`docs/`, `README.md`, `CONTRIBUTING.md`)
- [ ] Project template (`projects/_template/`)
- [ ] CI / dev loop (`prek`, workflows, validators)
- [x] Other: triage CLI helper and unit tests

## Test plan

- [x] `prek run --all-files` passes
- [x] For Python packages touched: `uv run pytest` / `ruff check` / `mypy` passes
- [ ] For Groovy bridges touched: command-line invocation tested end-to-end
- [x] For skill changes: eval suite passes for the affected skill
- [x] For skill *behaviour* changes: a new or updated eval fixture is included in this PR
- [x] Other: 9 helper unit tests and all 150 skill-eval framework tests pass

## RFC-AI-0004 compliance

- [ ] **HITL** — any new mutation is gated on explicit user confirmation
- [ ] **Sandbox** — no new unrestricted host access; network reach declared in the adapter
- [x] **Vendor neutrality** — placeholders used in skill and tool prose
- [ ] **Conversational + correctable** — agentic-override path documented if behaviour is adopter-tunable
- [ ] **Write-access discipline** — no autonomous outbound messages; drafts only, sent on confirmation
- [ ] **Privacy LLM** — private content does not reach a non-approved LLM; redactor invoked where needed

# Ai Assistance
took help from Codex for testing.

## Linked issues

Closes #73

## Notes for reviewers

The helper deliberately keeps the visible reference compact in capable terminals while ensuring the canonical URL remains visible everywhere else.

PR #1054, opened 2026-08-01
ci(skills): lint, type-check and test the Python under skills/

### LINKED ISSUE #1053: ci(skills): Python under skills/ is neither linted, type-checked, nor tested
## What

Python under `skills/` is not linted, type-checked, or tested by CI. The workspace checks that run `ruff`, `mypy`, and `pytest` iterate over uv workspace members and are filed under

```text
^(tools/[^/]+(/[^/]+)?/(src|tests|pyproject\.toml)|pyproject\.toml|tools/dev/run-workspace-check\.sh)
```

`skills/**` matches none of that. There are eight Python files under `skills/` today — helper scripts and guards:

```text
skills/ci-runner-audit/scripts/scan_ci_runners.py
skills/list-skills/scripts/list_skills.py
skills/pr-management-triage/guards/mark_ready.py
skills/pr-management-triage/guards/mention.py
skills/pr-management-triage/scripts/pr_link.py
skills/security-issue-fix/guards/security_language.py
skills/setup-status/scripts/collect_status.py
skills/write-skill/scripts/init_skill.py
```

`check-placeholders` covers them (`^(skills/.*|tools/.*)\.(md|sh|py|yaml|yml|toml)$`), but that is a placeholder check, not lint, types, or tests.

## Why it matters

#1049 added `skills/pr-management-triage/tests/test_pr_link.py` — the **first test file anywhere under `skills/`** — alongside a 133-line `pr_link.py`. The tests pass locally (9 of them) but nothing in CI collects them. Reverting `pr_link.py` would leave the checks green.

Running `ruff` over those files by hand immediately finds a real, if trivial, hit: `RUF100 Unused 'noqa' directive (unused: E402)` at `tests/test_pr_link.py:26`. Exactly the class of thing the repo lints for everywhere else.

Two of the eight files are `guards/` — `mark_ready.py`, `mention.py`, `security_language.py` — which exist specifically to stop an agent doing something unsafe. Uncovered guard code is the least comfortable place for this gap to sit.

This is the third instance of the same shape recently, and worth naming as a pattern rather than three incidents:

- #1011 shipped a regression test nothing collected — fixed by making `ai-tutors/` a uv workspace member.
- #982 / #1043 dealt with eval cases silently falling back to `MANUAL` instead of grading.
- #1049 (this one) adds tests under a tree no runner reaches.

Different mechanisms, same failure mode: a check that looks present, reports nothing, and is believed.

## Suggested fix

The `ai-tutors` remedy does not transfer. That worked because `ai-tutors/` is a directory of files with one script, so a `pyproject.toml` plus workspace membership was natural. `skills/` is different: skills are agent-facing markdown with helper scripts, symlinked into adopter repos. Making each skill a Python package would be heavy and would put build metadata into a tree that adopters symlink.

Two lighter options:

1. **One `pytest` invocation over `skills/`** wired into the existing tests workflow and a prek hook — collecting `skills/**/tests/` and `skills/**/test_*.py`, with `ruff` and `mypy` given the same paths. No per-skill packaging.
2. **A single `skills/pyproject.toml`** used only as a config carrier (`[tool.ruff]`, `[tool.mypy]`, `[tool.pytest.ini_options]`, `package = false`) and added to the workspace, so the existing driver picks the tree up as one member. Closer to how the repo already works, at the cost of one metadata file inside `skills/`.

Option 1 keeps `skills/` free of packaging metadata; option 2 reuses machinery that already exists and gets a `pytest (skills)` job for free. Either is fine — the important part is that something runs.

Note `check-workspace-members` will need to stay in step with whichever is chosen: it discovers members by scanning `tools/` and top-level directories (widened in #1011), so a `skills/` member would need to be discoverable or explicitly exempt.

potiuk:
## Summary

- Nothing checked the Python in `skills/`. The workspace checks iterate over `[tool.uv.workspace] members`, every member lived under `tools/`, so `ruff`, `mypy`, and `pytest` never saw `skills/**` — eight helper scripts and guards, plus the nine tests #1049 had just added, none of them run.
- Declares `skills/` as **one** workspace member whose `pyproject.toml` is a config carrier, not a package (`package = false`). The existing machinery does the rest: `run-workspace-check.sh` discovers applicable checks from the sections present, and `tests.yml` derives its matrix from the members list, so this adds a `pytest (skills)` job **with no workflow edit**.
- Turning the checks on immediately found seven real things, all fixed here.

Closes #1053.

## Type of change

- [ ] Groovy reference impl

## Test plan

- [x] `prek run --all-files` passes (23 hooks)
- [x] `ruff check`, `mypy`, and `pytest` all run and pass **through the workspace runner** for the new member — `uv run --project . python -m …` from `skills/`
- [x] **Verified the coverage is real, not nominal:** breaking the OSC 8 emission in `pr_link.py` now fails three tests under the workspace runner. Before this change the same break produced no signal at all — that is the whole point of the PR, so it seemed worth proving rather than asserting.
- [x] `check-workspace-members` passes — it already scans top-level directories (widened in #1011), so `skills` is discoverable and not reported as drift.

## RFC-AI-0004 compliance

- [ ] **Vendor neutrality** — placeholders used in all skill / tool prose

*(None apply — CI wiring plus type/lint fixes.)*

## Linked issues

Closes #1053. Follow-up to #1049, which added the first tests under `skills/`, and to #1011, which fixed the same class of gap for `ai-tutors/`.

**Why one member and not per-skill packaging.** Skills are symlinked into adopter repos one directory at a time (`.claude/skills/magpie-<name>` → `skills/<name>`), so a `pyproject.toml` inside each skill would leak build metadata into every adopter. `skills/pyproject.toml` is never symlinked — the relays point at `skills/<name>`, never at `skills/` itself. I checked that before choosing this shape.

**What the checks found.** None of it changes behaviour; these are the annotations and renames the checks require, which is rather the point — they were invisible until now.

- `scan_ci_runners.py` — `zip()` without `strict=`; **two same-named `futures` / `future` bindings of different types in one function**, which is why the second pair needed renaming before mypy could annotate them; an unannotated `rows`; and `step.get("with")` called twice, so the `isinstance` narrowing never stuck.
- `collect_status.py` — `l` as a variable name, and a mixed-value dict literal inferred as `dict[str, object]`, which made a later indexed assignment invalid.
- `test_pr_link.py` — a stale `# noqa: E402` suppressing nothing.

**One deliberate omission: `ruff format` is skipped** for this member via `[tool.magpie.checks] skip = ["ruff-format"]`. The formatter wants to reflow ~280 lines across six pre-existing scripts — long set literals and regex constants — none of which this change touches. Folding that in would bury the substance behind a mechanical reformat nobody can review line by line. Lint, types and tests all run; formatting can land as its own commit and the skip be deleted then. Happy to do that immediately after if you would rather have it in one go.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -95,6 +95,7 @@ package = false
 [tool.uv.workspace]
 members = [
   "ai-tutors",
+  "skills",
   "tools/agent-guard",
   "tools/agent-isolation",
   "tools/bitbucket",
--- a/skills/ci-runner-audit/scripts/scan_ci_runners.py
+++ b/skills/ci-runner-audit/scripts/scan_ci_runners.py
@@ -26,8 +26,9 @@ import json
 import re
 import subprocess
 import sys
-from concurrent.futures import ThreadPoolExecutor, as_completed
+from concurrent.futures import Future, ThreadPoolExecutor, as_completed
 from pathlib import Path
+from typing import Any
 from urllib.request import urlopen
 
 try:
@@ -177,9 +178,11 @@ def load_workflows_for_repos(repo_names: list[str], workers: int) -> list[dict]:
                 repos.append(repo)
     workflows: list[dict] = []
     with ThreadPoolExecutor(max_workers=workers) as executor:
-        futures = [executor.submit(list_workflows_for_repo, repo) for repo in repos]
-        for future in as_completed(futures):
-            workflows.extend(future.result())
+        workflow_futures: list[Future[list[dict]]] = [
+            executor.submit(list_workflows_for_repo, repo) for repo in repos
+        ]
+        for workflow_future in as_completed(workflow_futures):
+            workflows.extend(workflow_future.result())
     return sorted(workflows, key=lambda row: (row["repo"], row["path"]))
 
 
@@ -199,8 +202,8 @@ def matrix_rows(matrix: object) -> list[dict]:
             continue
         keys.append(str(key))
         values.append(value if isinstance(value, list) else [value])
-    rows = [{}]
-    for key, vals in zip(keys, values):
+    rows: list[dict[str, Any]] = [{}]
+    for key, vals in zip(keys, values, strict=True):
         rows = [{**row, key: val} for row in rows for val in vals]
 
     excludes = matrix.get("exclude")
@@ -303,7 +306,8 @@ def arch_hits(workflow: dict) -> list[dict]:
                 continue
             name = str(step.get("name", ""))
             uses = str(step.get("uses", ""))
-            action_inputs = step.get("with") if isinstance(step.get("with"), dict) else {}
+            raw_with = step.get("with")
+            action_inputs = raw_with if isinstance(raw_with, dict) else {}
             for key, value in action_inputs.items():
                 key_text = str(key).lower()
                 value_text = " ".join(lower_values(value))
--- a/skills/pr-management-triage/tests/test_pr_link.py
+++ b/skills/pr-management-triage/tests/test_pr_link.py
@@ -23,7 +23,7 @@ from pathlib import Path
 
 sys.path.insert(0, str(Path(__file__).resolve().parents[1] / "scripts"))
 
-import pr_link  # noqa: E402
+import pr_link
 
 
 class PrLinkTest(unittest.TestCase):
--- a/skills/pyproject.toml
+++ b/skills/pyproject.toml
@@ -53,7 +53,9 @@ package = false
 # here would bury the point of the change (turning the checks on) behind a
 # mechanical reformat nobody can review line by line. Lint, types, and tests
 # all run; formatting should land as its own commit, after which this skip can
-# be deleted.
+# be deleted. Tracked in
+# https://github.com/apache/magpie/issues/1076 — a skip with no tracking issue
+# is how a temporary exemption becomes a permanent one.
 [tool.magpie.checks]
 skip = ["ruff-format"]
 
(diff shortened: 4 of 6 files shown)


choo121600 (approved):
nice!
lgtm

potiuk:
Superseded by #1077, now merged as 8b3b474d.

#1077 started as the follow-up this PR asked for — deleting the temporary
`[tool.magpie.checks] skip = ["ruff-format"]` and applying the formatter —
but it had copied this branch's infrastructure rather than stacking on it,
so both PRs added the same `skills/pyproject.toml` and whichever merged
second would have conflicted. Rather than ask a first-time contributor to
rebase away work they had correctly identified as prerequisite, the two were
combined into #1077 with both of us as co-authors.

Verified before combining, rather than assumed: with this branch's
`ruff-format` skip removed and `uv run --directory skills --project . ruff
format .` applied, every `skills/**/*.py` file is byte-identical to #1077's
branch. `skills/pyproject.toml` there is this PR's file with exactly the
10-line skip block removed, and the root `pyproject.toml` member
registration and `uv.lock` delta are identical. So nothing from this PR was
dropped.

Three out-of-scope markdown reflows on #1077 were reverted before merge
(`ruff format` does not touch `.md`, and one had broken a deliberate column
alignment pairing a Python example with the Groovy example below it).

Net effect: everything this PR did has landed, plus the formatting pass, and
`ruff format (workspace)` now runs with no skip — so #1076 is closed
properly rather than deferred again. #1053 is closed by the merge.

PR #1055, opened 2026-08-01
fix(security): remove ASF address from skills

### LINKED ISSUE #998: Remove the hardcoded security@apache.org address from the security-cve-allocate skill
## Summary

`skills/security-cve-allocate/SKILL.md` contains a literal `security@apache.org` address.
Framework skills are meant to be project-agnostic, and this is the only hardcoded ASF security
address anywhere under `skills/`. Remove it without losing the guidance it carries.

## Background

Magpie separates two layers: the framework (project-agnostic skills and tools) and
adopter-specific configuration and adapters. `AGENTS.md` § *Repository purpose* says the
framework contains "no project-specific content", and PRINCIPLE 12 states it as "The framework
is project-agnostic; concrete names live in adopter config".

A fallback note about emailing the ASF security team was added to two files at once. In
`tools/cve-tool-vulnogram/allocation.md` it is correctly placed — that is the ASF-specific
adapter. In `skills/security-cve-allocate/SKILL.md` it is not: a non-ASF adopter running that
skill is told to email the Apache Software Foundation about their own CVE.

Two ways to fix it, either acceptable:

- Drop the note from the skill and have the skill point at the adapter section that already
  carries it. This also fixes a second problem — the two copies have already drifted, one saying
  "no governance-authorised member" and the other "no PMC member".
- Keep a note in the skill but use the `<asf-security-list>` placeholder, which the framework
  already uses for exactly this concept (see `AGENTS.md` lines 643 and 702).

Note that the automated placeholder check will not catch this: `tools/dev/check-placeholders.sh`
matches a fixed list of forbidden strings and that address is not on it. A green `prek` run is
not evidence the boundary held.

## Where to look

- `skills/security-cve-allocate/SKILL.md` — search for `security@apache.org` (one occurrence,
  in the fallback note near the "Wait for the user to report back a CVE token" instruction).
- `tools/cve-tool-vulnogram/allocation.md` — the correctly-placed copy of the same note.
- `AGENTS.md` § *Placeholder convention used in skill files*, and lines 643 / 702 for
  `<asf-security-list>` in use.

## Acceptance criteria

- [ ] `grep -rn "security@apache.org" skills/` returns nothing.
- [ ] The guidance is still reachable from the skill (either inline via a placeholder, or by a
      link to the adapter section).
- [ ] `tools/cve-tool-vulnogram/allocation.md` is unchanged or only gains the linked-to anchor.

## Estimated effort

~45 minutes for someone new to the codebase.

## Getting started

pollychen-lab:
## Summary

- Remove the hardcoded ASF security address from `security-cve-allocate` and point fallback guidance at the active CVE-tool adapter instead.
- Remove the remaining literal ASF security address from the GitHub advisory sync hand-off wording, keeping the generic `<security-team-list>` placeholder.

Fixes #998

## Validation

- `grep -rn "security@apache.org" skills/` returns no matches
- `uv run --project tools/skill-and-tool-validator --group dev skill-and-tool-validate`

### DIFF
--- a/skills/security-cve-allocate/SKILL.md
+++ b/skills/security-cve-allocate/SKILL.md
@@ -463,12 +463,12 @@ re-invoke this skill with the CVE ID as an override argument to
 resume from Step 4.
 
 **Fallback when no governance-authorised member is reachable:**
-if every authorised member is unavailable or unresponsive, send
-an email to `security@apache.org` with subject
-`CVE request for <one-line description>`. The ASF Security Team
-will allocate the CVE and send the ID back. Re-invoke this skill
-with the returned `CVE-YYYY-NNNNN` as an override to resume from
-Step 4.
+if every authorised member is unavailable or unresponsive, follow
+the fallback path documented by the active CVE-tool adapter (for
+the Vulnogram adapter, see
+[`tools/cve-tool-vulnogram/allocation.md` § *PMC-gated access*](../../tools/cve-tool-vulnogram/allocation.md#pmc-gated-access)).
+When the fallback returns a `CVE-YYYY-NNNNN`, re-invoke this skill
+with that ID as an override to resume from Step 4.
 
 **Wait for the user** to report back a `CVE-\d{4}-\d+` token. Do
 not proceed to Step 4 until that token has arrived. If the user
--- a/skills/security-issue-sync/github-advisory.md
+++ b/skills/security-issue-sync/github-advisory.md
@@ -126,7 +126,7 @@ path.
 ### Delivery — an email relay (draft, never auto-sent)
 
 Create a **draft** email to the org's advisory-admin security team
-(`<security-team-list>`; for ASF, `security@apache.org`) with
+(`<security-team-list>`) with
 `oauth-draft-create` — never send directly (SKILL Golden rule 1; and the
 Gmail MCP mangles the `security/advisories/GHSA-…` URLs into redirects, so
 use oauth-draft). **Always CC the project `<security-list>`** so the


potiuk (approved):
LGTM — and the `security-cve-allocate` half is a better fix than "delete the hardcoded address" usually gets.

### What I checked

Both occurrences really were there, and this removes both:

```text
skills/security-cve-allocate/SKILL.md:467       an email to `security@apache.org` with subject
skills/security-issue-sync/github-advisory.md:129  (`<security-team-list>`; for ASF, `security@apache.org`) with
```

Those are the only two in `skills/` on `main`, so the stated `grep -rn "security@apache.org" skills/` result holds after this change.

The substitution is the good part. The old fallback did not just *name* an ASF address, it encoded an ASF *procedure* — mail the ASF Security Team, they allocate and mail the ID back. Swapping that for "follow the fallback path documented by the active CVE-tool adapter" moves the ASF-specific mechanics behind the adapter boundary, which is where [`AGENTS.md` § Placeholder convention](https://github.com/apache/magpie/blob/main/AGENTS.md#placeholder-convention-used-in-skill-files) wants them — *"skills must stay project-agnostic so swapping projects is a config change, not a code change."* A non-ASF adopter reading the old text had no path at all; now they are pointed at whichever adapter they actually run.

I verified the citation resolves rather than trusting lychee: `tools/cve-tool-vulnogram/allocation.md` exists and carries `## PMC-gated access` at line 44.

### Smaller observation

Worth a follow-up, not a change here.

`github-advisory.md` now reads `(<security-team-list>) with …`, where before the parenthetical glossed it as *"for ASF, `security@apache.org`"*. Removing the gloss is right — that was the hardcoding. But it does mean the placeholder now carries the whole meaning on its own, and it is a one-off:

| placeholder | files using it | registered in `AGENTS.md`? |
|---|---|---|
| `<security-list>` | 13 | yes — line 363, sourced from `<project-config>/project.md → mailing_lists.security` |
| `<security-team-list>` | 1 (this file) | no |

`<security-team-list>` predates this PR, so nothing here invents it, and `check-placeholders` will not catch it — that hook scans for forbidden *literals*, not for unregistered placeholder *names*, which is why CI is green either way.

Normalising it to `<security-list>` would be a one-word change and would make the line resolvable by the same config key as the other thirteen. I would not fold it in here, since it is a pre-existing inconsistency in a file this PR is only touching in passing.

(Separately, and definitely not this PR's problem: `projects/_template/project.md` declares no `mailing_lists` block at all, so even the registered `<security-list>` has no key an adopter would fill in. That gap is worth its own issue.)

---

> will follow up.*
>

pollychen-lab:
@potiuk thanks

PR #1056, opened 2026-08-01
docs(adapters): fix add-a-harness recipe

### LINKED ISSUE #997: Fix the two broken instructions in the add-a-harness recipe
## Summary

`docs/adapters/add-a-harness.md` is the recipe for wiring a new agent runtime into Magpie. Two
of its copy-and-run instructions do not match the repository: the agent-guard file layout in
Step 2, and the verification command in Step 1. Correct both.

## Background

The doc exists, in its own words, "so the extension point is a recipe, not tribal knowledge", so
a wrong path or command costs exactly the reader it was written for.

**Step 2 shows a layout that does not exist.** It draws the agent-guard package as
`tools/agent-guard/agent_guard/` with an `adapters/` directory holding one file per harness
(`claude.py`, `opencode.py`, `<your-runtime>.py`). The real package root is
`tools/agent-guard/src/agent_guard/`, and there is no `adapters/` directory at all. Each
harness is a function in `__init__.py` — `opencode_main()`, `kiro_main()` — selected by flag
routing (`--opencode`, `--kiro`) near the bottom of the same file. A contributor who follows
the recipe creates a file that nothing imports.

The Kiro integration (PR #768) says in its commit message that it followed this recipe, and it
did **not** do what Step 2 describes — it added `kiro_main()` plus a `--kiro` branch. So the
recipe's first real user had to work out the true shape anyway.

**Step 1's verification command fails.** It gives
`uv run --project tools/symlink-lint python3 -m symlink_lint`. Running a package with
`python3 -m` needs a `__main__.py`, and `tools/symlink-lint/src/symlink_lint/` contains only
`__init__.py`, so the command exits with "No module named `symlink_lint.__main__`". The tool
ships a console script instead. This is the command the reader runs right after hand-creating
symlinks, so the spurious failure looks like their symlinks are broken.

## Where to look

- `docs/adapters/add-a-harness.md` — Step 1's verification command and Step 2's file-tree block.
- `tools/agent-guard/src/agent_guard/__init__.py` — search for `def opencode_main`,
  `def kiro_main`, and the flag routing near the end of the file, to describe the real shape.
- `tools/symlink-lint/pyproject.toml` — the `[project.scripts]` entry gives the real command
  name.
- Other docs in the repo that invoke per-tool commands show the `uv run` form in use.

## Acceptance criteria

- [ ] Step 2's file tree matches the real package: the `src/` layer is present and no
      `adapters/` directory is shown.
- [ ] Step 2 describes adding a harness the way the code actually does it (a `*_main()`
      function plus a flag branch), rather than a new file per harness.
- [ ] Step 1's verification command runs successfully in a clean checkout.

## Estimated effort

~1 hour for someone new to the codebase. No code changes; the work is reading
`agent_guard/__init__.py` closely enough to describe it correctly.

## Getting started

pollychen-lab:
## Summary

- Replace the broken `python3 -m symlink_lint` examples with the shipped `symlink-lint` console script.
- Update the action-guard file tree to match the current `tools/agent-guard/src/agent_guard/__init__.py` layout.
- Describe adding a hook-shaped harness as a `*_main()` adapter plus a `cli()` routing branch, matching the existing OpenCode and Kiro paths.

Fixes #997

## Validation

- `uv run --project tools/symlink-lint symlink-lint`
- `uv run --directory tools/agent-guard --project . --group dev pytest`
- `uv run --project tools/skill-and-tool-validator --group dev skill-and-tool-validate`
- `bash -n tools/spec-loop/loop.sh && bash -n tools/spec-loop/lib.sh`

### DIFF
--- a/docs/adapters/add-a-harness.md
+++ b/docs/adapters/add-a-harness.md
@@ -77,7 +77,7 @@ relay symlinks automatically. The rules are:
 verify after any manual link creation:
 
 ```bash
-uv run --project tools/symlink-lint python3 -m symlink_lint
+uv run --project tools/symlink-lint symlink-lint
 ```
 
 ## Step 2 — wire the action guard
@@ -93,20 +93,24 @@ existing OpenCode example.
 
 ```text
 tools/agent-guard/
-  src/agent_guard/__init__.py    ← dispatch() core (harness-agnostic)
-  src/agent_guard/__main__.py    ← CLI entry point (stdin JSON → stdout JSON)
-  src/agent_guard/guards.d/     ← individual guard scripts
+  src/agent_guard/__init__.py    ← dispatch() core + thin CLI adapters
+  src/agent_guard/guards.d/      ← individual guard scripts
   opencode/plugin.js             ← OpenCode adapter (JS plugin)
-  tests/                         ← test suite incl. per-harness tests
+  tests/                         ← test suite including per-harness cases
 ```
 
-The core (`__init__.py`) exposes a `dispatch()` function. Per-harness
-adapters translate the runtime's hook format into a `dispatch()` call.
-Your adapter must:
+The core (`__init__.py`) exposes `dispatch()` plus one thin `*_main()`
+function per hook-shaped harness (`opencode_main()`, `kiro_main()`,
+…). The `cli()` router near the end of the file selects those adapters
+from flags such as `--opencode` and `--kiro`. To add another hook-shaped
+harness, follow that pattern:
+
 1. Parse the harness's pre-tool hook payload (stdin JSON, env vars,
    CLI args, or a native plugin format — check the runtime docs).
 2. Call `dispatch(command_string)` from the core module.
 3. Return the harness's expected "allow" or "block" response.
+4. Add a `--<runtime>` branch in `cli()` that routes to the new
+   `*_main()` function.
 
 Run the guard's test suite to verify the adapter:
 
@@ -232,7 +236,7 @@ in its comment header.
 
 ```bash
 # 1. Symlink topology (no cycles, correct relay direction)
-uv run --project tools/symlink-lint python3 -m symlink_lint
+uv run --project tools/symlink-lint symlink-lint
 
 # 2. Skill + tool metadata (capability, prerequisites, organization lines)
 uv run --project tools/skill-and-tool-validator --group dev skill-and-tool-validate


potiuk (approved):
LGTM — all three corrections check out, and one of them fixes a command that could never have worked.

### The broken invocation was genuinely broken

`python3 -m symlink_lint` fails outright:

```text
No module named symlink_lint.__main__; 'symlink_lint' is a package and cannot be directly executed
```

`tools/symlink-lint/src/symlink_lint/` contains only `__init__.py` — no `__main__.py` — so `-m` has nothing to execute. The shipped console script is declared (`[project.scripts] symlink-lint = "symlink_lint:main"`) and runs clean. Anyone following the recipe literally hit a traceback at the first validation step; both occurrences are now fixed.

Worth noting why this was easy to miss: the sibling `agent_guard` package *does* have a `__main__.py`, so `python3 -m agent_guard` works. The two tools look interchangeable and are not.

### The `cli()` description matches the code

Verified against `tools/agent-guard/src/agent_guard/__init__.py`:

```text
554  def opencode_main() -> int
584  def kiro_main() -> int
729  def cli(argv: list[str] | None = None) -> int
```

So "one thin `*_main()` per hook-shaped harness, selected by a `cli()` router near the end of the file" is exactly right, and the new step 4 — add a `--<runtime>` branch in `cli()` — is the step the old text omitted entirely. That is the part that actually helps someone adding a harness.

### On dropping `__main__.py` from the tree

I flagged this to myself first, because the file does still exist, and then changed my mind on the evidence. Recording the reasoning since the deletion looks wrong at a glance:

The old row described `__main__.py` as *"CLI entry point (stdin JSON → stdout JSON)"*, and that is not how the guard actually runs. `tools/agent-guard/README.md` says the hook is invoked as `python3 <path>/agent_guard/__init__.py`, **never via `uv run`**, shipped as "a single self-contained file" — and `__init__.py` duly ends with `if __name__ == "__main__": raise SystemExit(cli())`. So `__init__.py` is the real entry point; `__main__.py` is a three-line `python -m` shim that the deployed path never touches. Removing the row deletes a claim that was actively misleading, in a document whose job is to tell a contributor where things live.

Only nit, and genuinely optional: `__main__.py` is still on disk, so someone listing the directory sees a file the tree does not mention. A single row — `← python -m shim → cli()` — would cost nothing and pre-empt the "what is this for?" question. Silence is defensible for a recipe rather than an inventory, so take it or leave it.

---

> will follow up.*
>

pollychen-lab:
@potiuk thanks

Issue #1057, opened 2026-08-01
docs(placeholders): the org-level security address has three names and none is registered


potiuk:
## What

The *org-level* security address — the foundation-wide advisory-admin inbox, as distinct from a project's own security list — is referred to by three different names, none of which is registered in the placeholder table in [`AGENTS.md` § Placeholder convention](https://github.com/apache/magpie/blob/main/AGENTS.md#placeholder-convention-used-in-skill-files):

| name | where it appears | notes |
|---|---|---|
| `<security-team-list>` | `skills/security-issue-sync/github-advisory.md` | one use, in the advisory-relay delivery step |
| `<asf-security-list>` | `AGENTS.md` (two uses, cross-project triage guidance) | ASF-branded, in a framework whose whole premise is governance-agnosticism |
| `security_inbox.foundation_security_address` | `projects/_template/project.md` | the actual org-level config path |

By contrast the *project* list is unambiguous: placeholder `<security-list>`, key `security_list`, registered in the table.

## Why it matters

The distinction is real and load-bearing. `github-advisory.md` uses both in the same instruction:

> Create a **draft** email to the org's advisory-admin security team (`<security-team-list>`) … **Always CC the project `<security-list>`** so the project security team stays looped in.

Two different addresses, deliberately. Anyone "tidying" the one-off `<security-team-list>` into the registered `<security-list>` — which looks like an obvious normalisation — would silently turn that into *"email X and CC X"*, dropping the project security team from an advisory relay. I nearly did exactly that while following up on #1055, and only stopped after reading the surrounding sentence.

So the current state is a trap: the odd-looking name is the correct one, and the "fix" is wrong.

There is a second-order problem too. `<asf-security-list>` bakes a vendor into a placeholder, in a repository that has spent real effort removing exactly that (#1055 removed the last hardcoded `security@apache.org` from `skills/`, #997 and #998 before it). A non-ASF adopter reading `AGENTS.md` sees a placeholder named after a foundation they are not part of.

## Suggested fix

Decide on one name for the org-level address, register it in the placeholder table with its source key, and use it consistently. Something org-neutral — `<org-security-list>`, or `<foundation-security-list>` to match the config key — would sit better alongside `<governance-body>` and `<project-stage>`, which are already named to avoid ASF-specific vocabulary.

Then:

- replace both `<asf-security-list>` uses in `AGENTS.md`;
- replace the `<security-team-list>` use in `github-advisory.md`;
[...]

PR #1058, opened 2026-08-01
fix(docs): point <security-list> at the config key that exists

potiuk:
## Summary

- Both places that tell an agent how to resolve `<security-list>` named `mailing_lists.security` — a key that does not exist anywhere in the tree. The adopter manifest declares `security_list`, under a `## Mailing lists` heading.
- An agent following either instruction looks for a key it will never find, and cannot resolve the placeholder.
- Fixed in `AGENTS.md` § Placeholder convention and in `skills/security-cve-allocate/SKILL.md`, where the sibling entries already use the flat form and this one was the odd one out.

## Type of change

- [ ] Groovy reference impl

## Test plan

- [x] `prek run` passes on both changed files
- [x] Verified against `projects/_template/project.md` that `security_list` is the declared key, alongside `private_list`, `users_list`, `dev_list`, `announce_list`, `commits_list`
- [x] Checked the rest of the placeholder table rather than assuming: `tracker_repo`, `upstream_repo`, and `upstream_default_branch` all resolve in the template. `mailing_lists.security` was the only cited key with no definition, so this is an isolated error and not a pattern
- [x] Confirmed `mailing_lists` appears nowhere in the repo except the two lines this PR corrects

## RFC-AI-0004 compliance

- [x] **Vendor neutrality** — the placeholder mechanism is what keeps concrete project values out of skills; a placeholder that cannot be resolved is that mechanism failing quietly

## Linked issues

Follow-up to #1055. The related org-level naming problem is #1057.

**What I deliberately did not change**, because it looks like the obvious tidy-up and is wrong:

`skills/security-issue-sync/github-advisory.md` uses `<security-team-list>` once, and it is tempting to normalise that to the registered `<security-list>`. It should not be. The two appear in the same instruction and mean different addresses:

> Create a **draft** email to the org's advisory-admin security team (`<security-team-list>`) … **Always CC the project `<security-list>`** so the project security team stays looped in.

Collapsing them would turn that into "email X and CC X", quietly dropping the project security team from an advisory relay. I started to make that change and stopped after reading the surrounding sentence.

The underlying problem is that the org-level address has three names in the tree — `<security-team-list>`, `<asf-security-list>` (ASF-branded, in a governance-agnostic framework), and the config path `security_inbox.foundation_security_address` — and none is registered in the placeholder table. That wants deciding deliberately rather than guessing at inside a typo fix, so it is #1057.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -360,7 +360,7 @@ configuration before executing any command:
 | `<framework>` | The framework root — `.apache-magpie/` (the gitignored snapshot) in adopting projects, `.` in framework standalone. Used in `uv run` and other invocations that address the framework's `tools/<name>/` subtrees. | Filesystem convention. |
 | `<tracker>` | GitHub slug of the (security) tracker repo (example: `airflow-s/airflow-s`). | `<project-config>/project.md` → `tracker_repo` |
 | `<upstream>` | GitHub slug of the upstream codebase the fixes land in (example: `apache/airflow`). | `<project-config>/project.md` → `upstream_repo` |
-| `<security-list>` | The project's security mailing list (example: `security@airflow.apache.org`). | `<project-config>/project.md` → `mailing_lists.security` |
+| `<security-list>` | The project's security mailing list (example: `security@airflow.apache.org`). | `<project-config>/project.md` → `security_list` (under **Mailing lists**) |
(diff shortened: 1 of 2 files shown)


choo121600 (approved):
I haven't actually tried it, but it looks pretty good.

PR #1062, opened 2026-08-02
Add multi-SKIP case to gfi-sweep eval suite

### LINKED ISSUE #1001: Add a multi-SKIP case so the gfi-sweep summary-only rule can actually fail
## Summary

In the `good-first-issue-sweep` Step 3 eval suite, the field that checks SKIP issues are shown
as a count summary is `true` in all four cases, three of them vacuously. Add a case with
several SKIP issues so the rule has something to fail on.

## Background

`skip_count_shown_as_summary_only` encodes a real presentation rule: SKIP issues must appear as
a count line, not as a list of titles and reasons. But of the four existing cases, only
`case-2-mixed` has any SKIP issues at all — and it has one. The output-spec says that when
there are no SKIP issues the field is `true` "vacuously satisfied", so the other three expect
`true` for free.

That means a model that always answers `true` passes the whole suite on this field, including
one that helpfully lists every skipped issue with its reason. The project's own guidance in
`docs/education/eval-driven-development.md` names this pattern under "All your cases expect the
same value".

The fix is a new case whose input has enough SKIP issues that listing them is the tempting
behaviour — five or six, with distinct-looking skip reasons.

## Where to look

- `tools/skill-evals/evals/good-first-issue-sweep/step-3-present-proposals/fixtures/` — add a
  new `case-5-` directory here.
- `.../fixtures/case-2-mixed/report.md` — the closest existing input to model yours on; it shows
  the classification-list format the step receives.
- `.../fixtures/case-2-mixed/expected.json` — the eight fields your `expected.json` needs.
- `.../fixtures/output-spec.md` — the definition of each field, including the vacuous-true rule.
- `tools/skill-evals/evals/good-first-issue-sweep/README.md` — the case table to extend, and the
  case count to bump.

## Acceptance criteria

- [ ] A new case supplies at least five SKIP issues with distinct skip reasons, plus at least
      one READY issue so the rest of the output is exercised.
- [ ] Its `expected.json` sets all eight fields from the output-spec.
- [ ] The suite README lists the new case and its count is updated.
- [ ] The suite still assembles: `PYTHONPATH=tools/skill-evals/src python3 -m
      skill_evals.runner tools/skill-evals/evals/good-first-issue-sweep/` runs without error.

## Estimated effort

~1-2 hours for someone new to the codebase; most of it is writing a realistic input.

## Getting started

deepusnath:
## Summary
- Every existing `step-3-present-proposals` case in the good-first-issue-sweep eval suite has 0 or 1 SKIP issues, so `skip_count_shown_as_summary_only` was only ever vacuously satisfied — a model that always returns `true` would pass.
- Adds `case-5-multi-skip`: 1 READY issue plus 6 SKIP issues spread across all three `skip_reason` categories (2 security-sensitive, 2 architectural-decision, 2 deprecation-decision), so the summary-only rule and the `(security: M, architectural: K, deprecation: J)` breakdown format from the skill's Step 3 SKIP section actually get exercised.
- Updates the suite README to document the new case and bump the step-3 case count from 4 to 5.

Closes #1001

## Test plan
- [x] `PYTHONPATH=tools/skill-evals/src python3 -m skill_evals.runner tools/skill-evals/evals/good-first-issue-sweep/step-3-present-proposals/fixtures/` — all 5 cases load and render cleanly (exit 0).
- [x] Verified `expected.json` contains all 8 output-spec fields with correct values for the new case.


potiuk:
Superseded by #1072, merged as d0571f6f.

Both PRs added `case-5-multi-skip` at the same paths in the same suite, so
only one could land — they would have conflicted on identical new files.
Nothing was wrong with this one: the counts are internally consistent
(1 READY + 6 SKIP = 7 issues, matching `ready_count: 1` /
`near_miss_count: 0`), `deprecation-decision` is a documented `skip_reason`,
and the README counters were correctly updated.

I went with #1072 because it is the later PR, its CI was green, and keeping a
NEAR-MISS alongside the SKIP block tests `skip_count_shown_as_summary_only`
and `near_miss_has_label_proposal: false` holding at the same time rather
than each in isolation.

Your design had a real argument in its favour that I want on the record:
two SKIPs per reason (2 security / 2 architectural / 2 deprecation) stresses
the summary-only rule harder than #1072's 2/1/1, and isolates it more
cleanly given NEAR-MISS behaviour is already covered by `case-2-mixed` and
`case-3-near-miss-only`. If we want to strengthen the mix, that is now a
small edit to the merged fixture rather than a new case.

Closing as superseded, not rejected. Thanks for the second look at the same
problem — please just fold a revision into the original PR next time rather
than opening a second one, so the two don't race for the same file.

PR #1065, opened 2026-08-03
docs: reflect the shipped 0.1.0 release in adoption docs

justinmclean:
## Summary

0.1.0 is released (promoted to dist/release/magpie/0.1.0), so the adoption docs no longer match reality:

- README + install-recipes: svn-zip is the recommended production path rather than "forthcoming"; git-branch is no longer "the default during the framework's pre-release phase".
- install-recipes Method 1 fetched apache-magpie-<version>-source-release.zip, which is not the artefact name that shipped. Every other doc in the tree says -source.zip, so the recipe as written 404s.
- install-recipes Method 1 templated a <PROJECT> adopter dist subdirectory. Magpie has its own dist area now, so hard-code it and split DIST_BASE (KEYS at the project root) from REL_BASE (artefacts under the per-version subdirectory).
- README cross-reference still called MISSION.md a draft proposal; it was adopted by Board resolution on 17 June 2026, and the NOTE block earlier in the README already says so.

## Type of change

- [ ] Groovy reference impl
- [X] Documentation (`docs/`, `README.md`, `CONTRIBUTING.md`)
- [ ] Other:

## Test plan

- [ ] Other:

### DIFF
--- a/README.md
+++ b/README.md
@@ -98,9 +98,9 @@ Pick an install method and follow the verbatim recipe in
 
 | Method | When to use | Reproducibility |
 |---|---|---|
-| `svn-zip` | Production once ASF official releases ship to `dist.apache.org` (signed + checksummed) | Frozen by version |
-| `git-tag` | Pin a specific framework version | Frozen by tag |
-| `git-branch` (default `main`) | WIP path — track the framework's `main` directly for the latest unreleased changes. | Tracks tip |
+| `svn-zip` (**recommended**) | Production. Signed + checksummed zip from `dist.apache.org`; `0.1.0` is the current release. | Frozen by version |
+| `git-tag` | Pin a version without going through the signed zip (e.g. testing a release candidate). | Frozen by tag |
+| `git-branch` (`main`) | WIP path — track the framework's `main` directly for the latest unreleased changes. | Tracks tip |
 
 Each recipe is a single shell block that:
 
--- a/docs/setup/install-recipes.md
+++ b/docs/setup/install-recipes.md
@@ -31,9 +31,9 @@ Pick the recipe that matches your distribution preference:
 
 | Method | When to use | Reproducibility |
 |---|---|---|
-| [**svn-zip**](#method-1--released-zip-from-asf-distribution) | Production adopters once the framework ships official ASF releases. Signed + checksummed. | Frozen by version |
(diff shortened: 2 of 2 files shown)


potiuk (dismissed):

potiuk (approved):
Approving. I rebased this onto current `main` since it had gone conflicting,
and verified the recipe against the live distribution area rather than
against the PR text.

### This fixes a real error, not just staleness

The point worth calling out: the previous recipe told adopters to substitute
**their own** project's dist subdirectory —

```
# Replace <PROJECT> with the host adopter's ASF dist subdirectory
# (e.g. `airflow` once releases land at
# https://dist.apache.org/repos/dist/release/airflow/).
```

— but the artefact being downloaded is the *Magpie framework's* zip, which
lives under Magpie's own dist area. An adopter following the old instructions
verbatim would have 404'd. Hardcoding `magpie` here is correct, not a
vendor-neutrality slip: this path names the framework being installed, not
the adopter's project.

### Verified against the live dist area

Every factual claim in the new recipe checks out against
`https://dist.apache.org/repos/dist/release/magpie/`:

- The artefact really is `apache-magpie-0.1.0-source.zip` — the old recipe's
  `apache-magpie-${VERSION}-source-release.zip` was wrong.
- `KEYS` really does sit at the project root, and the artefacts really do sit
  under a per-version `0.1.0/` subdirectory — so the `DIST_BASE` /
  `REL_BASE` split is right, and `curl -fsSLO ${DIST_BASE}/KEYS` correctly
  reaches for the root while the zip, `.asc`, and `.sha512` come from
  `REL_BASE`.

`prek run --from-ref main` is green, including lychee against the new
`dist.apache.org` URLs.

### One conflict, resolved toward `main`

`main` had independently rewritten the `MISSION.md` cross-reference line to
"founding mission of the established TLP", which already achieves what your
edit to that line was for. I kept `main`'s wording rather than re-applying
the longer "as filed in the establishment proposal adopted by the Board on
17 June 2026" phrasing — no reason to re-open a line someone had already
settled, and it keeps this PR on its actual subject. Say the word if you want
the Board date restored; it is a one-line follow-up.

Everything else rebased untouched, and your authorship is preserved on the
rebased commit.

---

> will follow up.*
>

PR #1070, opened 2026-08-05
fix(validator): scan tool-root Python and 'from socket import' for telemetry

### LINKED ISSUE #783: no-telemetry-import validator: broaden scan beyond src/ to match the PRINCIPLE 10 guarantee wording
Follow-up from the review of #763 (the `no-telemetry-import` SOFT validator check backing PRINCIPLE 10).

The check works correctly, but its enforcement scope is narrower than the guarantee stated in `tools/egress-gateway/tool.md`. Tracking the gaps here so #763 can merge:

1. **Scan scope is `src/`-only.** `validate_no_telemetry_imports` only descends into `tools/<name>/src/`, but several substrate tools keep Python at the tool root (e.g. `pr-management-stats/{dashboard,reference}.py`, `security-tracker-stats-dashboard/{fetch_roster,fetch_prs,render}.py`, `dashboard-generator/reference.py`). A network import in any of those isn't caught — yet `tool.md` promises "a substrate tool that accidentally grows a network import is caught before it is merged." Either broaden the scan to all `*.py` under the tool dir (minus `tests/`), or soften the `tool.md` wording.
2. **`from socket import …` not matched.** The `socket` pattern is `^\s*import\s+socket\b` only; add a `^\s*from\s+socket\s+import\b` companion. Note the library list is deliberately non-exhaustive.
3. **SOFT-only.** The check is advisory unless `--strict`, so an accidental import merges without `--strict` — worth noting in the guarantee wording.

Raised in the #763 review; deferred to keep the working check moving.

potiuk:
## Summary

Closes #783.

`tools/egress-gateway/tool.md` promised that *"a substrate tool that accidentally grows a network import is caught before it is merged."* For 11 Python files it did not.

- **Scan every `.py` the tool owns**, not just `tools/<name>/src/`. Several substrate tools keep Python at the tool root — `pr-management-stats/{dashboard,reference}.py`, `security-tracker-stats-dashboard/{fetch_roster,fetch_prs,render}.py`, `dashboard-generator/reference.py` — and a `src/`-only scan exempted every one of them. `tests/` stays excluded: a test may legitimately import a network module to assert it is unreachable.
- **Add the `from socket import` companion pattern.** Every sibling pattern already accepted both `import x` and `from x import`; `socket` accepted only the former. Clearly an oversight rather than a decision.
- **Correct the `tool.md` guarantee** so it states the real scope and the two limits it previously implied away.

## Type of change

- [ ] Groovy reference impl

## Test plan

- [x] `prek run --files` green on all three files (ruff, ruff-format, mypy, pytest, markdownlint, typos, lychee, skill-and-tool-validate)
- [x] `uv run --directory tools/skill-and-tool-validator --group dev pytest` passes
- [x] **Before/after proven by planting the defects.** `import requests` added to `tools/pr-management-stats/dashboard.py` and `from socket import socket` to `reference.py` — both real tool-root files:

  | Validator | Violations found |
  |---|---|
  | original (`main`) | **0** — both missed |
  | this PR | **2** — both caught |

- [x] **Clean tree is 0 violations before and after**, so no existing tool is newly flagged and this does not turn CI red on merge
- [x] Three tests added; the two behavioural ones fail on the previous implementation with `assert 0 == 1`

## RFC-AI-0004 compliance

- [x] **Sandbox** — this is the check backing PRINCIPLE 10's zero-default-egress guarantee; the change widens enforcement to the files that were silently exempt
- [ ] **Vendor neutrality**
- [ ] **Conversational + correctable**
- [ ] **Write-access discipline**
- [x] **Privacy LLM** — the no-telemetry invariant is what stops a substrate tool quietly acquiring an outbound call

## Linked issues

Closes #783. Follow-up from the review of #763.

## Notes for reviewers

**On the `tool.md` rewording.** I first considered only widening the scan and leaving the prose. That would still have overclaimed: the check is SOFT unless `--strict`, so an accidental import can merge on a run without the flag. The doc now says plainly that it is *"a tripwire the reviewer reads, not a gate that stops the merge"*, and that the library list is deliberately non-exhaustive. A guarantee that overstates its own strength is worse than a narrower one stated accurately — someone reads it and stops looking.

**On the `tests/` exclusion.** Deliberate, and the third test pins it. Broadening the scan without it would have started flagging test files that import `requests` precisely to assert it cannot reach anywhere — a check that fires on its own negative controls trains people to ignore it.

**Verification method.** The clean tree reports 0 violations both before and after, so a passing run proves nothing on its own. The planted-defect comparison in the test plan above is what establishes that the broadened scan actually reaches the new files.

### DIFF
--- a/tools/egress-gateway/tool.md
+++ b/tools/egress-gateway/tool.md
@@ -113,11 +113,19 @@ action (PRINCIPLE 10).  This guarantee rests on two pillars:
    all other `substrate:*` tools are network-free by design.
 
 2. **The validator enforces the invariant.**  `tools/skill-and-tool-validator/`
-   check #21 (`no-telemetry-import`) scans the source of every `substrate:*`
-   tool for network-calling imports (`requests`, `httpx`, `aiohttp`,
-   `urllib.request`, `http.client`, `socket`) and flags any hit as a SOFT
-   advisory.  A substrate tool that accidentally grows a network import is
-   caught before it is merged.
+   check #21 (`no-telemetry-import`) scans every Python file owned by a
+   `substrate:*` tool — under `src/` and at the tool root alike, excluding
+   `tests/` — for network-calling imports (`requests`, `httpx`, `aiohttp`,
+   `urllib.request`, `http.client`, `socket`, in both `import x` and
+   `from x import` form) and flags any hit as a SOFT advisory.  A substrate
+   tool that accidentally grows a network import is surfaced in the
+   validator's output before it is merged.
+
+   Two limits worth stating plainly.  The check is **advisory unless
+   `--strict`**, so an accidental import can merge on a run that does not
+   pass the flag — it is a tripwire the reviewer reads, not a gate that
+   stops the merge.  And the library list is deliberately non-exhaustive:
+   it catches the common accidents, not a determined author.
 
 The declared egress surfaces as of this writing:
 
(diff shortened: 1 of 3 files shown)


choo121600 (approved):
nice!

PR #1072, opened 2026-08-06
Add multi-SKIP case to good-first-issue-sweep step-3 evals

deepusnath:
The step-3-present-proposals fixtures only ever exercised a single SKIP issue at a time, so the "summary count only, grouped by reason" rule (issue #1001) couldn't actually fail: a model that just prints the one issue's title/reason still looks like a valid one-line summary.

case-5-multi-skip adds 4 SKIP issues spanning all three skip reasons (2 security, 1 architectural, 1 deprecation) alongside READY/NEAR-MISS issues, so a regression that lists individual SKIP titles/reasons or miscounts a category is actually caught.

## Summary

-

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

## RFC-AI-0004 compliance

### DIFF
--- a/tools/skill-evals/evals/good-first-issue-sweep/README.md
+++ b/tools/skill-evals/evals/good-first-issue-sweep/README.md
@@ -12,7 +12,7 @@ skill.
 | Step | Eval | Cases |
 |---|---|---|
 | Step 2 — Classify each issue | `step-2-classify` | 6 |
-| Step 3 — Present proposals | `step-3-present-proposals` | 4 |
+| Step 3 — Present proposals | `step-3-present-proposals` | 5 |
 
 ## step-2-classify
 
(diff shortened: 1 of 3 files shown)


potiuk (approved):
Approving. The gap this fills is real and worth naming precisely, because
it is easy to look at the existing suite and think it was already covered.

`case-2-mixed` does already assert `skip_count_shown_as_summary_only: true`
— but with exactly **one** SKIP, where "shown as a summary count" and
"listed individually" are nearly indistinguishable in the output. A model
that enumerates skip titles could pass that case. With four SKIPs spanning
all three skip reasons, enumerating becomes visibly wrong, so the invariant
is genuinely exercised for the first time.

Checked:

- Counts are internally consistent — the report header says 7 issues and
  lists exactly 2 READY (42, 88), 1 NEAR-MISS (77), and 4 SKIP (101, 102
  `security-sensitive`; 103 `architectural-decision`; 104
  `deprecation-decision`), matching `ready_count: 2` and
  `near_miss_count: 1`.
- `deprecation-decision` is a real value, not invented for the fixture —
  `skills/good-first-issue-sweep/SKILL.md:214` declares
  `security-sensitive | architectural-decision | deprecation-decision | null`.
  This is also the first fixture in the suite to exercise the deprecation
  reason at all; `step-2-classify` covers only G5 and G6.
- Keeping the NEAR-MISS in the mix means the case also re-verifies
  `near_miss_has_label_proposal: false` while several SKIPs are being
  summarised — the two rules holding simultaneously, rather than each in
  isolation.
- Both README counters are updated: the Coverage table (4 → 5) and the
  per-case table row.

### On the duplicate

This supersedes #1062, which proposed the same `case-5-multi-skip` at the
same paths with a different mix (1 READY + 6 SKIP, two per reason, no
NEAR-MISS). Because both create identical new files, only one could land.
I have gone with this one: it is the later PR, its CI is green, and keeping
a NEAR-MISS alongside the SKIP block tests the two invariants together.

#1062's design had a fair argument in its favour — two SKIPs per reason
stresses the summary rule harder than 2/1/1 — so if you want to strengthen
the mix later, that is a one-line change to this fixture rather than a new
case. Closing #1062 as superseded rather than rejected.

Thanks for both attempts; a second design is more useful than no second
opinion, it just needed to be one PR.

---

> will follow up.*
>

PR #1073, opened 2026-08-07
docs(security): add reporter-unresponsiveness escalation per ASF policy to security-issue-sync

### LINKED ISSUE #182: surface reporter-unresponsive escalation when reporter has not replied within a configurable threshold
## Summary

The ASF `security_committers` policy (https://www.apache.org/security/committers.html) states:

> The project team agrees on the fix, the announcement, and the release schedule
> with the reporter. If the reporter is unresponsive in a reasonable timeframe
> this should not block the project team from moving to the next steps,
> particularly if an issue is of high severity or impact.

The `security-issue-sync` skill tracks the reporter mail thread in detail but
has no time-based escalation. If a reporter goes silent after the initial report,
nothing in the framework prompts the security team to proceed without them. The
team must notice the staleness manually by reading the thread dates themselves,
and there is no proposal item that says "reporter has been silent for N days —
proceed?".

A grep across the entire repo for "unresponsive", "reasonable timeframe", and
"proceed without" returns zero matches in any skill, process doc, or roles doc.

## Required changes

In `security-issue-sync` Step 1c, after reading the reporter thread, add a
staleness check: if the last outbound message from the security team to the
reporter is older than a configurable threshold (suggested default: 14 days) and
no reply from the reporter has landed since, surface an explicit numbered
proposal item in Step 2b:

> "Reporter has not replied in N days — propose proceeding with fix and
> announcement without further reporter sign-off, per ASF policy."

Add a config key to `projects/_template/project.md` (e.g.
`reporter_response_timeout_days`, default 14) so adopters can tune the threshold
to their own norms.

Add a corresponding note to `docs/security/process.md` and
`docs/security/roles.md` documenting that reporter unresponsiveness does not
block the process and pointing to the configurable threshold.

## Policy reference

- https://www.apache.org/security/committers.html — step 11 under "Resolve"

harishkesavarao:
## Summary

Implements ASF security-committers policy: reporter unresponsiveness must not block the process.

- `skills/security-issue-sync/gather.md`: Step 1c now checks staleness. If the team's last message to the reporter is older than a configurable threshold with no reply since, the thread is flagged for Step 2b.
- `skills/security-issue-sync/signals-to-actions.md`: new proposal category emits the numbered item "Reporter has not replied in N days, propose proceeding with fix and announcement without further reporter sign-off, per ASF policy." Proposal only, never auto-applied. Reappears each sync pass while still stale.
- `projects/_template/project.md`: added reporter_response_timeout_days config key, default 14, under Security inbox.
- `docs/security/process.md`: note after Step 4 separating team-discussion stalls from reporter unresponsiveness, with a link to the policy and the config key.
- `docs/security/roles.md`: note in "Keeping the reporter informed" stating a silent reporter doesn't block progress, pointing to the same config key.

Ran skill-and-tool-validator. No new failures, only pre-existing warnings.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

## RFC-AI-0004 compliance

## Linked issues

Closes https://github.com/apache/magpie/issues/182


harishkesavarao:
@justinmclean , when you get a moment, could you please review the PR? Thank you in advance!

potiuk (approved):
This is a good change, and unusually well-evidenced for a first
contribution — I checked the policy quote against
`apache.org/security/committers.html` and it is verbatim-accurate on both
counts ("If the reporter is unresponsive in a reasonable timeframe this
should not block the project team from moving to the next steps,
particularly if an issue is of high severity or impact" and "The project
team agrees on the fix, the announcement, and the release schedule with
the reporter"). The negative case (`case-7`) is the part that makes the
positive case trustworthy: the staleness clock only runs while *we* are
the ones waiting, and that is now pinned by a fixture.

You also quietly fixed a pre-existing bug — `gather.md`'s 1c list had
**two** items numbered `5.`; the renumber makes it a clean 1–7.

I've pushed one fixup rather than send you round again for it.

### Config layering — the timeout belongs to the organization

`reporter_response_timeout_days` was declared in
`projects/_template/project.md` under `security_inbox:`, but that
section's own opening line says:

> Org-level except the concrete address: the inbox `kind`, the foundation
> security address, the `has_forwarder_relay` flag, and the list filter
> are inherited. Declare this project's inbound address:

And `AGENTS.md` § *Configuration resolution order*:

> A project declares only what differs from its organization; an
> organization declares only what differs from the framework baseline
> (`organizations/independent/` is that baseline).

A reporter-response timeout is a foundation-level default, not a
per-project inbound address. Left in the per-project block it would have
baked `14` into every adopter's copied manifest, so the framework could
never move the default — and it would have sat oddly next to the other
day-based knobs (`disclosure_governance.window_days`,
`grace_period_days`), which live in their own companion file with
`# TODO: adjust` markers.

So the fixup:

- declares `reporter_response_timeout_days: 14` in
  `organizations/ASF/organization.md` and
  `organizations/independent/organization.md`, alongside the rest of the
  `security_inbox` block;
- turns the `project.md` entry into a documented **optional override**,
  with your rationale text preserved — it was good, it just belongs in
  the override note rather than in a declared default;
- repoints the four prose references (`gather.md`,
  `signals-to-actions.md`, `docs/security/process.md`,
  `docs/security/roles.md`) at the configuration resolution order instead
  of at `project.md`, since that file no longer declares the value;
- corrects `tools/skill-evals/README.md`'s `security-issue-sync` count,
  which your two new cases take to 32.

Nothing else changed. `prek run --from-ref origin/main` is green,
including lychee on the new anchors, and all seven `step-2b` cases
assemble with `case-6` emitting `reporter_unresponsive` and `case-7`
emitting nothing.

One note for future PRs: the eval prompt for a step is composed from the
`SKILL.md` step section plus `output-spec.md` (`runner.py:194-198`) — not
from `signals-to-actions.md`. You got this right, and it is worth knowing
why it worked: putting the firing condition in `output-spec.md` is what
makes `case-6` a real test rather than an assertion the prompt never
motivates.

Thanks for taking the ASF-policy angle seriously enough to go read the
source and wire a config key for it.

---

> will follow up.*
>

PR #1078, opened 2026-08-10
feat(bitbucket): add guarded cloud issue comment create

KatalKavya96:
## Summary

* Adds the first narrowly scoped Bitbucket write operation: `magpie-bitbucket issue comment <id> --body-file <path>` for creating Bitbucket Cloud issue comments.
* Follows Magpie’s existing write-path discipline: the bridge executes an already-confirmed action, while explicit user confirmation remains the responsibility of the calling skill.
* Keeps Bitbucket Data Center native issue comment writes explicitly unsupported and directs users toward linked Jira coverage instead.

## Type of change

* [x] Tool / bridge contract (`tools/<system>/*.md`)
* [ ] Groovy reference impl
* [x] Cross-cutting (RFC, AGENTS.md, sandbox, privacy-LLM)
* [x] Documentation (`docs/`, `README.md`, `CONTRIBUTING.md`)
* [ ] Other:

## Test plan

* [x] For Python packages touched: `PYTHONPATH=src uv run --group dev pytest tests/test_bitbucket.py -q`
* [x] For Python packages touched: `uv run --group dev ruff check src tests --fix`
* [x] For Python packages touched: `uv run --group dev ruff format src tests`
* [x] For Python packages touched: `uv run --group dev mypy src tests`
* [x] Other: focused issue-comment creation tests pass and `git diff --check` is clean

## RFC-AI-0004 compliance

* [x] **HITL** — the bridge does not decide whether to mutate; the calling skill must obtain explicit user confirmation before invoking the write command
* [x] **Sandbox** — no new unrestricted host access; the existing Bitbucket HTTPS and redirect protections remain in place
* [x] **Vendor neutrality** — adds a narrow Bitbucket-specific tracker mutation without claiming complete `contract:tracker` or backend parity
* [x] **Conversational + correctable** — confirmation remains at the calling-skill boundary, consistent with existing Magpie write workflows
* [x] **Write-access discipline** — only the explicitly requested Cloud issue-comment mutation is executed; no autonomous comments or broader mutations are added
* [x] **Privacy LLM** — comment bodies and returned Bitbucket issue-comment content remain external data and follow the existing approved-LLM/privacy handling rules

## Linked issues

Refs #606

## Notes for reviewers

This is the first write operation in the Bitbucket bridge and intentionally keeps the scope small.

The new command is:

```bash
magpie-bitbucket issue comment <id> --body-file <path>
```

The write path follows the same discipline documented by the existing Jira bridge: the bridge executes the mutation, but confirmation is owned by the calling skill. There is deliberately no bridge-level `--confirm` flag.

Implementation details:

* Bitbucket Cloud only
* sends a JSON `POST` to the issue comments endpoint
* reads the comment body from `--body-file`
* rejects missing or whitespace-only body files before any outbound write
* reuses quoted path segments and authenticated HTTPS handling
* write requests reject redirects rather than replaying a mutation elsewhere
* normalizes the created comment through the existing issue-comment helpers
* Bitbucket Data Center returns an explicit unsupported error with linked Jira guidance

Tests cover the Cloud POST request/body, Data Center rejection, response normalization, CLI success, and missing/empty body failures before the mutation path is invoked.

All broader writes — issue updates, PR comments/reviews, task mutation, decline/merge, branch changes, and build mutations — remain out of scope for this PR.


KatalKavya96:
Hi @potiuk , I opened the next Bitbucket follow-up for #606.

This is the first narrowly scoped write path:

```bash
magpie-bitbucket issue comment <id> --body-file <path>
```

It follows the existing Magpie/Jira write discipline: explicit user confirmation stays in the calling skill, while the bridge only executes the already-confirmed mutation.

Scope is Cloud-only; Data Center native issue-comment writes remain explicitly unsupported with linked Jira guidance. Missing/empty body files fail before any outbound write, and write requests reuse the existing auth/HTTPS protections while refusing redirects.

Focused/full pytest, ruff, mypy, `git diff --check`, and `prek run --all-files` are passing.

potiuk (approved):
Approving. This is the first write operation in the Bitbucket bridge, so I
reviewed the mutation path rather than the feature, and it holds up — worth
recording what I actually checked.

**The security properties are real, not just claimed:**

- `post_json` calls `_require_https(url)` before building the request.
- Redirects are genuinely **rejected**, not followed:
  `NoAuthRedirectHandler.redirect_request` raises unconditionally, so the
  comment *"Writes never follow redirects: repeating a mutation at a
  redirected location is less safe than failing"* is accurate. It is also the
  stricter of the two handlers available — `get_text` uses
  `SameHostRedirectHandler`, `get_json` uses this one — and picking the
  stricter one for a mutation is the right call. A side effect worth noting:
  that error is raised from inside `opener.open`, which makes the
  `except BitbucketError: raise` clause load-bearing rather than decorative.
- No path-injection surface: `quote_path` is
  `urllib.parse.quote(value, safe="")`, so it escapes `/` as well — an
  `issue_id` of `../../admin` becomes `..%2F..%2Fadmin`. It is applied to
  workspace, repo slug, **and** issue id.
- `--body-file` with no inline `--body` alternative matches the framework's
  own body-file discipline and keeps contributor-authored text away from
  shell escaping.
- Data Center **raises explicitly** instead of silently no-opping, matching
  the existing `get_issue_comments` pattern in that module.
- I specifically checked whether an unexpected response shape could report
  `ok: false` on a write that actually succeeded — which would invite a
  double-post on retry. It cannot: `_cloud_issue_comment` returns an
  eleven-key dict literal, so it is never falsy.

**The tests are better than typical for a write path.** Beyond asserting URL,
method, content type and exact payload, the failure cases use
`mock_create_issue_comment.assert_not_called()` — proving the mutation is not
*attempted* when the body file is empty or missing, rather than only that an
error surfaces. That is the assertion that matters here.

**Bookkeeping is right too.** `labels-and-capabilities.md` moves the tool
from `partial-read-only` to `partial` *and* corrects the qualifier's own
explanatory sentence — dropping "read-only" from what the qualifier means is
easy to miss. `adapters.md` is updated in the same PR per the spec-sync rule
and now states the HITL boundary explicitly.

### One nit, for a follow-up rather than this PR

`tools/bitbucket/tests/test_bitbucket.py` has four thorough
`SameHostRedirectHandler` tests and none for `NoAuthRedirectHandler`, which
is now the guard on a mutation path. The behaviour is correct by
construction, and the gap is pre-existing — `get_json` already used that
handler — so this PR inherits it rather than introduces it. But "a write is
never replayed at a redirect target" is now a stated security property of
this bridge, and a one-line test would pin it. Not worth holding the merge.

Keeping the scope this small for the first write was the right instinct.

---

> will follow up.*
>

PR #1050, opened 2026-08-01
chore(ai-tutors): regenerate lesson-08 knowledge base

potiuk:
## Summary

- `inject-knowledge-base.py --check` reported `ai-tutors/lesson-08-eval-driven-development.md` as out of date on `main`. This is the regeneration; `--check` is now clean across all eleven prompts.
- The drift is real, not cosmetic: the tutor prompt embeds its source page verbatim, and #985 replaced worked example 2 with the actual `case-7` fixture — a `BashOperator` `KeyError` report whose body opens with a `SYSTEM:` injection. The prompt still described a dark-mode feature request carrying a different injection string. A tutor teaching prompt-injection defence from an example that no longer exists in the fixtures is the wrong thing to hand a learner.
- Found while closing out #1010 / #1011 (nested code fences). **Nothing in this diff is fence-related** — I verified both `lesson-04` and `lesson-08` were already stale on `main` *before* that work, so this is independent drift and was deliberately kept out of #1010 rather than bundled in.

## Type of change

- [ ] Skill change (`.claude/skills/<name>/`) — eval fixtures updated below
- [ ] Tool / bridge contract (`tools/<system>/*.md`)
- [ ] Python package (`tools/*/` with `pyproject.toml`)
- [ ] Groovy reference impl
- [ ] Cross-cutting (RFC, AGENTS.md, sandbox, privacy-LLM)
- [x] Documentation (`docs/`, `README.md`, `CONTRIBUTING.md`) — generated tutor prompt
- [ ] Project template (`projects/_template/`)
- [ ] CI / dev loop (`prek`, workflows, validators)
- [x] Other: regenerated artefact — `python3 ai-tutors/inject-knowledge-base.py` output, no hand edits

## Test plan

- [x] `prek run` passes on the changed file (markdownlint, lychee, typos, SPDX, doctoc)
- [x] `python3 ai-tutors/inject-knowledge-base.py --check` exits clean — before this change it listed `lesson-08`
- [ ] For Python packages touched — *none*
- [ ] For Groovy bridges touched — *n/a*
- [ ] For skill changes — *no skill changed*
- [x] Other: confirmed the diff contains no code-fence changes, i.e. this is a content refresh rather than an artefact of the recent `tag_bare_code_fences` fix

## RFC-AI-0004 compliance

- [ ] **HITL** — any new mutation is gated on explicit user confirmation
- [ ] **Sandbox** — no new unrestricted host access; network reach declared in the adapter
[...]

PR #1080, opened 2026-08-10
ci(ai-tutors): add prek hook enforcing inject-knowledge-base.py --check

### LINKED ISSUE #1051: ci(ai-tutors): nothing enforces inject-knowledge-base.py --check, so tutor prompts drift silently
## What

Nothing runs `python3 ai-tutors/inject-knowledge-base.py --check`. No CI job, no prek hook. A source-page edit under `docs/education/` can therefore leave the generated tutor prompt in `ai-tutors/` silently stale, and nothing surfaces it.

## Why it matters

This is not hypothetical — both generated prompts had drifted before anyone noticed:

- `lesson-04-your-first-skill.md` was stale against two source-page changes (`/write-skill` → `/magpie-write-skill`, and a changed capability-label set). Fixed incidentally by #1010.
- `lesson-08-eval-driven-development.md` was stale against #985, which replaced worked example 2 with the real `case-7` fixture. The prompt still taught prompt-injection defence using a *dark-mode feature request* example that no longer exists in the fixtures. Fixed by #1050.

Neither was caught by review. Both were found only because someone happened to run `--check` by hand while working on something else. The failure mode is quiet and the content is learner-facing, which is a bad combination: a tutor teaching from an example that no longer exists is worse than one that is merely out of date.

It also produced a near-miss. #928 regenerated `lesson-04` *before* those source-page changes landed, so by the time it was looked at, merging it would have rolled the file **backwards**. It was closed rather than merged, but only because someone diffed it against `main` first.

## Suggested fix

`ai-tutors/` became a uv workspace member in #1011, so it already has a `pyproject.toml`, a dev dependency group, and a `pytest (ai-tutors)` CI job. Adding the check is cheap. Either:

- a prek hook running `inject-knowledge-base.py --check`, scoped to `files: ^(ai-tutors/|docs/education/)`, so it fires when a source page or a prompt changes; or
- a pytest case in `ai-tutors/` asserting `--check` reports nothing, which reuses the job that already exists.

The hook is probably the better fit: the failure is "you edited a source page and forgot to regenerate", and a hook can say exactly that at commit time.

The script side needs no work — `--check` already prints the drifting prompts and exits non-zero, so it is ready to be wired in as-is. The gap is purely that nothing calls it.

dpol1:
## Summary

- Nothing ran `python3 ai-tutors/inject-knowledge-base.py --check`, so a
  `docs/education/` source-page edit could leave the generated tutor prompts
  in `ai-tutors/` silently stale — both shipped prompts had already drifted
  this way before being noticed (see the issue for the history).
- Wire `--check` as a local prek hook scoped to
  `^(ai-tutors/|docs/education/)`, mirroring the `vendor-neutrality-score`
  hook's generated-content pattern (own `repo: local` block,
  `pass_filenames: false`, explanatory comment). The script itself is
  untouched — `--check` already prints the drifting prompts and exits
  non-zero.
- CI already runs `prek run --all-files`, so the one config change gates both
  commit time and CI.

## Type of change

## Test plan

- [x] Red case: with the drift currently live on `main`
      (`ai-tutors/lesson-08-eval-driven-development.md`), the hook fails and
      names the stale file.
- [x] Green case: after regenerating lesson-08 locally, the hook passes.
      (The regeneration itself is not part of this PR — see the ordering note
      below.)
- [x] `prek run --files .pre-commit-config.yaml` passes, including the
      "Check if all hooks apply to the repository" gate that validates the
      new `files:` pattern.

## RFC-AI-0004 compliance

CI-only; no principles touched.

## Linked issues

Closes [apache/magpie#1051](https://github.com/apache/magpie/issues/1051)

## Notes for reviewers

**Ordering: blocked by [apache/magpie#1050](https://github.com/apache/magpie/pull/1050).**
`main` currently has a live lesson-08 drift, which #1050 fixes. Until that
merges, this PR's own CI will fail on exactly the new hook — a live
demonstration that the gate works, but it means this PR should merge
*after* #1050 (and then rebase/re-run to go green). Regenerating lesson-08
here instead would duplicate #1050, so this PR deliberately stays hook-only.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -204,6 +204,23 @@ repos:
         entry: uv run --project tools/vendor-neutrality-score vendor-neutrality-score --in-place
         files: ^(tools/[^/]+/README\.md|tools/privacy-llm/models\.md|tools/vendor-neutrality-score/.*|skills/[^/]+/SKILL\.md|docs/vendor-neutrality\.md)$
         pass_filenames: false
+  # AI-tutor knowledge-base sync — fail when a `docs/education/` source
+  # page (or a training lesson) changes without regenerating the tutor
+  # prompts in `ai-tutors/`. `--check` prints the stale prompts and exits
+  # non-zero; the contributor reruns `python3
+  # ai-tutors/inject-knowledge-base.py` (no flag) to regenerate and
+  # re-stages. `pass_filenames: false`: the script always scans every
+  # lesson; the `files:` filter only gates *whether* the check fires (and
+  # it always runs on `prek run --all-files` in CI). Deterministic and
+  # offline. Script + pytest suite live in ai-tutors/.
+  - repo: local
+    hooks:
+      - id: ai-tutors-kb-check
+        name: Check ai-tutors prompts are in sync with docs/education
+        language: system
+        entry: python3 ai-tutors/inject-knowledge-base.py --check
+        files: ^(ai-tutors/|docs/education/)
+        pass_filenames: false
   # Self-adoption symlink hygiene: reject cyclic symlinks and misdirected
   # skill relays. `types: [symlink]` fires it on any staged symlink (always
   # on `prek run --all-files`). Rules, rationale, and the pytest suite live


potiuk (approved):
Approving. Nothing was wrong with this PR — the red `prek` was the new hook
correctly doing its job, and the drift it found was real.

### What was actually failing

The only failing hook on the previous head was the one this PR adds:

```
Check ai-tutors prompts are in sync with docs/education...........Failed
```

I ran `python3 ai-tutors/inject-knowledge-base.py --check` against `main` to
see whether that was a bug in the hook or a genuine finding. It was genuine,
and precisely scoped:

```
1 AI tutor prompt(s) need knowledge-base refresh.
ai-tutors/lesson-08-eval-driven-development.md
```

Exactly one prompt, and it was already fixed in a separate open PR (#1050,
regenerating that lesson's knowledge base). So this PR was blocked behind a
regeneration rather than containing any defect — which is the best possible
outcome for a hook whose entire purpose is to catch that class of drift. It
found the drift before it was enforced.

### What I did

Merged #1050 so `main` is back in sync, then rebased this branch onto it
(cherry-picked; your authorship is preserved on `ff0599ab`). Verified on the
rebased branch:

- `python3 ai-tutors/inject-knowledge-base.py --check` exits clean with no
  stale prompts.
- `prek run ai-tutors-kb-check --all-files` passes. That is the mode that
  matters: the hook's `files:` filter is `^(ai-tutors/|docs/education/)`, and
  this PR only touches `.pre-commit-config.yaml`, so a `--from-ref` run would
  not fire it at all. Running it explicitly over the whole tree confirms it
  is wired up and actually executes in the CI path rather than silently
  never matching.

### On the hook itself

The `pass_filenames: false` choice is right and the comment explains why —
the script always scans every lesson, so the `files:` filter gates *whether*
the check fires, not what it inspects. Pointing the entry at `--check` rather
than the regenerating invocation is also the right call: a hook that silently
rewrites tracked content on a contributor's machine is a worse experience
than one that fails and tells them which command to run.

Thanks for closing this gap — a generated artefact with no enforcement is
exactly the kind of thing that drifts quietly for months.

---

> will follow up.*
>

PR #1088, opened 2026-08-13
fix(pr-management-triage): filter action_required runs server-side

potiuk:
`fetch-and-batch.md` builds the repo-wide `action_required` index from an
unfiltered `event=pull_request` listing, post-filtered on `conclusion` and
capped at the first three pages, justified by the claim that
`?status=action_required` *"matches no runs in this state and silently returns
an empty result"*.

The claim is incorrect, and the framework already contradicts itself —
`SKILL.md` Golden rule 1b uses `?status=action_required&head_sha=<SHA>`.

## Measured impact

Full triage sweep against a large adopter repo (2026-08):

| approach | pending head SHAs found |
|---|---|
| `?status=action_required`, paginated | **322** |
| documented 3-page `event=pull_request` post-filter | **9** |

`?status=action_required` reported 1109 matching runs. The 3-page window
reached back only ~12 hours, because that repo produces ~300 `pull_request`
runs in that time. Every PR in the ~313-SHA gap whose rollup was green from bot
checks alone (`Mergeable`, `WIP`, `DCO`) classified as `passing` and became a
`mark-ready` candidate — the precise false-positive class this index exists to
prevent, and the one Golden rule 1b is written to stop.

## Changes

- `fetch-and-batch.md`: filter server-side, walk every page, and note the
  1000-result listing cap with the `created:`-window workaround. The incorrect
  rationale is replaced with the measured one.
- `actions.md`: the two per-PR lookups are scoped to a single head SHA and were
  never affected by the truncation. Their code is unchanged; only the
  inaccurate claim in the comments is corrected, with a pointer to why the
  repo-wide case differs.

A `conclusion` post-filter over the *filtered* set is kept as belt-and-braces,
and stays correct for a host whose API build ignores `status`.

---

- [X] Yes — Claude Code (Opus 5)

Generated-by: Claude Code (Opus 5)

### DIFF
--- a/skills/pr-management-triage/actions.md
+++ b/skills/pr-management-triage/actions.md
@@ -359,13 +359,15 @@ executed.
 ```bash
 # Pre-check: index action_required runs at the head SHA.
 # Note: runs awaiting approval are returned as `status: "completed"`
-# with `conclusion: "action_required"`. The query parameter
-# `?status=action_required` matches no runs and would silently
-# return an empty result — post-filter on `conclusion` instead.
+# with `conclusion: "action_required"`. This lookup is already scoped
+# to one head SHA, so the `conclusion` post-filter is sufficient and
+# needs no `status=` narrowing. (For the repo-wide index in
+# fetch-and-batch.md the opposite holds: filter with
+# `status=action_required` server-side, because an unfiltered listing
+# truncates long before the backlog ends.)
 # One fetch covers both guards.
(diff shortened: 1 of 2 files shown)


choo121600 (dismissed):
good

potiuk:
Hey @choo121600 - can you re-approve this one :) ?

choo121600 (approved):

PR #1099, opened 2026-08-18
fix(tools): scope bridge.groovy config vars with @Field

### LINKED ISSUE #1098: bug: bridge.groovy throws MissingPropertyException — top-level def config vars invisible to cmd_* methods
### What's broken

Every `bridge.groovy` subcommand fails at runtime with `groovy.lang.MissingPropertyException: No such property: TRACKER_URL for class: bridge`. The bridge's configuration variables (`TRACKER_URL`, `API_TOKEN`, `AUTH_SCHEME`, `ENV`) are declared as top-level `def` locals, so the `cmd_*` methods and the `httpGet`/`httpPost` helpers cannot see them.

### Which layer

[`tools/jira/bridge.groovy`](https://github.com/apache/magpie/blob/main/tools/jira/bridge.groovy) lines 40–44:

```groovy
def ENV = System.getenv()
def TRACKER_URL  = ENV['ISSUE_TRACKER_URL'] ?: ''
def PROJECT_KEY  = ENV['ISSUE_TRACKER_PROJECT'] ?: ''
def API_TOKEN    = ENV['JIRA_API_TOKEN'] ?: ''
def AUTH_SCHEME  = ENV['JIRA_AUTH_SCHEME'] ?: 'Basic'
```

### How to reproduce

```bash
ISSUE_TRACKER_URL=https://issues.apache.org/jira \
  groovy tools/jira/bridge.groovy search 'project = FOO'
```

Result:

```
Caught: groovy.lang.MissingPropertyException: No such property: TRACKER_URL for class: bridge
	at bridge.cmd_search(bridge.groovy:182)
	at bridge.run(bridge.groovy:372)
```

The same failure hits every read and write subcommand. It also surfaces in CI: `tools/jira/tests/test_bridge_write.py` fails (`prek run --all-files`), because the test harness shells out to the real bridge.

### Expected vs actual

- **Expected:** the bridge resolves its configuration from the environment and runs the requested subcommand.
- **Actual:** it throws `MissingPropertyException` before any HTTP call.

**Root cause:** in a Groovy script, a bare top-level `def x = …` is a *local variable of the generated `run()` method*, not an entry in the script binding — so it is invisible inside methods compiled onto the script class. `TRACKER_URL` is merely the first such variable referenced (when building the request URL); `API_TOKEN` and `AUTH_SCHEME` fail the same way once that one is fixed. The top-level `if (!TRACKER_URL)` guard passes because it runs inside `run()`, which masks the problem until a `cmd_*` method is entered.

**Fix:** annotate the script-level config variables with `@groovy.transform.Field` (via `import groovy.transform.Field`) so they become fields visible to every method.

### Surface area (optional)

- `PROJECT_KEY` is affected by the same scoping issue but is currently dead code (defined, never read), so it has no runtime impact today — promoting it to `@Field` alongside the others keeps things consistent.
- The other Groovy tool, [`tools/dashboard-generator/reference.groovy`](https://github.com/apache/magpie/blob/main/tools/dashboard-generator/reference.groovy), is **not** affected: it defines no methods (linear script + closures, which do capture the script scope) and reads no environment variables.

milamberspace:
## Summary

- Every `bridge.groovy` subcommand threw `groovy.lang.MissingPropertyException: No such property: TRACKER_URL for class: bridge` before making any HTTP call.
- Root cause: the script-level config vars (`ENV`, `TRACKER_URL`, `API_TOKEN`, `AUTH_SCHEME`) were declared with a bare top-level `def`. In a Groovy script that makes them locals of the generated `run()` method, so the `cmd_*` methods and the `httpGet`/`httpPost` helpers — compiled onto the script class — cannot see them. The `if (!TRACKER_URL)` guard passed because it runs inside `run()`, masking the bug until a method was entered.
- Fix: annotate them with `@groovy.transform.Field` (via `import groovy.transform.Field`) so they become fields visible to every method. Added a short comment explaining the `@Field` requirement.

## Type of change

- [x] Groovy reference impl

## Test plan

- [x] `prek run --all-files` passes — including `pytest (workspace)`, which runs `tools/jira/tests/test_bridge_write.py` (the suite that was failing before this change).
- [x] For Groovy bridges touched: command-line invocation tested end-to-end — `groovy tools/jira/bridge.groovy search 'project = FOO'` now reaches the HTTP layer instead of throwing `MissingPropertyException`; the missing-`ISSUE_TRACKER_URL` guard still exits 2.

Before:

```
Caught: groovy.lang.MissingPropertyException: No such property: TRACKER_URL for class: bridge
	at bridge.cmd_search(bridge.groovy:182)
```

After: 28 passed in `tools/jira/tests/test_bridge_write.py`.

## RFC-AI-0004 compliance

- **Write-access discipline** — unchanged; the bridge still only executes confirmed mutations, this fix only makes existing config resolution work. No other principles touched.

## Linked issues

Closes #1098

- `PROJECT_KEY` is promoted to `@Field` too for consistency, though it is currently dead code (defined, never read) — no runtime impact either way.
- The other Groovy tool, `tools/dashboard-generator/reference.groovy`, is not affected: it defines no methods (linear script + closures, which capture the script scope) and reads no environment variables.

### DIFF
--- a/tools/jira/bridge.groovy
+++ b/tools/jira/bridge.groovy
@@ -4,6 +4,7 @@
 @Grab('org.apache.groovy:groovy-json:4.0.21')
 import groovy.json.JsonOutput
 import groovy.json.JsonSlurper
+import groovy.transform.Field
 
 /**
  * JIRA REST bridge for the issue-* skill family.
@@ -37,11 +38,15 @@ import groovy.json.JsonSlurper
  * the calling skill — the bridge only executes confirmed actions.
  */
 
-def ENV = System.getenv()
-def TRACKER_URL  = ENV['ISSUE_TRACKER_URL'] ?: ''
-def PROJECT_KEY  = ENV['ISSUE_TRACKER_PROJECT'] ?: ''
-def API_TOKEN    = ENV['JIRA_API_TOKEN'] ?: ''
-def AUTH_SCHEME  = ENV['JIRA_AUTH_SCHEME'] ?: 'Basic'
+// These are @Field so the cmd_* methods and the httpGet/httpPost helpers
+// can see them. A bare top-level `def` in a Groovy script is a local of the
+// generated run() method and is NOT visible inside methods — referencing it
+// there throws MissingPropertyException.
+@Field Map ENV = System.getenv()
+@Field String TRACKER_URL  = ENV['ISSUE_TRACKER_URL'] ?: ''
+@Field String PROJECT_KEY  = ENV['ISSUE_TRACKER_PROJECT'] ?: ''
+@Field String API_TOKEN    = ENV['JIRA_API_TOKEN'] ?: ''
+@Field String AUTH_SCHEME  = ENV['JIRA_AUTH_SCHEME'] ?: 'Basic'
 
 if (!TRACKER_URL) {
     System.err.println('error: ISSUE_TRACKER_URL not set in the environment (the calling skill resolves it from <project-config>/issue-tracker-config.md and exports it)')


dpol1 (approved):
Approving — the bug is real and reproducible from the source on `main`, the `@Field` conversion is complete, and the test evidence in the PR body checks out against the tree. One naming nit below, not blocking.

### The bug is confirmed on `main`

`httpGet` (line 51), `httpWrite` (70), `httpMultipart` (96) and the `cmd_*` methods all reference the script-level config vars, which as bare top-level `def`s are locals of the generated `run()` method and invisible from methods compiled onto the script class. The stack line quoted in the PR body — `MissingPropertyException at bridge.cmd_search(bridge.groovy:182)` — matches exactly the `${TRACKER_URL}` interpolation at [line 182 on `main`](https://github.com/apache/magpie/blob/main/tools/jira/bridge.groovy#L182). The `@Field` conversion covers all five vars, no other script-level state remains in the file, and the early `if (!TRACKER_URL)` guard keeps working since fields stay visible from `run()`.

### Smaller observations

- `tools/jira/bridge.groovy` (new comment above the `@Field` block) — the comment says "the `httpGet`/`httpPost` helpers", but no `httpPost` exists in this file; the write-path helpers are `httpWrite` and `httpMultipart`. Worth naming the real methods so the comment doesn't point at a phantom identifier.

---

> *This review was drafted by an AI-assisted tool and checked and
> submitted by an Apache Magpie contributor — not a maintainer, so
> a maintainer's review remains the authoritative next step. If
> something feels off, please reply on the PR.*
>

potiuk (approved):

PR #913, opened 2026-07-19
feat(setup): install/uninstall primary commands; adopt/unadopt aliases

potiuk:
Now that Magpie can be installed from agent marketplaces (#907), **"install" is the more familiar term than "adopt."** This makes `install`/`uninstall` the primary command surface, with `adopt`/`unadopt` kept as **aliases** (nothing breaks for anyone using them).

## Changes
- `/magpie-setup install` — first-time install (the default no-arg action); `/magpie-setup adopt` is a documented **alias**.
- `/magpie-setup uninstall` — reverses it; `/magpie-setup unadopt` is a documented **alias**.
- Renamed `skills/setup/adopt.md` → `install.md`, `skills/setup/unadopt.md` → `uninstall.md`, and `docs/setup/unadopt.md` → `uninstall.md`; all path references repointed.
- Updated the skill's `argument-hint`, sub-action list, and command table.

## Deliberately kept
- The **"adopter" / "adopter repo"** noun (the ecosystem role) — unchanged.
- The **"Adopt a Magpie"** landing-page slogan.
- Deep internal prose still uses "adopt" informally (a valid alias) — the full verb reword is **PR B** (follow-up).

Scope: command surface + file renames only. Part of the adoption→installation terminology shift discussed on the list.

PR #1115, opened 2026-08-26
fix(setup): wire the agent-guard hook into settings.local.json, not settings.json

### LINKED ISSUE #824: setup: write the agent-guard hook to gitignored settings.local.json instead of committing it to settings.json
## Problem

Today the adopt flow commits the agent-guard PreToolUse hook wiring into the
tracked `.claude/settings.json`, while the script it invokes
(`.claude/hooks/agent-guard.py`) stays **gitignored** and is only synced in by
`/magpie-setup`. That split has caused a chain of follow-up fixes:

- **#786** — a fresh clone has the committed wiring but not yet the script, so
  the hook command had to be guarded (`[ -f … ] && python3 … || true`) to
  no-op when the script is absent.
- **#823** — that guard is a POSIX-shell conditional, which breaks on Windows
  (PowerShell, no Git Bash), so it had to be rewritten as an inline
  `python3 -c` existence check.

Both are workarounds for the same root cause: **we commit machine-specific hook
wiring for a script we deliberately don't commit.** It also enlarges the
always-committed footprint for an opt-in feature.

## Proposal

Have `/magpie-setup` write the hook entry into the **gitignored, per-machine**
`.claude/settings.local.json` at the same moment it deposits
`agent-guard.py` — and stop committing it in `.claude/settings.json`.

Because the wiring then only ever exists once the script is present:

- **No fresh-clone guard needed** (#786 becomes moot) — the command can be a
  plain `python3 "$CLAUDE_PROJECT_DIR/.claude/hooks/agent-guard.py"`.
- **No cross-platform shell gymnastics needed** (#823 becomes moot) — plain
  `python3 "…"` runs on POSIX and Windows alike.
- **Zero committed footprint** for the hook — nothing tracked in the adopter
  repo, matching the fact that both the script and its trigger are
  per-machine framework artefacts, not project config.

`.claude/settings.local.json` is Claude Code's documented per-user override
file and is already the natural home for machine-local settings; adopters
gitignore it (as this repo's `.gitignore` already does for `settings.local`).

## Scope of the framework change

- `skills/setup/adopt.md` — Step 12 (and the verify step): write/merge the
  hook block into `.claude/settings.local.json` instead of surfacing a
  `settings.json` snippet for the maintainer to commit; drop the `-f`/guard.
- `.gitignore` template — ensure `/.claude/settings.local.json` is ignored.
- `tools/agent-guard/README.md` — document the `settings.local.json` wiring
  and the plain command.
- unadopt/verify flows — read/clean `settings.local.json`.
- Idempotent merge: don't clobber an adopter's existing
  `settings.local.json` keys.

## Context

Raised from Apache JMeter's adoption review. The JMeter PMC reviewer (@vlsi)
independently asked "*Can magpie add the hook entry to
`.claude/settings.local.json` on setup?*" — this issue is that request,
generalised. JMeter has already dropped the committed `.claude/settings.json`
from its adoption PR in anticipation.

If accepted, this **supersedes #823** (and retro-justifies removing the #786
guard).

AmirF194:
## Summary

- The adopt flow commits the agent-guard PreToolUse wiring into the tracked `.claude/settings.json` while the script it invokes stays gitignored, which forced two prior workarounds for the same root cause (#786, #823).
- Moves the wiring into the gitignored, per-machine `.claude/settings.local.json`, written at the same moment the script is deposited, so the fresh-clone race that both workarounds guarded against no longer exists.
- Supersedes #823 and retro-justifies removing the #786 guard, per the issue.

## Type of change

- [ ] Skill change (`.claude/skills/<name>/`): eval fixtures updated below
- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] Other: ran the individual hooks by hand rather than `prek run --all-files` (installing `prek`'s full tool set was not practical here): `markdownlint-cli2` and `typos` on all 6 changed files (clean); `skill-and-tool-validate` run directly with `PYTHONPATH` against the stdlib-only validator package (0 hard failures, same 29 pre-existing soft warnings on unrelated skills, none on the touched files). Could not run the `lychee` link-check hook (its `cargo install` build ran the local disk near empty and was aborted); checked by hand instead that every link this diff adds or edits is an in-repo anchor (`worktree-init.md#step-1d-...`, `adopt.md#step-10-...`, `adopt.md#step-12-...`, `../../tools/agent-guard/README.md`) and confirmed each target heading/file exists on this branch. Also extracted the diff's own idempotent-merge Python and ran it standalone against five fixtures (missing file, unrelated existing keys, already-wired no-op, hand-corrupted JSON, a second pre-existing `Bash` matcher) to confirm the merge behaves as described in each case. Checked `docs/vendor-neutrality.md`'s inputs (`**Vendor:**` / `**Harness:**` fields in `tools/agent-guard/README.md`) are untouched by this diff, so the `vendor-neutrality-score` hook has nothing to regenerate.

## RFC-AI-0004 compliance

- [ ] **HITL**: any new mutation is gated on explicit user confirmation
- [ ] **Sandbox**: no new unrestricted host access; network reach declared in the adapter
- [x] **Vendor neutrality**: placeholders (`<PROJECT>`, `<tracker>`, `<upstream>`, `<security-list>`) used in all skill / tool prose (the `check-placeholders` prek hook is the mechanical gate)
- [ ] **Conversational + correctable**: agentic-override path documented if behaviour is adopter-tunable
- [ ] **Write-access discipline**: no autonomous outbound messages; drafts only, sent on confirmation
- [ ] **Privacy LLM**: private content does not reach a non-approved LLM; redactor invoked where needed

## Linked issues

Fixes #824

## Notes for reviewers

`.gitignore` already lists `/.claude/settings.local.json` as a base entry (Step 7), so nothing there needed to change; only the destination the docs point `/magpie-setup` at moved.

### DIFF
--- a/skills/setup/install.md
+++ b/skills/setup/install.md
@@ -630,13 +630,14 @@ it is framework code synced from the snapshot
 not an adopter artefact, so it is regenerated on `/magpie-setup`
 rather than committed (the same rule that keeps the snapshot and
 the framework-skill symlinks out of git — the only committed
-framework artefact is `magpie-setup`). The committed
-`.claude/settings.json` still references it; the **wiring** is
-committed, the **script** is not. Because the script is gitignored,
-**no** worktree inherits it via git — every worktree is seeded from
-the main checkout by the post-checkout hook
+framework artefact is `magpie-setup`). The `hooks.PreToolUse`
+**wiring** that references it also lives in `.claude/settings.local.json`
+(above), gitignored alongside the script itself. Neither is
+committed, so no worktree inherits either via git. Every worktree is
+seeded from the main checkout by the post-checkout hook
 ([Step 10](#step-10--worktree-aware-post-checkout-hook-fresh-only))
-or by [`worktree-init` Step 1d](worktree-init.md#step-1d--seed-the-worktrees-agent-guard-pretooluse-hook).
+or by [`worktree-init` Step 1d](worktree-init.md#step-1d--seed-the-worktrees-agent-guard-pretooluse-hook),
+which seed the wiring together with the script.
 An adopter who keeps their own non-framework guards under
 `.claude/hooks/guards.d/` and wants them tracked should commit them
 with `git add -f` (the directory is ignored by default).
@@ -681,9 +682,11 @@ framework skill.
 
 `.claude/settings.local.json` is the project-local
 per-machine settings file that
+[Step 12 pass 1](#step-12--post-install-sync--worktree-propagation--sandbox-allowlist--sanity-check)
+populates with the agent-guard `hooks.PreToolUse` wiring and that
 [Step 12 pass 3](#step-12--post-install-sync--worktree-propagation--sandbox-allowlist--sanity-check)
-populates with the project-root sandbox-allowlist entry (and
-that each worktree carries independently). Most adopters
+separately populates with the project-root sandbox-allowlist entry
+(and that each worktree carries independently). Most adopters
 already gitignore this file by Claude Code convention; the
 install flow checks for the line and adds it if missing.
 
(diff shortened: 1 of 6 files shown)


AmirF194:
Flagging for whoever reviews this: #913 also touches the same "Hook wired in settings.json" bullet in `verify.md` (and a nearby line in `worktree-init.md`), as part of its broader adopt/install rename. Its edit there is a link-text rename only, not a fix for the issue this PR addresses, so I do not think it is a competing claim, but the two will conflict on merge and whichever lands second will need a rebase.

potiuk (approved):

AmirF194:
Thanks for the review. The red `ruff format (workspace)` check flags three files this PR does not touch (contributor-nomination/fetch.md, contributor-sentiment/SKILL.md, issue-reproducer/verification.md), so it looks like pre-existing drift rather than something introduced here. Happy to rebase once #1092 settles if that clears it.

AmirF194:
Rebased onto main to resolve the conflict with #913 flagged above. Both bullets now point at install.md (the renamed file) while keeping the settings.local.json content from this PR. No other changes.

AmirF194:
Found and fixed a leftover: the #913 rebase repointed two of the three `adopt.md` links in worktree-init.md to `install.md`, but missed the one inside the settings.local.json wiring bullet this PR adds. That was the actual cause of the `prek` and `validate skills and tool contracts` CI failures, not the pre-existing skill-line-limit warnings. Pushed 8b71010.

potiuk (approved):

AmirF194:
Thanks for reviewing and merging this, appreciate the second look after the rebase fixup.

PR #1116, opened 2026-08-26
chore(asf.yaml): refresh collaborators list and keep approvals through pushes

potiuk:
## Summary

- `collaborators:` is trimmed to the non-ASF-members named in MISSION.md —
  `johnslavik`, plus `shahar1`, `vincbeck` and `bugraoz93`, the three Airflow
  committers from that section, now active entries rather than commented-out
  records of intent. `choo121600` and the eight non-Airflow-committer ASF
  members from the PMC roster are removed. The list goes from 9 entries to 4,
  well under ASF Infra's cap of 10.
- `dismiss_stale_reviews` flips to `false`, so an approval survives later pushes
  to the branch: a rebase or a review-comment fixup no longer forces a
  re-approval round-trip.
- Both surrounding comment blocks are rewritten to describe the values that are
  actually in the file — they previously contradicted the config.

## Type of change

## Test plan

- [x] `prek run --files .asf.yaml` passes
- [x] `.asf.yaml` parses as valid YAML
- No code paths touched; the effective change lands when asfyaml applies the
  file on merge to `main`.

## RFC-AI-0004 compliance

Not applicable — repository configuration only, no skill, tool, or agent
behaviour changes.

## Notes for reviewers

The dropped PMC-roster members lose the explicit triage grant this file gave
them; those with `apache` GitHub org access keep reach through that route.
Worth a second pair of eyes on whether that trade is intended before merge.

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -54,29 +54,11 @@ github:
     # Non-ASF-members named in MISSION.md → "involved as
     # collaborators on the project from day one":
     - johnslavik         # Bartosz Sławecki
-    # The four below are listed in MISSION.md's non-ASF-members
-    # section but are already Apache Airflow committers, so they
-    # have triage rights via the apache GitHub org through their
-    # Airflow committer status — no `collaborators:` slot needed.
-    # Kept here as a record of intent; uncomment if and when their
-    # committer status changes.
-    # - choo121600       # Yeonguk Choo (Airflow committer)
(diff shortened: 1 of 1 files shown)


choo121600 (approved):

potiuk:
And this one to fix the ruff check https://github.com/apache/magpie/pull/1117

PR #818, opened 2026-07-11
infra(.asf.yaml): require one approving review on main

potiuk:
Now that Apache Magpie is an established Top-Level Project with its own PMC, this adds branch protection requiring **at least one approving review from a committer** before any PR merges into `main`, replacing the bootstrap posture where a maintainer could self-merge after CI green.

## Change
```yaml
protected_branches:
  main:
    required_pull_request_reviews:
      required_approving_review_count: 1
      dismiss_stale_reviews: true
```

- **`dismiss_stale_reviews: true`** — a new push after approval dismisses the stale approval, so the approving reviewer signs off on the final state of the diff.
- **`require_code_owner_reviews`** is left off — there is no CODEOWNERS file, so any committer's approval satisfies the requirement.

## Notes
- Split out from the status/docs PR #817 so the governance/infra change can be reviewed on its own. The two touch adjacent `.asf.yaml` comment lines, so whichever merges second will need a trivial rebase.
- ASF Infra reconciles `protected_branches` within a few minutes of merge to `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1120, opened 2026-08-29
infra(.asf.yaml): temporarily lift the approval requirement on main

potiuk:
## What

Comments out the `required_pull_request_reviews` block in the `main`
protected-branch config in `.asf.yaml`, so a PR into `main` no longer needs an
approving review from a second committer before it can merge.

## Why

The marketplace-installation replacement and the reworked permission handling
are both being brought up now, and both land as many small, fast iterations
that have to be tested against a real merged `main` rather than against a
branch. A per-PR approval round-trip serialises that loop and stalls it for
hours at a time.

## What is *not* changing

This is a deliberate, time-boxed exception, not a governance change. Every
other gate on `main` stays exactly as it is:

- required status checks — `zizmor`, `prek`, `tests-ok` (`strict: false`)
- `required_conversation_resolution: true`
- `required_linear_history: true`
- `required_signatures: false`

Nothing merges unreviewed by CI. What comes off is only the second-pair-of-eyes
requirement — the part that serialises the loop.

## Restoring

The block is commented out rather than deleted, so restoring it is a pure
uncomment (remove the three leading `# ` markers). The rationale and the
restore instruction are both written into `.asf.yaml` next to the block, so the
exception documents its own expiry. ASF Infra reconciles the change within a
few minutes of the merge to `main`.

Supersedes the arrangement from #818 for the duration of that work.

https://claude.ai/code/session_01WNKZGHBjQqMHcWwFc1dxmd

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -92,17 +92,30 @@ github:
   # ever existed, so the protection now lives here next to the rest
   # of the repo config.
   #
-  # ONE APPROVING REVIEW REQUIRED ON `main`
-  # ---------------------------------------
+  # APPROVAL REQUIREMENT TEMPORARILY LIFTED ON `main`
+  # -------------------------------------------------
   # During the framework's bootstrap phase (running under the Airflow
   # PMC umbrella with a small set of committers, see MISSION.md) PR
   # approvals were NOT required and a maintainer could self-merge
-  # after CI green. Now that Magpie is a Top-Level Project with its
-  # own PMC, every PR into `main` requires at least one approving
-  # review from a committer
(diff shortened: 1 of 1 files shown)


potiuk:
I would really love to complete implementation and testing of the two **structural** improvements -> marketplace and permission confirrmation avoidance, and I will need quick main iteration for that one, so I'd appreciate approving this one to temporarily get things iterated faster :)

potiuk:
Temporarily switching to Commit-Then-Review basically.

choo121600 (approved):

eladkal (approved):

PR #1108, opened 2026-08-21
docs(guidelines): adopt semantic line breaks (SemBr) across repository

onlyarnav:
## Summary

- Adopts the Semantic Line Breaks (SemBr) standard across the repository's guidelines and templates.
- Adds rules and rationale in `AGENTS.md` and `docs/editorial-guidelines.md` requiring one-sentence-per-line formatting for clear, minimal git diffs.
- Links this one-statement-per-line philosophy in `CONTRIBUTING.md` (English as code) and `skills/write-skill/SKILL.md` (skill authoring guidelines).

## Type of change

- [x] Cross-cutting (RFC, AGENTS.md, sandbox, privacy-LLM)
- [x] Documentation (`docs/`, `README.md`, `CONTRIBUTING.md`)

## RFC-AI-0004 compliance

- [x] **Vendor neutrality** — placeholders (`<PROJECT>`, `<tracker>`, `<upstream>`, `<security-list>`) used in all skill / tool prose.

## Linked issues

Closes #1094 

***
Generated-by: Claude Code (Opus 5)

PR #622, opened 2026-06-29
feat(organizations): introduce Organization-adapter entity + document adapters

potiuk:
## What

First, additive step of the organization-adapter work (plan approved
separately). Introduces a new configuration entity and documents the
**Adapter** concept in the vendor-neutrality page — no behaviour change
yet (the reflow that makes projects *inherit* is the next PR).

## New entity — `organizations/<org>/`

- `organizations/README.md` — entity overview + resolution order.
- `organizations/ASF/` — the **reference org adapter**: PMC governance
  vocabulary, Vulnogram CVE authority, PonyMail archive,
  `apache-projects-mcp` metadata, ASF-security forwarder, and the
  `*.apache.org` / `cveprocess.apache.org` / `dist.apache.org` infra
  values.
- `organizations/independent/` — the no-formal-org baseline (DCO, GHSA,
  GitHub Releases; no list/forwarder/metadata backends).
- `organizations/_template/` — authoring skeleton.

## Resolution contract (`AGENTS.md`)

A project names its organization once — `organization: <org>` (default
`independent`) — and every placeholder / dotted key resolves
`project.md → organizations/<org>/ → framework default`, first hit wins.
Skills never branch on the organization. Adds the `<governance-body>`
and `<project-stage>` placeholders.

## Docs (`docs/vendor-neutrality.md`)

New **Adapters**, **Organization adapters**, and **Authoring your own
adapter** sections. The last documents the two supported paths:
**contribute upstream** under Apache-2.0, or **link to an
externally-defined adapter** (discovery-only per PRINCIPLES §13 — never
auto-fetched).

## Supporting

- Validator + placeholder linter: allowlist `organizations/`; register
  the two new placeholders.
- Spec: `tools/spec-loop/specs/organization-adapters.md` (experimental).

## Follow-ups (tracked)

- PR2: reflow `projects/_template` + `non-asf-example` to `organization:`
  and drop the now-inherited ASF values.
- PR3: make the `asf:false` skill families fully agnostic.
- PR4: replace per-family `asf: true/false` with an optional
  `organization:` scope; add the external-adapter discovery index.

## Verification
`skill-and-tool-validate` EXIT 0; doctoc, markdownlint, lychee,
placeholder check all green. (Python `workspace-mypy`/`workspace-pytest`
hooks skipped at commit — can't spawn in my env — but validator
mypy+pytest verified green via `uv run`.)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1135, opened 2026-08-31
docs(rfcs): RFC-AI-0008 — the coordinator role (multi-hop relay of security reports)

ppkarwasz:
## What

Adds **RFC-AI-0008** (status: Draft): names the **coordinator** persona of coordinated vulnerability disclosure — the party that receives, triages on behalf of others, and forwards to the owning vendor — and defines:

- CVD role vocabulary (finder / reporter / coordinator / vendor), with a strictly **pairwise** relay contract instead of named chain levels;
- the coordinator **profile** (functions + termination authority + upstreams + downstream directory);
- **provenance fields** carried verbatim across hops (finder credits, append-only reporter chain, origin references, embargo terms);
- the **handling-identifier exchange** between hops;
- **bidirectional** forwarder adapters (emit templates + detect signatures in one directory, closing the current two-repo drift gap);
- coordinator-mode skills (`security-coordinator-import` / `-assess`) generalized from the ASF Security team's out-of-tree pipeline, which becomes the reference adopter.

## Review fallout (second commit)

Cross-reference verification of the RFC against the tree surfaced two pre-existing inconsistencies, fixed here:

- `tools/forwarder-relay/README.md` still documented `contact_handle: '@raboof'` as the shipping default, contradicting the newer org-level `security@apache.org` in `organizations/ASF/organization.md` (#622). The shared inbox is now the documented default; an individual liaison's handle remains the override example. The one stale worked example in `docs/security/forwarder-routing-policy.md` is updated to match.
- `CONTRIBUTING.md`'s directory-tree comment enumerated "RFC-AI-0002, 0003, 0004" (stale since 0005); it now says "RFC-AI-0002 onward".

Modified paragraphs are reflowed to semantic line breaks per #1108.

https://claude.ai/code/session_018SKtoF3YBbXye9jmnZp1Pt

### DIFF
--- a/docs/rfcs/RFC-AI-0008.md
+++ b/docs/rfcs/RFC-AI-0008.md
@@ -1,0 +1,744 @@
+<!-- SPDX-License-Identifier: Apache-2.0
+     https://www.apache.org/licenses/LICENSE-2.0 -->
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+**Table of Contents**  *generated with [DocToc](https://github.com/thlorenz/doctoc)*
+
+- [RFC-AI-0008: The coordinator role — multi-hop relay of security reports](#rfc-ai-0008-the-coordinator-role--multi-hop-relay-of-security-reports)
+  - [Abstract](#abstract)
+  - [Status of this document](#status-of-this-document)
+  - [Motivation](#motivation)
+  - [Proposal](#proposal)
+    - [CVD vocabulary, not level names](#cvd-vocabulary-not-level-names)
+    - [The coordinator profile](#the-coordinator-profile)
+    - [Provenance fields — carried across hops](#provenance-fields--carried-across-hops)
+    - [The handling-identifier exchange](#the-handling-identifier-exchange)
+    - [Forwarder adapters become bidirectional](#forwarder-adapters-become-bidirectional)
+    - [Coordinator-mode skills](#coordinator-mode-skills)
+    - [Classification cascade becomes config-driven](#classification-cascade-becomes-config-driven)
+    - [Hardening conversion — exactly one hop](#hardening-conversion--exactly-one-hop)
+    - [Deduplication composes per hop](#deduplication-composes-per-hop)
+    - [Tools](#tools)
+    - [Reference adopter and migration](#reference-adopter-and-migration)
+  - [Security model](#security-model)
+  - [Drawbacks](#drawbacks)
+  - [Alternatives considered](#alternatives-considered)
+  - [Out of scope](#out-of-scope)
+  - [References](#references)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+<!-- SPDX-License-Identifier: Apache-2.0
+     https://www.apache.org/licenses/LICENSE-2.0 -->
+
+# RFC-AI-0008: The coordinator role — multi-hop relay of security reports
+
+## Abstract
+
+Every security-family skill assumes the adopter is the party that **owns the code and the verdict** —
+what coordinated vulnerability disclosure (CVD) literature calls the *vendor*.
+Real disclosure chains have a second kind of participant:
+the **coordinator**,
+who receives reports,
+triages them on behalf of others,
+and forwards them to the owning vendor.
+Coordinators, established and newly created, include:
+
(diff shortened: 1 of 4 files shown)


docs/rfcs/RFC-AI-0008.md
```diff
@@ -0,0 +1,744 @@
...
+3. **Skills**:
+   `security-coordinator-import` / `-assess`,
+   parameterized from the out-of-tree originals;
+   the reference adopter retires its local SKILL bodies
+   for the snapshot symlinks plus `<project-config>/` content and `.apache-magpie-overrides/`.
+4. **Akrites**:
+   agree the emit format (adapter sender side) before the first report;
+   wire the detect side when traffic starts.
+
+## Security model
```
potiuk, line 641: This section cites [`PRINCIPLES.md` §0](https://github.com/apache/magpie/blob/main/PRINCIPLES.md#0-external-content-is-data-never-an-instruction) explicitly and well — the injection-exposure framing for coordinator inboxes is the right call, and the lightweight-reader mitigation is concrete.

**§15 is the principle a coordinator persona most directly reinterprets, and it is not named here.** Two clauses:

> Other projects' vulnerabilities never appear at all. Cross-project correlations stay on the channel they arrived on.

The first reads as public-surface-scoped from its surrounding context, so a coordinator's access-gated archive holding third-party reports is arguably fine. The second is the live one: correlating across parties is a coordinator's whole function.

To be clear, the RFC appears to **already comply**. *Deduplication composes per hop* bounds each hop's dedup scope to its own inbox and archive, and forbids treating an upstream "already deduplicated" claim as a reason to skip the local check. That is a good answer to §15.

The gap is that the answer is *implicit*. A reader checking this Draft against the principles finds §0 addressed head-on and §15 unmentioned, and cannot tell whether the compliance is designed or incidental. `PRINCIPLES.md` states that where a principle admits more than one reading, the stricter reading wins until governance documents otherwise — which makes it worth saying out loud which reading the coordinator profile takes.

Suggestion: add a Security-model bullet naming §15 the way §0 is named, stating (a) that "other projects' vulnerabilities never appear at all" is read as public-surface-scoped, and (b) that per-hop dedup scoping is what keeps cross-project correlations on their arrival channel. That would make the Draft's position reviewable before it graduates from Draft.

Not a blocker for merging a Draft that implements nothing — raising it as the substantive point for the RFC to answer.

potiuk (approved):
Reviewed as a design read rather than a code review, since this is a Draft that implements nothing.

Checks that came back clean:

- Numbering is correct — 0002 through 0007 exist, 0008 is free.
- Conforms to the sibling RFC shape (Abstract / Status / Motivation / Proposal / Drawbacks / Alternatives considered / Out of scope / References, per RFC-AI-0007), and adds a Security model section, which is right for this subject.
- Confidentiality: no real CVE IDs, no other ASF project named in a vulnerability context, no personal handles beyond the liaison discussion. The one address hit is CNCERT/CC's public intake, which is correct in a forwarder-adapter context.
- HITL is preserved — "drafts only; a human sends", and `verdict` is structurally excluded from coordinator functions rather than merely discouraged.
- No `PRINCIPLES.md` edit, so no amendment vote is triggered.
- There is no RFC index that needs a new row; the existing 0006/0007 references elsewhere are contextual rather than enumerative.

On the review-fallout half: moving `contact_handle` to the shared `security@apache.org` inbox and demoting the individual handle to an override example is the right direction — it takes a single person out of shipped config while keeping the liaison pattern documented.

The Drawbacks section naming the single-reference-adopter overfitting risk is good practice; that is the failure mode most likely to bite here.

One substantive point inline, about §15 and the Security model section. I believe the RFC already complies via the per-hop dedup scoping — the issue is that the compliance is implicit where §0's is explicit. Left open for you rather than resolved, since it is a design question rather than a nit.

---

docs/rfcs/RFC-AI-0008.md
potiuk: Addressed directly in 53206a1c rather than leaving it for a follow-up, since it is a wording addition rather than a design change — shout if you would rather phrase it differently.

The new Security-model bullet names §15 and separates its two clauses: *"other projects' vulnerabilities never appear at all"* read as public-surface-scoped (§15 governs what leaves the archive, not what it stores), and *"cross-project correlations stay on the channel they arrived on"* as the binding clause that per-hop dedup scoping satisfies. Added the matching References row next to the existing §0 one.

No normative change — it states the position *Deduplication composes per hop* already takes. CI is green on the new head.

Issue #1137, opened 2026-08-31
fix(symlink-lint): self-adoption relays drift from skills/ — two skills unreachable, no completeness rule


dpol1:
## What

The self-adoption relay directories have drifted out of sync with `skills/`. The source tree has **71** skill directories; the relays don't:

| dir | count | missing |
|---|---|---|
| `skills/` | 71 | — |
| `.agents/skills/` | 70 | `magpie-dependency-license-audit` |
| `.claude/skills/` | 70 | `magpie-dependency-license-audit` |
| `.github/skills/` | 70 | `magpie-dependency-license-audit` |
| `.kiro/skills/` | 69 | `magpie-dependency-license-audit`, `magpie-report-framework-issue` |

`plugins/magpie-repo-health/` and `plugins/magpie-utilities/` carry both skills correctly — only the relay surface drifted.

```bash
comm -23 <(ls skills | grep -v pyproject | sort) \
         <(ls .kiro/skills | sed 's/^magpie-//' | sort)
# dependency-license-audit
# report-framework-issue
```

## Why it matters

A skill with no relay is invisible to every harness that discovers skills through its agent dir — which is exactly the surface self-adoption exists for. `dependency-license-audit` shipped in #814 with an eval suite and a plugin entry, and has been unreachable from the agent dirs ever since; nobody noticed.

Nothing catches this. `symlink-lint` validates the symlinks that *exist* (no cycles, relay-through-canonical) but has no completeness rule, so a skill that never got its links is never flagged. prek and CI are green on this tree. The same failure class already bit once on the plugin surface (#1109, a new skill missing from `plugins/`) and got a deterministic check there (`check-family-plugins`); the relay surface has no equivalent, so this will recur the next time a skill lands without its symlinks.

## Suggested fix

Two pieces, in order:

1. **Wire the missing links** — canonical `.agents/skills/magpie-<skill>` entries for the two skills, plus the relays in every wired agent dir.
2. **Add a completeness rule to `symlink-lint`** — every directory under `skills/` must have its canonical `.agents/skills/magpie-<skill>` entry, and every wired agent dir must relay the full canonical set — with a test. Scoped to the framework's own working tree, so an adopter's legitimately-dangling links stay exempt exactly as they are for rules 1–2 today.

Worth doing as one change so the lint lands together with the last drift it would have caught.

## Related

- #814 — added `dependency-license-audit` (skill + evals + plugin entry; relays never created).
[...]

PR #1143, opened 2026-09-01
fix(symlink-lint): enforce relay completeness and sync missing self-adoption symlinks

onlyarnav:
## Summary

- The self-adoption relay directories drifted out of sync with `skills/` (71 skill directories):
  - `.agents/skills/` (canonical), `.claude/skills/`, `.github/skills/`, and `.kiro/skills/` were missing `magpie-dependency-license-audit`.
  - `.kiro/skills/` was missing `magpie-report-framework-issue`.
- Added Rule 3 (Completeness) to `symlink-lint`: in the framework checkout (where `skills/` lives), every skill directory under `skills/` must have its canonical entry under `.agents/skills/` and every wired agent directory (`.claude/`, `.github/`, `.kiro/`, etc.) must relay the full canonical set. Adopter checkouts (where root `skills/` is absent) remain exempt.
- Added comprehensive unit test coverage for Rule 3 in `tools/symlink-lint/tests/test_symlink_lint.py`.
- Updated documentation in `tools/symlink-lint/README.md`, `docs/labels-and-capabilities.md`, and `tools/spec-loop/specs/meta-and-quality-tooling.md`.

## Type of change

## Test plan

- [x] `uv run --directory tools/symlink-lint --project . ruff check src tests` passes
- [x] `uv run --directory tools/symlink-lint --project . mypy src tests` passes
- [x] `uv run --project tools/symlink-lint pytest tools/symlink-lint/tests` passes
- [x] `python tools/symlink-lint/src/symlink_lint/__init__.py --archive` passes (exit code 0)
- [x] Verified git index entries for all 5 missing symlinks (.agents, .claude, .github, .kiro) are staged as mode `120000`

## RFC-AI-0004 compliance

- [x] **Vendor neutrality** — ensures all supported agent directories receive full relays for framework skills

Generated-by: Antigravity (Gemini 3.7 Flash)


dpol1 (commented):
Thanks for picking #1137 up so fast — the five links are right (mode `120000`, canonical → `../../skills/…`, relays → `../../.agents/skills/magpie-…`), and rule 3 run against `main` reports exactly the five missing links and nothing else. Skill discovery mirrors the validator (`is_dir()` + no leading dot + `PRUNE_DIR_NAMES`), so `skills/pyproject.toml` and the cache dirs are ignored, and agent-dir discovery stays convention-based like rules 1–2. Locally on the PR head: `symlink-lint` and `--archive` exit 0, `ruff check` / `mypy` / pytest (27) green. Branch is clean against `main` (`MERGEABLE`); merge is `BLOCKED` only by the red `prek` check.

Two things need to land before this can merge, plus a few smaller ones.

### CI is red — `workspace-ruff-format` (`__init__.py:145-147`)

Confirmed from run 33528122998, not guessed: `ruff format (workspace)...Failed`, `1 file would be reformatted`. The three-line `find_missing_relays(...)` signature fits the tool's 110-column limit, so ruff collapses it to one line. `uv run --directory tools/symlink-lint --project . --group dev ruff format src` and recommit. (AGENTS.md § Local setup: "Before opening or updating a PR, run `prek run --all-files` … as a hard pre-flight gate.")

### The hook never fires on the case rule 3 exists for (`.pre-commit-config.yaml:234`)

The `symlink-lint` hook is `types: [symlink]`. Rule 3's failure mode — a new `skills/<x>/` committed with no links — stages no symlink, so locally the hook does not run on exactly that commit; only CI's `--all-files` catches it, which is how #1137's drift landed in the first place. The archive hook at `:250-251` already has the right shape: add a `files:` pattern covering `skills/.*` and the agent `skills/` trees with `types_or: [symlink, file]`, and update the comment at `:224-227` and `README.md:106-107` ("fired on any staged symlink"). (AGENTS.md § Local setup: "fix the underlying issue or update the hook config in the same PR"; README.md:53-54 promises rule 3 "catches new skills that landed without their self-adoption symlinks".)

### Smaller points

- **Scoping heuristic (`__init__.py:162-164`).** "Framework checkout" is inferred from `root/skills` being any directory. An adopter repo with its own top-level `skills/` gets told to add `.agents/skills/magpie-<subdir>` for every subdirectory (reproduced on a scratch tree). Rule 2 makes a similar assumption already, so this is not new exposure — but the README/docstring now claim adopters are exempt, which overstates it. Either gate on a framework marker (e.g. `skills/setup/SKILL.md`), or reword the exemption.
- **Trusted-source pointer dirs (`__init__.py:166-170`).** A `skills/<name>/source.md` pointer dir (no `SKILL.md`) is treated as a skill, so rule 3 demands a canonical link → `../../skills/<name>`. Per `skills/setup/skill-sources.md:222-231` that link's real target is `../../.apache-magpie-sources/<id>/skills/<name>/` and it is gitignored — so a fresh clone with any pointer dir fails the lint (reproduced). None exist in-tree today, so this is latent; skip dirs without `SKILL.md` the way the validator's `is_skill_source_pointer` does, and add the test case.
- **Tests (`tests/test_symlink_lint.py:192-197`).** `test_completeness_unwired_agent_dir_not_required` builds the same tree as the clean test — no agent dir *without* a `skills/` child is ever created, so the behaviour in its name is untested; `(tmp_path / ".github").mkdir()` before the assert fixes it. Missing cases: pointer dir (above), `skills/pyproject.toml`.
- **Out-of-scope changes.** `os.path` → `posixpath` (`:53`, `:252-253`), `_rel` returning `str` (`:262-266`), and the module-wide `pytestmark = skipif(not _can_symlink())` (`tests:53-55`) are not in #1137 or the commit message. The skipif in particular silently skips the whole suite — archive and `main()` tests included — wherever `os.symlink` fails. Drop them here or split into their own PR with the rationale. (AGENTS.md § Before submitting: "Re-read the diff and check that every change is intentional.")
- **Doc drift to three rules.** `README.md:78` ("Unlike rules 1–2"), `docs/adapters/add-a-harness.md:76` ("enforces both rules"), `.pre-commit-config.yaml:224-225` comment.
- **Wording.** `README.md:54-55` / `__init__.py:41-42` conflate the dangling-skip (applies everywhere) with rule 3's `skills/`-absent exemption. `docs/labels-and-capabilities.md:305` and the spec edit at `:43-44` now have "(canonical/relay target-correctness)" trailing "incomplete … sets" — it belongs after "misdirected skill relays".
- **Nits.** `__init__.py:185-190`: every target in `skills/setup/agents.md` is a dot-dir, so `entry.name.startswith(".")` tightens the wired-dir scan cheaply. PR body: `Fixes #1137` (or `Refs`) instead of the `## possible fix for:` heading, so the link is machine-readable. `tools/spec-loop/.last-sync` was not bumped alongside the spec edit — though it is already behind `main` for unrelated reasons, so that is a judgment call.

---

> *This review was drafted by an AI-assisted tool and posted by
> a contributor who does not have confirmed Apache Magpie maintainer
> access. The findings below are this tool's analysis only, not
> a maintainer sign-off; an Apache Magpie maintainer will still need
> to look at the PR before it moves forward. If you think a
> finding is mis-applied, please reply on the PR.*
>

dpol1:
I'm gonna make the commit from myself - the review was just to explain the original idea for fixing that

dpol1 (approved):

.pre-commit-config.yaml
```diff
@@ -221,17 +221,19 @@ repos:
...
+  # `prek run --all-files`). Rules, rationale, and the pytest suite live
   # in tools/symlink-lint (README + tests).
   - repo: local
     hooks:
       - id: symlink-lint
         name: symlink-lint
         language: system
         entry: python3 tools/symlink-lint/src/symlink_lint/__init__.py
-        types: [symlink]
+        files: ^(\.(agents|claude|github|kiro)/skills/.*|skills/.*)$
```
potiuk, line 235: Non-blocking, forward-looking.

This regex triggers the hook on the four agent dirs that exist today — but rule 3 discovers wired agent dirs *dynamically* (any top-level dotdir containing a `skills/` folder), so the hook's trigger scope is narrower than the rule's enforcement scope.

Concretely: `docs/adapters/add-a-harness.md` is the doc for adding a new harness dir, and `tools/symlink-lint/README.md:33` already names `.windsurf/` and `.goose/` as expected relay dirs. Someone following that doc to add `.windsurf/skills/` gets no local hook feedback on those staged files. `prek run --all-files` in CI still catches it, so this is a local-feedback gap rather than a correctness hole.

Either widen it to match what the rule actually scans:

```suggestion
        files: ^(\.[^/]+/skills/.*|skills/.*)$
```

…or keep the explicit list and add a comment noting it needs updating whenever a new harness dir lands, so the coupling is visible to the next person editing `add-a-harness.md`.

Resolving this myself so it doesn't hold up the merge — it's fine as a follow-up.

potiuk (approved):
Nice piece of work. The rule is self-validating in the best way: turning it on immediately surfaced five real gaps — `magpie-dependency-license-audit` missing from all four agent dirs and `magpie-report-framework-issue` missing from `.kiro/` — and the same PR fixes them.

What I verified beyond CI:

- 28 unit tests pass at this head.
- Linter exits 0 on the PR tree; deleting one canonical and one relay link produces both errors, each naming the exact expected target. The error output is genuinely actionable rather than just "something is wrong".
- The adopter exemption is gated on `skills/setup/SKILL.md` rather than on `skills/` merely existing, so an adopter with their own `skills/custom/` isn't falsely flagged — and there's a test pinning that.
- Trusted-source pointer dirs (`source.md`, no `SKILL.md`) are excluded. That's a non-obvious interaction with the skill-sources mechanism, and it's covered by a test.
- The old rule 3 → 4 renumbering is complete across the module docstring, the README (both hook descriptions included), `docs/labels-and-capabilities.md`, and the spec-loop spec — so the `AGENTS.md` spec-sync pre-check is satisfied.

`types_or: [symlink, file]` is necessary rather than redundant, since the default `types: [file]` would skip symlinks — worth noting for anyone who reads that line and wonders.

One non-blocking nit inline about the hook's `files:` regex vs. the rule's dynamic discovery. I'm resolving it myself so it doesn't block the merge; it can be a follow-up.

---

PR #1144, opened 2026-09-02
fix(skills): derive the two remaining Airflow literals from project config

### LINKED ISSUE #1139: fix(skills): two hardcoded Airflow values slip past check-placeholders.sh
## What

Two skill docs hardcode Apache Airflow values in operative (non-example) positions:

| where | literal | should come from |
|---|---|---|
| `skills/pr-management-stats/fetch.md:213` | `repository(owner:"apache",name:"airflow")` in the GraphQL batch-fetch recipe | the configured `<upstream>` slug (`upstream_repo` in `<project-config>/project.md`), split into owner/name |
| `skills/release-announce-draft/SKILL.md:363` | `closer.lua?path=airflow/<version>/...` in the site-bump constraint | the `project_dist_name` key in `<project-config>/release-management-config.md` — the skill already reads that file (cf. its `product_name` derivation at line 237) |

`tools/dev/check-placeholders.sh` passes on both: its `FORBIDDEN_PATTERNS` list is four fixed strings (`apache/airflow`, `airflow-s/airflow-s`, `Apache Airflow`, `apache.org/airflow`) and neither lowercase form matches.

## Why it matters

`AGENTS.md` calls this out directly: *"Writing a literal project value directly into a skill is a refactor bug — skills must stay project-agnostic so swapping projects is a config change, not a code change."* An agent following the `fetch.md` recipe against any other adopter queries the wrong repository; an adopter's announce draft points its download links at Airflow's dist tree. And because the linter's pattern list doesn't cover the lowercase forms, this class of residue can keep slipping in — #1132 just cleaned the same class out of `projects/_template/`, but these two sit in skill docs, outside both its scope and the linter's allowlist.

No new placeholder is needed for either fix — both values are already derivable from existing configuration, which is the point of the placeholder table's "thread a needed value in via the project manifest" rule.

## Suggested fix

- `fetch.md:213` — derive `owner` / `name` from `<upstream>`, the way the surrounding steps already treat it.
- `release-announce-draft/SKILL.md:363` — render the `closer.lua` path from `project_dist_name`.
- `check-placeholders.sh` — add `name:"airflow"` and `path=airflow` to `FORBIDDEN_PATTERNS` so the lowercase class is caught at commit time.

For scope: a `grep -rn 'airflow' skills/` sweep shows the remaining hits are marked examples (`candidate-rules.md`'s "the shape for an Airflow-like project", the `issue-triage` JIRA-key example, the slop-detection fixtures) — this issue is only the two operative literals plus the linter patterns.

## Related

- #1132 — same residue class, template surface.
- `AGENTS.md` § Placeholder convention — the rule both literals violate.

dpol1:
## Summary

- `pr-management-stats/fetch.md` queried `repository(owner:"apache",name:"airflow")` in its GraphQL batch-fetch recipe, and `release-announce-draft` pinned the download URL to `closer.lua?path=airflow/<version>/`. Both are operative instructions, not examples — an agent following them against any other adopter queries the wrong repository and points the announce at the wrong dist tree.
- Both values now come from configuration that already exists: the query binds `$owner` / `$repo` from `<upstream>` the way the other GraphQL recipes (`contributor-*`, `pr-management-code-review`) already do, and the dist path renders from `<project>`, which the same announce skill already uses for `dist/release/<project>/<version>/`. No new placeholder.
- `check-placeholders.sh` learns the two lowercase forms (`name:"airflow"`, `path=airflow`). Before this change it passed on both lines — which is how they survived the template clean-up in #1132.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] Linter first, fix second: with the two patterns added, `tools/dev/check-placeholders.sh` fails on exactly `fetch.md:213` and `release-announce-draft/SKILL.md:363` and nothing else; after the two edits it passes again.
- [x] Eval suites: both edits are prose inside example blocks — no step behaviour changes — so no fixture update is needed per AGENTS.md § *When the rule fires* (pure prose edits).
- [x] `grep -rn 'airflow' skills/` sweep: the remaining hits are marked examples (`candidate-rules.md`, the `issue-triage` JIRA-key example, slop-detection fixtures) and stay as they are.

## RFC-AI-0004 compliance

## Linked issues

Fixes [apache/magpie#1139](https://github.com/apache/magpie/issues/1139).

### DIFF
--- a/tools/dev/check-placeholders.sh
+++ b/tools/dev/check-placeholders.sh
@@ -51,6 +51,11 @@ FORBIDDEN_PATTERNS=(
   "airflow-s/airflow-s"
   "Apache Airflow"
   "apache.org/airflow"
+  # Lowercase forms that slip past the four above: a GraphQL
+  # `repository(owner:"apache",name:"airflow")` argument and a
+  # `closer.lua?path=airflow/` dist path.
+  'name:"airflow"'
+  "path=airflow"
 )
 
 # Files / directories where Airflow references are intentional:
--- a/skills/pr-management-stats/fetch.md
+++ b/skills/pr-management-stats/fetch.md
@@ -209,8 +209,8 @@ Two stages:
 2. **Fetch comments in aliased batches** of **30 PRs per GraphQL call**. Each alias queries one PR with `comments(last: 25) { nodes { author { login } authorAssociation createdAt body } }`. A single query with 30 aliases stays well inside GitHub's complexity budget; larger batch sizes occasionally hit `MAX_NODE_LIMIT_EXCEEDED`. Budget: ~54 GraphQL calls for 1600 PRs; end-to-end around 60 seconds on a warm token.
 
 ```graphql
-query {
-  repository(owner:"apache",name:"airflow") {
+query($owner: String!, $repo: String!) {   # bound from <upstream> split on "/"
+  repository(owner:$owner, name:$repo) {
     pr63407: pullRequest(number:63407) {
       number author{login} authorAssociation closedAt mergedAt state merged
       labels(first:30){nodes{name}}
--- a/skills/release-announce-draft/SKILL.md
+++ b/skills/release-announce-draft/SKILL.md
@@ -360,7 +360,7 @@ before including it.
 
 - Download links in the site files must resolve through the `closer.lua`
   mirror redirector (e.g.
-  `https://www.apache.org/dyn/closer.lua?path=airflow/<version>/...`),
+  `https://www.apache.org/dyn/closer.lua?path=<project>/<version>/...`),
   not through a direct `dist.apache.org` URL.
 - The PR is opened (not merged) by this skill; a committer merges it
   after the `[ANNOUNCE]` email is sent.


tools/dev/check-placeholders.sh
```diff
@@ -51,6 +51,11 @@ FORBIDDEN_PATTERNS=(
   "airflow-s/airflow-s"
   "Apache Airflow"
   "apache.org/airflow"
+  # Lowercase forms that slip past the four above: a GraphQL
+  # `repository(owner:"apache",name:"airflow")` argument and a
+  # `closer.lua?path=airflow/` dist path.
+  'name:"airflow"'
```
potiuk, line 57: `FORBIDDEN_PATTERNS` are matched with `grep -F`, so `'name:"airflow"'` only catches the no-space form. The whitespace variant slips through every pattern in the list, old and new:

```graphql
repository(owner: "apache", name: "airflow")
```

I verified it: `grep -F 'name:"airflow"'` and `grep -F 'apache/airflow'` both return 0 matches on that line. Since this PR is specifically about forms that evade the existing four, the adjacent-whitespace variant is the same bug class one step over.

Cheapest fix that keeps the fixed-string contract is a second entry:

```suggestion
  'name:"airflow"'
  'name: "airflow"'
```

Non-blocking — the change is a strict improvement as-is, and this can equally be a follow-up.

potiuk (approved):
Good catch on both literals, and thank you for closing the linter hole in the same PR rather than leaving it as a follow-up — that's what keeps [§12](https://github.com/apache/magpie/blob/main/PRINCIPLES.md#12-the-framework-is-project-agnostic-concrete-names-live-in-adopter-config) enforceable instead of aspirational. The comment explaining *why* these two evaded the existing four patterns is exactly the right thing to leave behind.

Verified: `prek` is green on this head and `check-placeholders` runs as a prek hook, so the new patterns provably pass against the fixed tree. `<project>` is the established placeholder for the ASF project slug (used the same way in `tools/asf-svn/`), so `path=<project>/<version>` is consistent, not a new convention.

One non-blocking nit inline on the whitespace variant. Approving — it's fine as a follow-up.

Aside, not for this PR: `<project>` is used 313 times across the repo but isn't in the `AGENTS.md` placeholder table alongside `<project-config>` / `<tracker>` / `<upstream>`. Given "do not invent new placeholders" is a rule there, documenting the one that's already load-bearing seems worth a separate issue.

---

PR #1147, opened 2026-09-03
fix(validator): tokenize gh-list-no-limit's flag check instead of substring match

AmirF194:
## Summary

- `validate_gh_list_limit` decides whether a `gh issue list` / `gh pr list` call is missing
  `--limit` by testing whether the literal substring `--limit` occurs anywhere on the
  logical line, so a trailing shell comment or a quoted argument that merely mentions
  `--limit` silently suppresses the warning even when the command has no such flag.
- Fixes it by tokenizing the line with `shlex.split(..., comments=True)` and checking for an
  actual `--limit` (or `--limit=N`) token instead of a bare substring match.

## Type of change

## Test plan

- Added three cases to `TestGhListLimit`: a trailing comment containing `--limit`, a quoted
  argument containing `--limit`, and the `--limit=100` equals-form (control, must stay
  silent). The first two fail on `main` and pass on this branch, verified both ways on
  Python 3.12. Full `tests/test_validator.py` (466 cases) still passes.
- I have not checked every `--limit` spelling `gh` itself accepts, only `--limit N` and
  `--limit=N`.

## RFC-AI-0004 compliance

- [x] **Sandbox**: no new host access, pure text processing like the code it replaces

## Linked issues

Self-discovered while reading the validator surface; no existing issue.

`_GH_LIST_RE` matching and the fenced-block boundary are unchanged. The only touched
function is `validate_gh_list_limit`, single call site.

### DIFF
--- a/tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
+++ b/tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
@@ -151,6 +151,7 @@ from __future__ import annotations
 import argparse
 import contextlib
 import re
+import shlex
 import subprocess
 import sys
 from collections.abc import Iterable
@@ -2726,7 +2727,11 @@ def validate_gh_list_limit(path: Path, text: str) -> Iterable[Violation]:
             if line_end == -1:
                 line_end = len(joined)
             logical_line = joined[line_start:line_end]
-            if "--limit" in logical_line:
+            try:
+                tokens = shlex.split(logical_line, comments=True)
+            except ValueError:
+                tokens = []  # unbalanced quoting, can't tell, so flag it
+            if any(tok == "--limit" or tok.startswith("--limit=") for tok in tokens):
                 continue
             line_no = text[: block_match.start()].count("\n") + joined[: cmd_match.start()].count("\n") + 1
             yield Violation(
--- a/tools/skill-and-tool-validator/tests/test_validator.py
+++ b/tools/skill-and-tool-validator/tests/test_validator.py
@@ -1800,6 +1800,23 @@ class TestGhListLimit:
         violations = list(validate_gh_list_limit(path, text))
         assert not any("gh-list-no-limit" in v.message for v in violations)
 
+    def test_fires_when_limit_only_appears_in_a_trailing_comment(self, tmp_path: Path) -> None:
+        path = tmp_path / "SKILL.md"
+        text = _fenced("gh issue list --repo <repo> --state open  # add --limit later if slow")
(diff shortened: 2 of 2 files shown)


tools/skill-and-tool-validator/src/skill_and_tool_validator/__init__.py
```diff
@@ -2726,7 +2727,11 @@ def validate_gh_list_limit(path: Path, text: str) -> Iterable[Violation]:
             if line_end == -1:
                 line_end = len(joined)
             logical_line = joined[line_start:line_end]
-            if "--limit" in logical_line:
+            try:
+                tokens = shlex.split(logical_line, comments=True)
+            except ValueError:
+                tokens = []  # unbalanced quoting, can't tell, so flag it
+            if any(tok == "--limit" or tok.startswith("--limit=") for tok in tokens):
```
potiuk, line 2734: Non-blocking, and pre-existing rather than introduced here — but this PR is precisely about the check's accuracy, so it is the natural place for it.

`gh` accepts a short form for this flag:

```console
$ gh issue list --help | grep -i limit
  -L, --limit int          Maximum number of issues to fetch (default 30)
```

So a correctly-limited `gh issue list --repo <repo> -L 50` is still reported as `gh-list-no-limit`. I confirmed it against your branch — it flags. The old substring check missed it too, so this is not a regression, just the same gap surviving the rewrite.

Now that the check is token-based, closing it is a one-token change:

```suggestion
            if any(tok in ("--limit", "-L") or tok.startswith("--limit=") for tok in tokens):
```

Resolving this myself so it does not hold up the merge — happy for it to be a follow-up.

potiuk (approved):
Good fix, and the right shape: a substring test on a shell line was always going to over-match, and tokenising is the correct answer rather than layering more string heuristics on top.

I verified the behaviour on your branch rather than just reading it:

| Input | Flagged |
|---|---|
| `--limit` in a trailing comment | yes (was wrongly silent) |
| `--limit` inside `--search "..."` | yes (was wrongly silent) |
| real `--limit 100` | no |
| `--limit=100` | no |
| unbalanced quote | yes |

The `except ValueError: tokens = []` path is the detail I liked most — an unparseable line flags rather than silently passing, and the comment says why. That is the correct direction to fail for a lint rule, and it is easy to get backwards.

One non-blocking nit inline about the `-L` short form. I am resolving it so it does not block the merge; it can be a follow-up.

---

AmirF194:
Appreciate you actually running the edge cases instead of just reading the diff, the trailing-comment and quoted-argument rows were the ones I was least sure about. Good call resolving the -L nit as a follow-up rather than blocking on it.

PR #1149, opened 2026-09-05
fix(sandbox-lint): normalise every trailing slash, not just one

### LINKED ISSUE #1148: sandbox-lint's path normaliser strips only one trailing slash, so a double-slash credential path bypasses the M.29 check
### What's broken

`tools/sandbox-lint`'s `_normalise()` strips a single trailing slash before comparing a `.claude/settings.json` filesystem path against `FORBIDDEN_ALLOW_READ` / `FORBIDDEN_ALLOW_WRITE` (mitigation M.29). A path with two or more trailing slashes, e.g. `~/.ssh//`, only loses one, so `check_invariants()` reports no violation for an `allowRead`/`allowWrite` entry that in practice still grants access to a credential directory.

### Which layer

`tools/sandbox-lint/src/sandbox_lint/__init__.py`, `_normalise()` (line 123) and every invariant check that consumes it through `_normalised_set()`.

### How to reproduce

```python
from sandbox_lint import check_invariants
settings = {
    "sandbox": {
        "enabled": True,
        "filesystem": {
            "denyRead": ["~/"],
            "allowRead": ["~/.ssh//"],
            "allowWrite": [],
        },
    },
    "permissions": {"deny": [...]},  # the full REQUIRED_PERMISSIONS_DENY set
}
print(check_invariants(settings))  # []
```

### Expected vs actual

Expected: an `allowRead` entry of `~/.ssh//` is caught the same way `~/.ssh` and `~/.ssh/` already are, since it names the same directory.

Actual: `check_invariants()` returns an empty error list, so the credential-path check silently passes.

### Surface area

Same gap applies to `denyRead` and `allowWrite`. I have a fix ready (`_normalise` stripping all trailing slashes instead of one) with regression tests, opening as a PR against this issue.

AmirF194:
## Summary

- `_normalise()` only stripped one trailing slash, so `~/.ssh//` (two slashes) never matched the normalised `~/.ssh` forbidden entry and `check_invariants()` passed a config it should have rejected.
- Fixes #1148.

## Type of change

## Test plan

- Added a double-trailing-slash case to the existing `allowRead`/`allowWrite` parametrized tests; both are red on main, green on this branch.
- `ruff check`, `ruff format --check` and `mypy` on `tools/sandbox-lint` are clean (mypy needs `pytest` installed to resolve the test-file imports; not a change this PR introduces).
- Full `tools/sandbox-lint` suite: 55 passed, in a clean `python:3.11-slim` container. Did not run `prek` itself (not available in this sandbox), ran the underlying ruff/mypy/pytest commands by hand instead.

## RFC-AI-0004 compliance

- [x] **Sandbox**: no new host access, this only makes the existing credential-path check stricter.

## Linked issues

Fixes #1148.

### DIFF
--- a/tools/sandbox-lint/src/sandbox_lint/__init__.py
+++ b/tools/sandbox-lint/src/sandbox_lint/__init__.py
@@ -121,15 +121,13 @@ REQUIRED_DENY_READ = {"~/"}
 
 
 def _normalise(path: str) -> str:
-    """Strip a single trailing slash so '~/.aws' and '~/.aws/' compare equal.
+    """Strip all trailing slashes so '~/.aws' and '~/.aws//' compare equal.
 
-    The sandbox treats both forms as the same directory; the validator
-    must too, otherwise the forbidden-list could be bypassed by a
-    trailing-slash variant.
+    The sandbox treats every trailing-slash count as the same directory;
+    the validator must too, otherwise the forbidden-list could be bypassed
+    by a multi-slash variant.
     """
-    if path.endswith("/") and len(path) > 1:
-        return path[:-1]
-    return path
+    return path.rstrip("/") or "/"
 
 
 def _normalised_set(paths: list[str]) -> set[str]:
--- a/tools/sandbox-lint/tests/test_validator.py
+++ b/tools/sandbox-lint/tests/test_validator.py
@@ -163,6 +163,7 @@ def test_invariant_missing_deny_read_root(baseline: dict[str, Any]) -> None:
         "~/.config/gcloud",
         "/",
         "~/",
+        "~/.ssh//",
     ],
 )
 def test_invariant_allow_read_rejects_credential_paths(baseline: dict[str, Any], forbidden: str) -> None:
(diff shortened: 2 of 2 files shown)


potiuk (approved):
Correct fix, and the bypass it closes is real: `~/.ssh//` survived the strip-one logic and so slipped the credential-path forbidden list entirely.

`path.rstrip("/") or "/"` is the right form — the `or "/"` guard matters, since a bare `/` would otherwise normalise to the empty string and stop matching the root entry.

I checked the two things that could have gone wrong:

- **Edge cases** — `/` → `/`, `//` → `/`, `~/` → `~`, `/a/b//` → `/a/b`, and `""` → `/`. The last one is a behaviour change from the old code (which returned `""`), but it fails closed: a malformed empty entry now matches the forbidden root rather than silently comparing equal to nothing.
- **Call sites** — all four uses are set-membership tests that normalise *both* sides (`_normalised_set` on the config, `_normalise` on the required/forbidden constant). Nothing does prefix matching that depended on a retained trailing slash, so stripping more is safe here.

Test coverage matches the fix: `~/.ssh//` on the read side and `~/.aws//` on the write side.

---

AmirF194:
Thanks for checking the call sites too, that was the part I hadn't fully convinced myself of. Good catch that the empty-string case is a behavior change, glad it fails closed.

PR #1151, opened 2026-09-05
fix(agent-guard): scope Segment.raw to its own tokens, not the whole command

### LINKED ISSUE #1150: agent-guard: Segment.raw is the whole compound command, not the segment's own text
### What's broken

`split_segments()` gives every `Segment` in a compound command (`a && b`, `a; b`, ...) the exact same `raw` text: the whole input string, not that segment's own text. `guard_commit_trailer` scans `seg.raw` for a `Co-Authored-By:` trailer, so a `git commit` with a completely clean message gets denied if any other segment on the same line happens to mention that phrase (a comment, a grep, an echo explaining the rule).

### Which layer

`tools/agent-guard/src/agent_guard/__init__.py`, `Segment.__init__` / `split_segments()` (around line 137 and 155-172). `Segment`'s own docstring says `raw` is "the original text of the segment", but `split_segments` constructs every segment as `Segment(current, command)`, passing the full, unsliced input each time.

### How to reproduce

```
python3 -c "
import sys; sys.path.insert(0, 'tools/agent-guard/src')
from agent_guard import dispatch
print(dispatch('echo \"note: never add a Co-Authored-By: trailer\" && git commit -m \"clean fix, no trailer here\"'))
"
```

### Expected vs actual

Expected: `None` (allow), since the commit's own message has no trailer.
Actual: the commit-trailer deny string, because `seg.raw` for the `git commit` segment is the entire compound command, including the unrelated `echo` segment's text.

### Surface area

`GuardContext.raw` exposes the same field to skill-contributed `guards.d` guards, but a repo-wide grep shows only the bundled `guard_commit_trailer` currently reads `.raw`, so the fix's blast radius is that one guard.

AmirF194:
`split_segments()` builds every `Segment` in a compound command as `Segment(current, command)`, passing the whole input string as `raw` regardless of which segment it is. `guard_commit_trailer` searches `seg.raw` for `co-authored-by:`, so it was really searching the entire line, not the commit's own message. A `git commit` with a clean message gets denied if an unrelated earlier segment (a comment, a grep, an echo) happens to contain that phrase.

Fixed by giving each segment `raw = shlex.join(current)`, a reconstruction of that segment's own tokens only. It still supports the heredoc-survival case the docstring describes: `shlex.split` doesn't parse `<<` specially, so a heredoc's delimiter and body just become ordinary tokens attached to the segment they trail, and `shlex.join` puts them back as one searchable string. Checked the field's only other reader, `GuardContext.raw`: none of the four `guards.d`/skill-contributed guards use it, so the fix's blast radius is exactly the one guard this issue is about.

Added `test_compound_command_other_segment_not_denied`, the new test breaks without the change and passes with it. Full `tools/agent-guard` suite (92 tests), ruff, ruff format and mypy all come back clean in a `python:3.11-slim` container.

Fixes #1150

### DIFF
--- a/tools/agent-guard/src/agent_guard/__init__.py
+++ b/tools/agent-guard/src/agent_guard/__init__.py
@@ -113,8 +113,8 @@ class Segment:
     """One simple command from a (possibly compound) shell line.
 
     ``argv`` has leading ``NAME=value`` env assignments stripped into ``env``;
-    ``raw`` is the original text of the segment (used for substring scans that
-    survive heredocs, e.g. the Co-Authored-By trailer).
+    ``raw`` is a reconstruction of this segment's own tokens only (used for
+    substring scans that survive heredocs, e.g. the Co-Authored-By trailer).
     """
 
     def __init__(self, tokens: list[str], raw: str) -> None:
@@ -164,12 +164,12 @@ def split_segments(command: str) -> list[Segment]:
     for tok in tokens:
         if tok in SHELL_OPERATORS:
             if current:
-                segments.append(Segment(current, command))
+                segments.append(Segment(current, shlex.join(current)))
                 current = []
         else:
             current.append(tok)
     if current:
-        segments.append(Segment(current, command))
+        segments.append(Segment(current, shlex.join(current)))
     return segments
 
 
--- a/tools/agent-guard/tests/test_guards.py
+++ b/tools/agent-guard/tests/test_guards.py
@@ -218,6 +218,13 @@ def test_compound_command_guarded():
     assert dispatch('cd /tmp && git commit -m "x\nCo-Authored-By: a"') is not None
 
 
+def test_compound_command_other_segment_not_denied():
+    # The phrase lives in an unrelated earlier segment, not in the commit's
+    # own message, so the commit segment must not inherit it.
+    command = 'echo "note: never add a Co-Authored-By: trailer" && git commit -m "clean fix"'
+    assert dispatch(command) is None
+
+
 def test_malformed_command_allows():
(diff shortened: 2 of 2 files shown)


potiuk (approved):

potiuk (approved):
Good catch, and the diagnosis is exactly right: passing the whole `command` as every segment's `raw` meant a substring scan could not tell which segment the text came from, so any mention of the phrase anywhere on the line poisoned every other segment.

Since this is a security guard, I tested the false-negative direction specifically — narrowing `raw` could plausibly have weakened detection rather than just removing the over-match. It does not:

| Command | Result |
|---|---|
| `git commit -m "fix\nCo-Authored-By: ..."` | denied |
| `cd /tmp && git commit -m "x\nCo-Authored-By: a"` | denied |
| `git commit --amend -m "Co-Authored-By: x"` | denied |
| heredoc commit containing the trailer | **denied** |
| heredoc commit, clean message | allowed |
| `echo "...Co-Authored-By..." && git commit -m "clean fix"` | allowed (the fix) |

The heredoc row is the one that mattered: the docstring you updated claims `raw` exists for "substring scans that survive heredocs", so `shlex.join` had to preserve that property — and it does, because the heredoc body is already in the segment's own token list.

I also confirmed the bug on `main` before the fix: that last case is denied there today, so this is a real false positive being removed, not a hypothetical.

Docstring updated to match the new semantics, which is the part these changes usually miss.

---

AmirF194:
Good call testing the false-negative direction specifically, the heredoc row is exactly the one I'd have worried about if I'd only checked that the over-match went away.

PR #1157, opened 2026-09-08
Clarify ATR beta status and hybrid mode in runbook

dave2wave:
Updated release management instructions to reflect ATR's beta status and clarify hybrid mode.

## Summary

- ATR is now Beta software and this playbook must be updated.
- A draft start to try to nudge a conversation

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [ ] Other:

## RFC-AI-0004 compliance

### DIFF
--- a/docs/release-management/atr-release-runbook.md
+++ b/docs/release-management/atr-release-runbook.md
@@ -101,6 +101,7 @@ tired human at 2am, enforced the mechanics.
 > on ASF Slack before cutting a *real* release through ATR.
 
 > [!IMPORTANT]
+> **ATR now Beta** please update
 > **Hybrid mode (Magpie's current stance): SVN hosts, ATR votes.**
 > Because ATR is still alpha, Magpie does **not** yet let it host or
 > publish releases — but it *does* use ATR's automated checks and vote


potiuk:
Thanks for flagging this @dave2wave — you were right that the playbook needed updating, and the nudge worked.

Picked up in #1181, which does the full sweep: the ATR runbook alone stated the status in four places, and `manual-release-process.md`, `projects/magpie/release-management-config.md` and the adopter template `projects/_template/release-management-config.md` each repeated it.

While doing that sweep, one thing seemed worth separating out. Several passages made our hybrid stance a *consequence* of ATR's maturity — "because ATR is still alpha, Magpie does not yet let it host or publish releases" — which reads as though the policy moves on its own when the status moves. It doesn't: whether ATR hosts and publishes our releases is a PMC decision. So #1181 corrects the status, states the policy as a governance decision, and notes that the ratification vote on `dev@` is now the only remaining blocker. The policy itself is unchanged.

#1181 is deliberately held as a **draft**, because your PR raised a question larger than the wording. #1182 asks which alpha-era assumptions still hold under beta — client/API stability, which host to target, the hosting-trust argument, dual staging, and the `[VOTE]` template that we had to hand-assemble during 0.1.0. You and @sbp are cc'd there, and your input would be welcome; whatever comes back gets folded into #1181 before it lands.

Closing this one in favour of #1181. The prompt for the conversation is much appreciated.

Issue #314, opened 2026-05-26
feat(adapter/gemini): Google Gemini CLI skill runtime


potiuk:
**Goal:** Bring Google's Gemini CLI to feature parity with Claude Code as a skill runtime. Grounds: [RFC-AI-0004 Principle 3 — Vendor neutrality](../tree/main/docs/rfcs/RFC-AI-0004.md), which names Gemini CLI by name.

**Why Gemini CLI:** Open-source, Apache 2.0, official Google offering with a growing extension ecosystem. Strong reach into the Google-Cloud and Workspace adopter universe. Gemini CLI also has first-class MCP support, which means many of the framework's MCP-shaped integrations (Gmail, Slack, Google Drive, GitHub) should "just work" once the skill-format adapter lands.

**What "parity" means:**

- Skills under `.claude/skills/<name>/SKILL.md` are invokable from Gemini CLI
- The `tools/*` bridges are reachable (language-agnostic — port directly)
- Sandbox parity: Gemini CLI has a `--sandbox` mode; document the mapping to the framework's `bubblewrap` + network-allowlist baseline
- HITL primitives: Gemini CLI's "approve before run" mode is the native fit — verify and document

**Suggested approach:**

- Map `SKILL.md` → Gemini CLI's `.gemini/extensions/` extension format (or `GEMINI.md` for project-level instructions)
- Reuse the framework's MCP servers where Gemini supports them
- Document the skill-invocation pattern: how a maintainer runs a framework skill end-to-end via Gemini
- Map the `setup-isolated-*` sandbox baseline to Gemini's sandbox flags

**Reference:**

- RFC-AI-0004: [`docs/rfcs/RFC-AI-0004.md`](../tree/main/docs/rfcs/RFC-AI-0004.md)
- Gemini CLI: https://github.com/google-gemini/gemini-cli
- Gemini CLI extensions: https://github.com/google-gemini/gemini-cli/blob/main/docs/extension.md
- Existing skill shape: [`.claude/skills/`](../tree/main/.claude/skills/)

maheshsv:
I will work this issue

justinmclean:
@maheshsv do you still want to work on this?

sjyangkevin:
I'd like to work on this as my first contribution if it's still available.

potiuk:
Feel free.

sjyangkevin:
**Edited 2026-09-12**:

Reading the following two guides, and I feel it provides the answer.

[...]

PR #1205, opened 2026-09-13
feat(gemini): add tool sandboxing, policies and secure setup lifecycle

sjyangkevin:
## Summary

- Follow up on PR #1199 with Gemini CLI tool sandboxing, approval policies, existing BeforeTool guard integration, and static profile validation. Replace spec-loop's `--yolo` launch with Default Mode.
- Route install, verify, update, and doctor skills to Gemini-specific procedures for extension and snapshot installations. Add one Gemini routing scenario per skill while preserving existing evaluation criteria.
- Add `AGENT_ISO_ALLOW` so authentication can receive explicitly selected variables through the clean-environment wrapper, including API keys or Vertex AI project/location settings. Retain `CLAUDE_ISO_ALLOW` compatibility and consolidate setup instructions and runtime boundaries in `docs/adapters/gemini.md`.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] Other:
  - Full preflight passed after rebasing onto upstream `main` (`3c5c139f`), including workspace pytest, Ruff, mypy, links, and skill/spec validators. The local run used Python 3.11, a temporary jq installation, and a clean test directory outside `/tmp` to avoid its existing `.git` entry. Spec-loop runner fixtures also passed.
  - Agent-guard: **120 passed**. Sandbox-lint: **142 passed**, two optional integration tests skipped in the default run.
  - Separately enabled Gemini 0.59.0 integration checks passed: **328 policy decisions**, native read boundaries, workspace writes, outside-write refusal, and network refusal. These exercise runtime APIs independently of an authenticated model session.
  - All **eight Codex/Gemini routing cases passed** with exact comparison. Full affected skill eval suites: **72/78 passed**, with six failures in unchanged non-routing cases; expected answers were not adjusted.
  - Disposable extension and snapshot setups exercised discovery, settings loading, guard registration, and configuration merge/cleanup procedures.
  - Manual `slm-triage` pilot on Linux with Gemini 0.59.0 and Vertex AI ADC exercised all four lifecycle routes, instruction/skill loading, static lint, policy comparison, guard behavior, and shell/native-edit approval. Interactive Plan Mode checks were intentionally deferred.

## RFC-AI-0004 compliance

## Linked issues

Refs #314.

**Runtime boundaries:** Native file tools and approved shell commands have different read boundaries. On the tested Linux backend, approved shell commands can read host files outside the workspace; native credential-path policies do not prevent equivalent shell reads. Tool sandboxing does not isolate the whole CLI, hooks, or MCP servers. Existing sandbox grants can widen access. Headless spec-loop operations requiring approval are denied.

**Open pilot findings:** Shell commands completed but repeatedly emitted missing `bgpids.tmp` warnings. An outside-write probe requested expanded access, including writes to `/usr/bin` and Python's library directory; expansion was declined and no file appeared. The cause remains unresolved. The pilot used bubblewrap **0.6.1**, below Magpie's **0.11.2** pin; upgrading has not been shown to resolve these symptoms.

**Skill reliability:** Routing worked, but the interactive pilot required corrections to the linked-extension guard path, Python interpreter selection, linter fallback, and verification/diagnostic claims. These remain review findings.

**Validation limits:** Some policy-denial results were reported by Gemini without captured tool output. OS write/network enforcement and environment stripping were not independently established in the manual pilot. Runtime validation covers Linux with Gemini 0.59.0; macOS and Windows remain unverified.

**AI disclosure:** Implementation and PR text were assisted by Codex (GPT-6); skill evaluations used Claude, and the manual setup pilot used Gemini CLI.

## Test Screenshots

Screenshot from 2026-09-12 23-23-00


potiuk (changes requested):
## Rebased onto `main`

I've rebased this branch onto current `main` (now at `5cf0e09a`) and force-pushed it — the conflict was in `CONTRIBUTING.md`'s harness table, where #1206 added the Local LLM row while this PR rewrote the Gemini row. I took each side's own row, nothing dropped.

## One blocker

**`grep_search` is not a Gemini CLI tool, so the credential deny rules are inert.**

`.gemini/policies/magpie.toml:45,52` and `tools/sandbox-lint/src/sandbox_lint/gemini.py:66` list `grep_search`. The canonical name is `search_file_content`:

```js
// gemini-cli-core/dist/src/tools/tool-names.js
export const GREP_TOOL_NAME = 'search_file_content';
// gemini-cli-core/dist/src/tools/grep.js:441
static Name = GREP_TOOL_NAME;
```

`grep_search` occurs zero times in the 0.59.0 bundle. This isn't cosmetic, because Gemini's default tier already allows the tool outright:

```toml
# policy/policies/read-only.toml
[[rule]]
toolName = "search_file_content"
decision = "allow"
priority = 50            # effective 1.050
```

Your priority-900 denies are effective 2.900 and would win — but they never match. So `search_file_content` falls through to the built-in allow with no prompt, and no `ask_user` rule covers it either (those are scoped to `run_shell_command`, `write_file`, `replace`, MCP `*`). Since grep returns matching file *content*, every in-workspace file the policy sets out to protect is readable through it: `.env`, `.env.local`, `.gemini/settings.json`, `.gemini/policies/**`, `.npmrc`, `.pypirc`.

The same typo in `NATIVE_TOOLS` means the linter validates the fiction and passes. Worth fixing in three places, not two — `gemini_runtime.mjs`'s `samples` has no grep entry, and that native probe is the one check that would otherwise have caught this. A `['search_file_content', { pattern: 'x', path: '.env' }, 'deny']` sample would keep it from recurring.

## One to decide

`google_web_search` is also `allow`/priority 50 in `read-only.toml` and has no rule here. With `sandboxNetworkAccess: false` closing shell egress, it's a remaining outbound channel carrying model-controlled text — and `permission-audit` already treats the equivalent Kiro tool as an exfiltration channel. Deliberate, or an oversight? (`web_fetch` and `save_memory` are fine — both default to `ask_user` in `write.toml`.)

## Otherwise

This is strong work. The `\\\\` handling for JSON-escaped Windows paths is correct, the `gh` allow regex properly excludes shell metacharacters so `&&`, `>`, and `$(...)` fall back to `ask_user`, and driving the real `PolicyEngine` across 4 modes × 2 interactivity settings is a far better test than static assertions — that harness is why I trusted the rest of the policy quickly.

I verified all of the above against the 0.59.0 bundle installed locally, matching the version your integration test pins.

sjyangkevin:
Thanks for the review. Addressed in commit `d54634ca`.

### grep_search

I checked the official `v0.59.0` source, which defines `grep_search` as canonical and `search_file_content` as its legacy alias. I retained the canonical name and added native tests proving that both names match the credential denies. The source evidence is linked in the runtime guide as well. [Gemini CLI's document](https://geminicli.com/docs/reference/tools/#file-system) also shows `search_file_content` is a legacy alias.

### regression coverage

The probe now checks canonical policy names, validates search and multi-file argument schemas, and verifies that protected-path requests match Magpie’s explicit deny rule. This also corrects the previous probe’s argument keys.

- web search: added an explicit priority-600 `ask_user` rule for `google_web_search`. Every search requires approval, including in Plan mode, and headless requests are refused. The linter requires this rule, with regression tests covering its removal or weakening.
- updated the documentation and specification. Validation passed: 149 tests, including 464 native policy decisions across four approval modes and both interactivity settings, plus the applicable commit hooks.

potiuk (approved):

PR #1216, opened 2026-09-14
fix(gemini): require approval for Gemini web fetches and MCP resource reads

sjyangkevin:
## Summary

- Gemini's built-in policy allows `web_fetch` without approval in auto-edit. Native `read_mcp_resource` also defaults to allow and does not match Magpie's MCP server-tool wildcard.
- Add explicit approval rules for both tools in every mode, including auto-edit and Plan Mode; headless calls are refused. Cached resource listing remains allowed.
- Extend the linter and native runtime regression tests, and document the approval behavior in the adapter and sandbox spec.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] For Python packages touched: pytest, Ruff, and mypy pass; sandbox-lint has 159 passing tests and one optional live Linux sandbox test skipped.
- [x] Other: Gemini CLI 0.59.0's native PolicyEngine passes 488 decisions across four modes and interactive/headless execution, including explicit approval-rule matching and headless refusal. The new resource-read assertion failed against the previous policy.

## RFC-AI-0004 compliance

- [x] **HITL** — both tools require explicit per-call approval, including in Plan Mode
- [ ] **Vendor neutrality** — placeholders (`<PROJECT>`, `<tracker>`, `<upstream>`, `<security-list>`) used in all skill / tool prose

## Linked issues

Ref: #314 

Plan Mode permits these reads with approval to support gathering context; ordinary MCP server tools remain denied there. The tests evaluate native policy decisions without authenticated model calls.

**Generative AI disclosure:** Codex (GPT-6) assisted with implementation, documentation, tests, and preparation of this PR description.

## Screenshots

By default, `read_mcp_resource` is `allow` in all modes, and `web_fetch` is `allow` in auto-edit mode.

Screenshot from 2026-09-13 19-59-23

`web_fetch` runs without explicit asking for approval in auto edit mode.

Screenshot from 2026-09-13 19-51-59

`read_mcp_resource` also runs without explicit asking for approval in `default` approval mode.

Screenshot from 2026-09-13 19-59-01

The PR fixes the above by enforcing it to ask for approval through policies.


potiuk (approved):

potiuk:
Very cool

andreahlert:
LGTM thanks!

potiuk (approved):
Reviewed against the branch locally, drove the `sandbox-lint` invariants by hand, and verified the premise against upstream rather than the PR description. **LGTM** — the change is correct and strictly narrows the posture.

### Verification

`packages/core/src/policy/policies/read-only.toml` at `v0.59.0` lists `read_mcp_resource` and `list_mcp_resources` — alongside `google_web_search` — in the built-in read-only `allow` block at priority 50, i.e. default tier `1.050`. The new User-tier rules land at `4.600` and win in every mode, so this is a narrowing in Plan Mode too, not a widening. `web_fetch` is *not* in that list, consistent with the claim that it was the auto-edit allowance doing the work there. All three names are real constants in `tool-names.ts` at that tag, so the `ALL_BUILTIN_TOOL_NAMES` assertion in the probe will hold.

I exercised `check_gemini_invariants` directly against the branch profile: baseline is clean, and all 15 mutation cases (3 tools x `missing` / `allow` / `default_only` / `interactive_only` / `args_only`) produce the expected error. The `query_only` -> `args_only` rename is just the parametrisation generalising past the web-search-specific name.

The new priority assertion is the best part of the diff. Pinning `result.rule?.priority === USER_POLICY_TIER + 0.6` across all four modes catches the silent-widening case where a future built-in rule outranks the shipped one and the decision still happens to read `ask_user`.

### Non-blocking notes

1. **The strongest guard is opt-in.** The auto-edit override for `web_fetch` is the actual security claim here, and the only thing proving it is `gemini_runtime.mjs`, which skips unless `MAGPIE_GEMINI_BUNDLE` is set. `sandbox-lint` can assert the rule exists, not that it outranks the built-in. That is the pre-existing pattern from the `google_web_search` work, so not a regression — but it does mean every Gemini upgrade needs a manual probe run, which is worth stating explicitly somewhere durable.

2. **The samples are not schema-validated.** `web_fetch` and `read_mcp_resource` are not in the `schemas` map, so `{ prompt: ... }` and `{ uri: ... }` are never checked against the runtime's declared parameters — an upstream argument rename would pass silently. It does not affect matching today (neither rule uses `argsPattern`), but the file already has the machinery and this PR is precisely about asserting against the real runtime. Cheap to add.

3. **Linter error string.** `web_fetch` / `read_mcp_resource` now inherit `"(shell requires the command argument matcher)"`, which is nonsense for them. Pre-existing for `write_file` / `replace`, but the loop is up to six tools with one shell-specific parenthetical.

4. **Worth a sentence in the docs:** leaving `list_mcp_resources` allowed discloses the configured servers' resource registry to the model without approval. Low risk and the doc is explicit — but the stated justification ("lists the cached resource registry without making a resource-read request") is an assertion about upstream behaviour that nothing in the suite pins.

Nice cross-adapter consistency, worth calling out: `sandbox_lint/kiro.py` already flags auto-approved `web_fetch` as "an exfiltration channel". This brings Gemini in line with a posture Magpie enforces elsewhere.

---

**Generative AI disclosure:** this review was prepared with Claude Code (Opus 5); the upstream claims above were verified against `google-gemini/gemini-cli` sources at `v0.59.0`.

PR #1229, opened 2026-09-15
fix(agent-isolation): read the main checkout's settings in a linked worktree

johnslavik:
## Summary

- **The status line claimed a sandbox a worktree session did not have.** Inside a linked git worktree it walked `<cwd>/.claude/` and then user scope — but Claude Code scopes a linked worktree's project to the **main checkout**, which is where `/sandbox` persists the toggle and where `.claude/.cc-writes` lands. The worktree's own `settings.local.json` holds only the per-worktree filesystem allowlist `sandbox-add-project-root.sh` writes, with no `enabled` key, so the walk fell through to `~/.claude/settings.json` and painted a green `[sandbox]` on a session the operator had deliberately switched *out* of the sandbox. That is the one failure mode this status line exists to prevent.
- **Fix:** resolve the working-tree root and the main checkout by comparing `git rev-parse --git-dir` with `--git-common-dir` — they differ in a linked worktree and nowhere else — and insert both between `<cwd>` and user scope in the settings walk. The working-tree root also closes a latent miss: a `<cwd>` *below* the tree root has no `<cwd>/.claude/` to read, so project scope was skipped there too.
- **Same resolution replaces the hardcoded `*/.claude/worktrees/*` substring match** in the folder segment, so `<source>/<worktree>` rendering covers every layout instead of the single one it enumerated: worktrunk's sibling `<repo>.<branch>/` directories (repeated repo prefix stripped, so `cpython.gh-156021` reads `cpython/gh-156021`), Claude Code's own `.claude/worktrees/<name>`, and a plain `git worktree add`. Because the test is a property of git rather than of a naming scheme, a worktree manager does not have to be enumerated here to be supported.

Before / after, in a worktrunk worktree whose main checkout has `"enabled": false`:

```text
before   [sandbox]     cpython.gh-156021 | gh-156909 | Opus 5
after    [NO SANDBOX]  cpython/gh-156021 | gh-156909 | Opus 5
```

## Type of change

- [x] Python package (`tools/*/` with `pyproject.toml`) — new test module under `tools/agent-isolation/tests/`
- [ ] Groovy reference impl
- [x] Cross-cutting (RFC, AGENTS.md, sandbox, privacy-LLM) — sandbox-state visibility
- [ ] Other:

## Test plan

- [ ] `prek run --all-files` passes — **not run** in the authoring environment; please let CI be the gate.
- [x] For Python packages touched: `uv run pytest` passes in `tools/agent-isolation/` — 136 passed, 14 of them new in `tests/test_sandbox_status_line.py`.
- [x] Other — see below.

New test module covers, against real `git worktree` fixtures built in `tmp_path`:

- settings precedence: project `true` / `false` / `autoAllowBashIfSandboxed`, fall-through to user scope, non-git directory;
- the regression itself — worktree with no `enabled`, main checkout `false`, user scope `true` must render `[NO SANDBOX]`;
- a `<cwd>` below the worktree root, and below a plain checkout's root (must *not* be mistaken for a worktree — this caught a real bug in the first draft, where `--git-dir` comes back absolute from a subdirectory and relative from the root, so the raw string compare reported every subdirectory as a linked worktree);
- folder rendering per layout: `repo.branch`, `repo-branch`, `.claude/worktrees/<name>`, an unrelated worktree name, and a plain checkout.

Also exercised by hand against the real worktrunk checkouts that surfaced the bug (main, worktree, and subdirectories of each).

## RFC-AI-0004 compliance

- [x] **Sandbox** — no new host access. The script reads two to four more local `settings*.json` paths and runs two extra `git rev-parse` calls; no network, no writes.
- [ ] **Vendor neutrality** — placeholders used in all skill / tool prose
- [ ] **Privacy LLM** — private content does not reach a non-approved LLM

## Linked issues

None — found in use, not filed first.

## Notes for reviewers

- **How the main-checkout behaviour was established.** Not from Claude Code's source, which is not readable here, but by elimination on a live session: the `/sandbox` panel showed *No Sandbox* in a worktree session whose own `settings.local.json` has no `enabled` key and whose user scope has `"enabled": true` — the only `false` in reach was the main checkout's. Corroborated by `.claude/.cc-writes` existing in the main checkout and in no worktree. Worth a second opinion from anyone who can confirm it against the harness directly.
- **Ordering choice.** `<cwd>` and the worktree root are walked *before* the main checkout, so an `enabled` explicitly set in a worktree still wins. Claude Code appears to read only the main checkout, so the two agree wherever the main checkout sets the key — which is the common case and the one the bug report came from. The divergence is a worktree file that sets `enabled` by hand; there, this errs toward the more specific file.
- **Spec sync.** `tools/spec-loop/.last-sync` is 19 commits behind `main`, which predates this change. No spec under `tools/spec-loop/specs/` mentions the status line or worktrees, so nothing here needs a spec update — flagging the drift rather than folding an unrelated sync into this PR.


tools/agent-isolation/tests/test_sandbox_status_line.py
```diff
@@ -0,0 +1,250 @@
...
+    )
+    _git(path, "commit", "-q", "--allow-empty", "-m", "init")
+    return path
+
+
+def _settings(path: Path, sandbox: dict) -> None:
+    """Write <path>/.claude/settings.local.json with a sandbox block."""
+    claude = path / ".claude"
+    claude.mkdir(parents=True, exist_ok=True)
+    (claude / "settings.local.json").write_text(json.dumps({"sandbox": sandbox}))
```
johnslavik, line 99: Why only test `.claude` and not other harnesses? Is status line CC-only?

tools/agent-isolation/tests/test_sandbox_status_line.py
```diff
@@ -0,0 +1,250 @@
...
+
+    def test_unrelated_worktree_name_is_left_alone(self, tmp_path: Path, home: Path) -> None:
+        repo = _init_repo(tmp_path / "repo")
+        wt = tmp_path / "elsewhere"
+        _git(repo, "worktree", "add", "-q", "-b", "feature-x", str(wt))
+        assert _folder(_run(wt, home, tmp_path)) == "repo/elsewhere"
+
+    def test_main_checkout_renders_one_segment(self, tmp_path: Path, home: Path) -> None:
+        repo = _init_repo(tmp_path / "repo")
+        assert _folder(_run(repo, home, tmp_path)) == "repo"
```
johnslavik, line 260: Can we make all these tests parametrized? Probably could be done without classes `@pytest.mark.parametrize`

tools/agent-isolation/tests/test_sandbox_status_line.py
johnslavik: (Opus 5)

Yes — CC-only, and you're right that the file never said so. Fixed in 9104082: the script header and the test docstring now state it outright.

The helper is Claude Code-specific three times over: it is wired through Claude Code's `statusLine` setting, it parses Claude Code's statusLine JSON payload on stdin (`cwd`, `model.display_name`), and it reads Claude Code's `sandbox.enabled` schema. No other harness in the repo has a status-line hook of that shape — `statusLine` appears nowhere outside Claude Code wiring:

```
$ grep -rn "statusLine" --include="*.md" --include="*.json" --include="*.sh" . | grep -vi claude | grep -v skill-evals
CONTRIBUTING.md:509:  shell scripts (`agent-iso.sh`, sandbox status-line helpers,
```

(the only hit is a prose mention of this script; the skill-evals hits are fixtures asserting this script's own install.)

Codex, Gemini, OpenCode and Kiro carry their sandbox posture in their own config — `.gemini/settings.json` + `policies/magpie.toml`, and so on — and surface it, where they surface it at all, through their own UI. So `.claude/` is the only settings directory the tests can meaningfully exercise. If a harness grows a status line it gets its own helper and its own tests, per `docs/adapters/add-a-harness.md`.

One thing worth flagging: `tools/agent-isolation/README.md:24` still says **Harness: agnostic** for the whole directory, which is now true of every script in it except this one. Left alone here rather than widening this PR — say the word and I'll fix it in this branch or a follow-up.

tools/agent-isolation/tests/test_sandbox_status_line.py
johnslavik: (Opus 5)

Done in 9104082 — classes gone, one parametrized test.

Every case was the same three steps with different data: build a layout under `tmp_path`, run the script, assert on one segment. That is now a frozen `Case` dataclass, a single `_build`, and one `@pytest.mark.parametrize`. 13 cases, each named by its `id`, and a new layout is one entry instead of a method.

It also caught something the classes hid: each old test asserted only the segment it was written for, so the folder-layout tests never checked the tag and vice versa. Every case now asserts both.

```
$ uv run --directory tools/agent-isolation pytest tests/test_sandbox_status_line.py -q
.............                                                            [100%]
13 passed
```

And against the pre-fix script, to confirm the cases still bite:

```
$ git show HEAD~1:tools/agent-isolation/sandbox-status-line.sh > sandbox-status-line.sh
$ uv run --directory tools/agent-isolation pytest tests/test_sandbox_status_line.py -q
FAILED test_status_line[main-checkout-subdirectory-is-not-a-worktree]
FAILED test_status_line[worktree-reads-main-checkout]
FAILED test_status_line[worktree-subdirectory-reads-main-checkout]
FAILED test_status_line[worktree-own-settings-win-over-main]
FAILED test_status_line[folder-worktrunk-dot-separator]
FAILED test_status_line[folder-worktrunk-dash-separator]
FAILED test_status_line[folder-unrelated-worktree-name-left-alone]
```

7 of 13 — the four settings-walk cases and three of the four layout cases. `folder-claude-code-worktrees-layout` passes on both, as it should: that is the one layout the old substring match already knew.

johnslavik:
I don't like parametrize usage here, let's keep class-based tests.

tools/agent-isolation/tests/test_sandbox_status_line.py
johnslavik: Yes, let's document this.

cc @potiuk

tools/agent-isolation/tests/test_sandbox_status_line.py
johnslavik: (Opus 5)

Reverted in 3a0f9bc — the class-grouped tests are back, so my reply above no longer describes the branch. 14 tests, classes as before.

The Claude-Code-only docstring note added in the same commit is kept.

tools/agent-isolation/tests/test_sandbox_status_line.py
johnslavik: (Opus 5)

Documented in a337b03, in three places:

- `tools/agent-isolation/README.md:24` — the directory tag now reads **Harness:** agnostic, except `sandbox-status-line.sh` — Claude Code only, so the exception is visible at the same glance as the claim it qualifies.
- The script's Files row — why it is the exception (wired through `statusLine`, fed Claude Code's payload, reads Claude Code's `sandbox.enabled` schema), and that Codex, Gemini, OpenCode and Kiro carry their posture in their own config.
- `docs/setup/secure-agent-setup.md` → Sandbox-state status line — same note, plus a pointer to the harness-agnostic half of sandbox visibility: the bypass-warning hook, which fires on the tool call rather than in the footer.

All three point at `docs/adapters/add-a-harness.md` for a harness that grows a status line.

Suite still green:

```
$ uv run --directory tools/agent-isolation pytest tests/ -q
136 passed
```

johnslavik: Revert a337b03 if you'd like this out 👍🏽

potiuk:
Ha... thanks.. I just realised that bug was also affecting **my Mac** :D

tools/agent-isolation/tests/test_sandbox_status_line.py
```diff
@@ -0,0 +1,260 @@
+# Licensed to the Apache Software Foundation (ASF) under one
```
potiuk, line 1: Nice! More tests :)

potiuk (commented):
Really nice fix — comparing `--git-dir` with `--git-common-dir` is the right primitive, and it's the kind of test that stays true when someone invents a new worktree manager. The test module is genuinely good.

Two things I found poking at it. I've pushed a fixup for both rather than sending you round again — shout if you disagree with either and I'll drop it.

**1. The walk order could paint a false green.** The premise of the change is that Claude Code scopes a linked worktree's project to the main checkout. But `<cwd>` and `<worktree-root>` are walked *before* `main_root`, so an `enabled` written by hand into a worktree's own settings wins over the checkout the harness actually reads:

```
worktree      .claude/settings.local.json = {"enabled": true}
main checkout .claude/settings.local.json = {"enabled": false}
→ [sandbox] repo/feat        # real state, per your own model: NO SANDBOX
```

`test_worktree_own_settings_win_over_main` locks that in, which sits awkwardly with the module docstring two screens above: *"it must never claim a sandbox the session does not have."* "More specific wins" is the right instinct for most settings, but this file is a fail-safe indicator — the more specific file here is precisely the one that can't describe the session. The fixup puts the main checkout first and inverts that test.

**2. Bare-repo worktrees resolved to an arbitrary directory.** `main_root=$(cd "$gcommon/.." …)` assumes the common dir is `<main>/.git`. For `git clone --bare` plus `git worktree add` it's `<repo>.git`, so the parent is whatever directory happens to contain the bare repo:

```
--git-common-dir : …/container/proj.git
derived main_root: …/container        ← not a checkout
→ [NO SANDBOX] container/feature      ← read container/.claude as project scope
```

The folder name is wrong every time in that layout, and a stray `.claude/settings.json` one level up gets read as project scope — which can send the tag green as easily as red. The fixup guards on the `.git` suffix: a bare common dir yields no project scope (fall through to user scope, which is the honest answer), while the folder segment still names the repo from the bare dir, so `proj.git` renders `proj/feature-x`.

Also folded in: guard the comparison on a non-empty `--git-dir`, so a `--git-dir` that fails to resolve stops reading as "linked worktree".

Nit, not fixed: section 3 re-runs `rev-parse --git-dir` that section 1 already resolved. One `git rev-parse --show-toplevel --git-dir --git-common-dir` would do all three in a single call — worth it in something that runs on every render, but not worth churning this PR for.

---

> *This review was drafted by an AI-assisted tool and confirmed by a Magpie maintainer. The findings above are observations, not blockers; a Magpie maintainer — a real person — will take the next look at the PR. If you think a finding is mis-applied, please reply on the PR and a maintainer will weigh in.*
>
> *More on how Magpie handles maintainer review:* [CONTRIBUTING.md § Opening a pull request](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md#opening-a-pull-request).

johnslavik:
(Sonnet 5)

Pulled in your fixup commit and checked both fixes against the branch:

- Walk order: the main checkout's settings now come first, so a hand-edited worktree `enabled` can no longer outvote it. Confirmed with `test_main_checkout_beats_a_hand_written_worktree_setting`.
- Bare repo: the `.git`-suffix guard means a bare common dir falls through to user scope instead of reading its parent as project scope, and the folder name still renders from the bare dir's basename. Confirmed with `test_bare_repo_worktree_does_not_read_the_bare_dir_s_parent`.

Also confirmed the non-empty `--git-dir` guard is in place.

Ran the full suite locally:
```
$ uv run pytest tests/test_sandbox_status_line.py -v
============================= 15 passed in 12.09s ==============================
```

Thanks for pushing the fixup directly, no changes needed from my side. Leaving the double `rev-parse` nit as-is per your call.

PR #1247, opened 2026-09-16
feat(cve-json)!: require a purl on every affected entry

potiuk:
## Summary

[#1197](https://github.com/apache/magpie/pull/1197) added `affected[].packageURL` as **opt-in, off by default**, so that "an adopter who sets neither gets byte-identical records to today". That backward-compatibility choice is what this PR revisits.

A CVE record needs a purl to be promoted by the CNA. With the field off by default, the **default configuration silently produces records that are rejected later** — after the ID is allocated, after the advisory is drafted, at the point where going back is most expensive. The schema calls `packageURL` optional; the promotion path does not.

## What changed

A purl is now produced for every entry, by three means in order:

1. **`product.purl_type`** when configured — unchanged behaviour.
2. **Derived from `product.default_collection_url`** when unset. An adopter publishing to PyPI has already said so; having to repeat it in a second key is exactly why the field stayed unset. Only hosts whose type is unambiguous are mapped — PyPI, npm, crates.io, RubyGems, NuGet — because a guessed type yields a valid-looking identifier pointing at the **wrong ecosystem**, which is worse than asking.
3. **A `ValueError` naming the key to set**, instead of emitting a record without the field.

`purl_type = "none"` opts out deliberately — a Helm chart served from the project's own site, for instance, where the purl spec models no type. The distinction that matters: a record without a purl is now a decision someone wrote down, rather than a default nobody noticed.

This matches the posture the version parser already takes — it refuses to emit a record whose `version` is a range expression rather than shipping JSON the CNA will reject.

### Only the purl

Follow-up from review. [apache/security-vulnogram#253](https://github.com/apache/security-vulnogram/pull/253) promoted the Package URL to the **recommended** identifier and derives `collectionURL` / `packageName` from it on publication when both are absent — while reporting a stored pair that disagrees with the purl. Writing both leaves two identifiers free to drift, so an entry carries one:

- `packageURL` alone for every product that has a purl;
- `collectionURL` / `packageName` only for `purl_type = "none"`.

Two consumers stopped assuming `packageName`: the attachment table (reads the name back out of the purl; column renamed *PyPI URL* → *Project URL*) and the `oauth-api` push guard (normalises both shapes, so it doesn't refuse pushes to records authored before the switch).

## Breaking change

An adopter whose `purl_type` is unset **and** whose `default_collection_url` is not a recognised package host now gets an error instead of a record without a purl. The error names all three remedies. Adopters on PyPI/npm/crates.io/RubyGems/NuGet are unaffected — derivation covers them with no config change.

`affected[]` entries also no longer carry `collectionURL` / `packageName` unless the product is configured `purl_type = "none"`; the CVE tool derives the pair from the purl on publication.

## Type of change

Feature + breaking change.

## Test plan

- **334 tests pass** in `generate-cve-json`, **99** in `oauth-api`.
- Four existing tests encoded the old opt-in contract and were rewritten to the new one: two "omits the field without a type" cases became derivation cases; the namespace-required case now asserts the error names `purl_namespace`; the Helm-chart fallback case now uses `purl_type = "none"`, which documents the escape hatch in a realistic setting rather than an invented one.
- The `build_affected` tests that asserted `packageName` now assert the purl, since that is the identifier the entry carries.
- New: derivation parametrised across five hosts plus an unknown host and the empty string, configured-beats-derived, the `"none"` opt-out, and the raise-with-actionable-message case; the identifier shape (purl alone vs. the legacy pair) and its alphabetical key order; `split_purl` round-trips including an npm scope and a hand-edited purl carrying qualifiers; the attachment table deriving an npm package and link from a purl, and rendering `—` for a purl type with no project page; and the push guard treating a purl and the `packageName` it derives as the same package.
- `prek run --files …` clean: ruff check, ruff format, mypy, pytest, markdownlint, lychee.

Found while syncing eight CVEs on an adopter tracker: the records had been generated, pushed and advisories sent before anyone noticed the field was absent.


ppkarwasz:
What we did in apache/security-vulnogram#253 is to promote Package URL to the **recommended** identifier.

Therefore the skill should rather stop producing `collectionURL / packageName` and provide only a `packageURL` field. The legacy identifiers will be populated during publication to MITRE. There is no need to fill them, unless there is no Package URL for your ecosystem.

potiuk:
> What we did in [apache/security-vulnogram#253](https://github.com/apache/security-vulnogram/pull/253) is to promote Package URL to the **recommended** identifier.
> 
> Therefore the skill should rather stop producing `collectionURL / packageName` and provide only a `packageURL` field. The legacy identifiers will be populated during publication to MITRE. There is no need to fill them, unless there is no Package URL for your ecosystem.

Thanks. On it.

potiuk:
Done in 22daa97a — entries now carry `packageURL` alone, and `collectionURL` / `packageName` only for a product configured `purl_type = "none"`, where there's no purl to derive them from.

Two knock-ons worth flagging: the issue-attachment table's package and link cells now read the name back out of the purl and map its type to a project page (PyPI, npm, crates.io, RubyGems, NuGet), and the `oauth-api` push guard — which compares `(package, product)` against the already-published record — normalises both identifier shapes to the same package, or it would refuse every push to a record authored before the switch.

potiuk:
This will also allow us to update past CVEs as explained in slack.

potiuk:
(though - as discussed - I am not going to do it for now - it's a possiblity and we might decide to do or not to do).

ppkarwasz:
> This will also allow us to update past CVEs as explained in slack.

These should be carefully planned: every update to a CVE record triggers a new review in NVD and probably GitHub too.

I have seen at least an example, where my PR to GitHub Advisories was merged, but part of the update was dropped, due to an update of the record from NVD.

PR #1289, opened 2026-09-19
feat(agent-isolation): hint when gh ran inside the sandbox and catalogue the TLS -26276 shape

potiuk:
## Summary

- A `gh` call that does not get excluded from the sandbox fails on macOS with `x509: OSStatus -26276` (or `HTTP 401`): Go verifies TLS through Security.framework and the Seatbelt profile blocks the trustd / keychain mach services. This cannot be fixed in-sandbox (no fallback roots in the Homebrew `gh`, `SSL_CERT_FILE` ignored on darwin, no mach-service setting in Claude Code), so the `gh *` exclusion is the only remedy — and it applies only when **every part** of the Bash invocation is `cd …` or `gh …`. A pipe, `$(…)`, a loop, or any file redirection (even `> /dev/null`) puts `gh` back in the sandbox. The redirection case is a Claude Code regression, reported as [anthropics/claude-code#95532](https://github.com/anthropics/claude-code/issues/95532).
- This PR makes that failure self-explaining at the moment it happens: a new catalog entry in `docs/setup/sandbox-troubleshooting.md` (measured shape table, root cause, the `gh tofile` alias workaround), a matching `[sandbox-hint]` signature in `sandbox-error-hint.sh`, a sixth doctor probe with an eval fixture, an eleventh verify check, and a new spec `tools/spec-loop/specs/sandbox-diagnostics.md` that describes the catalog / hook / doctor / verify surface as one contract.
- Drive-by: the hint hook had no tests (14 added, covering all five signatures and the fail-open paths); the hint-signature table in `secure-agent-setup.md` was missing the `/tmp` row.

## Type of change

- [ ] Groovy reference impl
- [ ] Other:

## Test plan

- [x] `prek run --files <all changed files>` passes every hook (rebased on #1288, which fixed the pre-existing `test_gpg_touch_overlay` failure that blocked the earlier commit)
- [x] `uv run --project tools/agent-isolation --group dev python -m pytest tools/agent-isolation/tests/test_sandbox_error_hint.py` — 14 passed (written red first: the four gh cases failed before the hook change)
- [x] Invocation-shape table measured on Claude Code 2.1.278 / macOS 26, one Bash tool call per row, sandbox on
- [x] `gh tofile` alias verified from a sandboxed call: writes to the scratch tree, refuses a path outside `$PWD` / the scratch tree with exit 2
- [x] Skill evals: print-mode extraction confirmed for `setup-isolated-setup-doctor` (`interpret-probes`, `after-report`) and `setup-isolated-setup-verify` (`step-1-classify`) — the renamed sections extract and include the new probe / check
- [x] For skill *behaviour* changes: `interpret-probes` gains a `gh_sandbox_status` field (schema extended the same way #1288 added `signing_key_status`), the eight existing fixtures carry `"skip"` for it, and `case-9-gh-sandbox-fail` covers the ✗ shape
- [x] Spec: `spec-validate` and `check-doc-sync` pass; the spec's own validation block cross-checks every hook anchor against the catalog headings
- [x] Token table regenerated (`skill-token-count --write`)

## RFC-AI-0004 compliance

- [x] **HITL** — the new probe and check are read-only; the `gh tofile` alias is documented as an operator-installed convenience and refuses paths outside the working directory / scratch tree
- [x] **Sandbox** — no new host access; the change documents an existing exclusion and when it does not apply
- [x] **Vendor neutrality** — no project-specific values introduced
- [ ] **Conversational + correctable** — n/a
- [ ] **Write-access discipline** — n/a
- [ ] **Privacy LLM** — n/a

## Linked issues

Refs [anthropics/claude-code#95532](https://github.com/anthropics/claude-code/issues/95532) (the upstream redirection regression), [anthropics/claude-code#85857](https://github.com/anthropics/claude-code/issues/85857) (trustd mach-lookup root cause). Follows #1288 (Probe 5 / check 10 there; this PR adds Probe 6 / check 11).

- Spec-sync pre-check: no spec under `tools/spec-loop/specs/` named the hint hook, the doctor probes, or the verify checks, so this PR adds `sandbox-diagnostics.md` for that surface (catalog as source of truth; hook, doctor and verify as discoverability layers; the `gh` exclusion shape rule as contract), cross-links it from `agent-isolation-sandbox.md`, lists it in both spec indexes, and bumps `.last-sync` to the base `main` tip.
- The invocation-shape rule is measured behaviour on 2.1.278 and will change when #95532 is fixed; the catalog entry, the probe text and the verify check name the issue so they can be edited together on that day.


potiuk:
Hopefully one of the last to avoid unnecessary prompts and bypassing sandbox.

PR #1297, opened 2026-09-20
feat(tools/osv): query OSV.dev for vulnerability cross-references during security triage (#311)

### LINKED ISSUE #311: feat(tools/osv): query OSV.dev for vulnerability cross-references during security triage
Add `tools/osv/` as a read-only bridge to **OSV.dev** (Google's open-source vulnerability database). Complements the existing `tools/cve-org/` and `tools/vulnogram/` by providing OSV's per-ecosystem (PyPI, npm, Maven, Go, RubyGems, crates.io, Packagist, …) vulnerability records.

**Suggested capabilities:**

- `query --package <name> --ecosystem <pypi|npm|maven|...>` — list known vulns for a package
- `query --commit <sha> --repo <url>` — does a commit fix or introduce a known vuln?
- `vuln <OSV-ID>` — full record (references, affected ranges, fix commits, aliases including CVE/GHSA)
- `aliases <id>` — cross-walk CVE ↔ GHSA ↔ OSV (the alias graph is the most valuable single piece of OSV)

**Why:** During `security-issue-triage`, before allocating a CVE, it's high-value to check whether OSV already has a record — to detect duplicates, to find the matching GHSA, and to find which downstream consumers are already warned. OSV is the **best machine-readable vulnerability cross-reference** available, and integrates cleanly with the existing CVE workflow.

Concrete consumers:

- `security-issue-triage` — early dedupe against the known-vuln set
- `security-issue-deduplicate` — alias resolution before merging trackers
- `security-cve-allocate` — sanity check that no existing OSV/GHSA already covers the report

**Reference:**

- OSV API: https://google.github.io/osv.dev/api/
- Existing CVE / vulnerability tooling: [`tools/cve-org/`](../tree/main/tools/cve-org/), [`tools/vulnogram/`](../tree/main/tools/vulnogram/)

Kaap10:
## Summary

- Implements `tools/osv/` (`contract:security-cross-ref`) as a read-only adapter bridging the Open Source Vulnerability (OSV.dev) database to Magpie.
- Provides POSIX `curl` + `jq` operations to resolve vulnerability aliases (CVE ↔ GHSA ↔ OSV IDs), query advisories by package/version, and verify commit fixes without leaking embargoed reports.
- Closes the open `contract:security-cross-ref` extension point across the registry, taxonomy docs, validator, and vendor-neutrality score.

## Type of change

- [ ] Groovy reference impl

## Test plan

- [x] `validate_tools()`, `validate_adapter_authoring()`, and `validate_capability_taxonomy_coverage()` return `[]` (0 errors)
- [x] `vendor-neutrality-score` doc block in `docs/vendor-neutrality.md` verified in sync (`Docs in sync: True`)
- [x] `git diff --check` passes with zero whitespace/formatting defects
- [x] Live OSV.dev REST API endpoints tested and verified end-to-end:
  - `GET /v1/vulns/GHSA-7rjr-3q55-vv33` → resolves alias `CVE-2021-45046`
  - `POST /v1/query` with `{"package": {"name": "jinja2", "ecosystem": "PyPI"}, "version": "2.11.2"}` → returns 10 advisory records
  - `POST /v1/query` with commit SHA → returns 0 vulns (valid empty result)
  - `POST /v1/querybatch` with 2 packages → returns batch results (10 + 63 vulns)

## RFC-AI-0004 compliance

- [x] **HITL** — read-only query operations; skills propose matches to human triagers for confirmation
- [x] **Sandbox** — network reach explicitly declared (`api.osv.dev` and `osv.dev`); no credentials or privileged access required
- [x] **Vendor neutrality** — standard placeholders used throughout; backend is swappable via `<project-config>/project.md` (`security_cross_ref.tool: osv`)
- [x] **Conversational + correctable** — per-project ecosystem and tool selection documented for adopter configuration
- [x] **Write-access discipline** — strictly read-only adapter; no state mutations or external reporting
- [x] **Privacy LLM** — explicit confidentiality and embargo boundaries documented per `PRINCIPLES.md §3` (private reproducer commits, draft CVE copy, and embargoed tracker numbers are forbidden from being sent to OSV.dev)

## Linked issues

Closes #311

- Follows the read-only documentation-adapter pattern established by `tools/cve-org/`.
- Complements `tools/cve-org/` (authoritative MITRE publication states) and `tools/cve-tool-vulnogram/` (ASF CNA record editing) by providing ecosystem advisory packages and alias resolution graphs during `security-issue-triage`, `security-issue-deduplicate`, `security-cve-allocate`, and `dependency-audit`.


tools/osv/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+- **Runtime:** None of its own — this directory documents a read-only adapter; queries are `curl` + `jq` recipes (see `tool.md`).
+- **CLIs:** `curl` and `jq`.
+- **Credentials / auth:** None — OSV.dev is a public vulnerability database with an open, unauthenticated REST API.
+- **Network:** `api.osv.dev` (OSV.dev REST API v1) and `osv.dev` (public web UI).
+
+## Configuration
+
+Adopters select this backend with `<project-config>/project.md` →
+`security_cross_ref.tool: osv` when performing automated vulnerability
```
potiuk: `security_cross_ref.tool` and `security_cross_ref.ecosystem` are not defined in `projects/_template/project.md`, so an adopter following this section has nowhere to put them and no template documenting the allowed values. Per [AGENTS.md § Placeholder convention](https://github.com/apache/magpie/blob/main/AGENTS.md#placeholder-convention-used-in-skill-files), a value a tool needs is threaded through the project manifest. Please add the keys (with the ecosystem vocabulary) to the template — or drop this section until a skill actually reads them.

tools/osv/tool.md
```diff
@@ -0,0 +1,178 @@
...
+- **`tools/cve-org/`**: Use `cve-org` for verifying authoritative CVE publication status against MITRE/CVE Services. OSV aggregates CVE data, but `cve-org` remains the primary source of truth for CVE Services container states (`RESERVED`, `PUBLISHED`, `REJECTED`).
+- **`tools/cve-tool-vulnogram/`**: Use `cve-tool-vulnogram` for allocating CVE IDs and editing draft records within the ASF CNA infrastructure.
+- **NVD Direct**: If CVSS severity enrichment from NIST NVD is required directly without ecosystem mapping, an NVD direct adapter can be used.
+
+## Per-project configuration
+
+Adopters configure cross-referencing in `<project-config>/project.md`:
+
+```yaml
+security_cross_ref:
```
potiuk, line 165: Same as the README: this key does not exist in `projects/_template/project.md`. Once it does, link the template section from here so the two stay in step.

tools/osv/tool.md
```diff
@@ -0,0 +1,178 @@
...
+   (mapping CVE ↔ GHSA ↔ OSV IDs).
+2. **Package Version Query (`query-package`)**: Given a package name and ecosystem
+   (e.g., `PyPI`, `Maven`, `npm`, `Go`, `crates.io`, `NuGet`, `RubyGems`, `Packagist`),
+   retrieve all published advisories affecting that specific version.
+3. **Commit Query (`query-commit`)**: Given a public upstream Git commit hash, find any
+   published vulnerabilities associated with or resolved by that commit.
+4. **Batch Query (`query-batch`)**: Query multiple package versions or commit hashes in a
+   single HTTP request.
+
+### Consuming skills
```
potiuk: None of these four skills references `tools/osv` yet (no hit for `osv` under `skills/`), and this PR changes no skill and adds no eval fixture. As written, an adopter would expect triage and dedup to consult OSV and they will not. Please phrase this as intended consumers that are not yet wired, or wire one of them with its eval suite in a follow-up ([AGENTS.md § Reusable skills](https://github.com/apache/magpie/blob/main/AGENTS.md#reusable-skills): a skill change ships with its evals).

docs/adapters/registry.md
```diff
@@ -56,7 +56,7 @@ extension point = a documented, labelled slot with a tracking issue.
...
 | Agent harness | Claude Code, [Codex](codex.md) `experimental` ([#313](https://github.com/apache/magpie/issues/313)), [Gemini CLI](gemini.md) `experimental` ([#314](https://github.com/apache/magpie/issues/314)), [Local LLM (Ollama / llama.cpp / vLLM)](local-llm.md) ([#315](https://github.com/apache/magpie/issues/315)), [Cursor](cursor.md) ([#316](https://github.com/apache/magpie/issues/316)), [Goose](goose.md) `guide only` ([#319](https://github.com/apache/magpie/issues/319)), [Aider](aider.md) `guide only` ([#317](https://github.com/apache/magpie/issues/317)), [GitHub Copilot](copilot.md) `guide only` ([#318](https://github.com/apache/magpie/issues/318)) | Amazon Q [#320](https://github.com/apache/magpie/issues/320)–OpenHands [#322](https://github.com/apache/magpie/issues/322) |
-| Security cross-ref | — | OSV.dev [#311](https://github.com/apache/magpie/issues/311) |
+| Security cross-ref | [`tools/osv`](../../tools/osv/) | NVD direct, GHSA direct |
```
potiuk: The registry's own definition (line 56) is "extension point = a documented, labelled slot with a tracking issue", and every other row links one. "NVD direct, GHSA direct" have none. Please open the two tracking issues and link them here, or leave the column as `—` until someone wants to pick them up.

potiuk (changes requested):
Thanks — the adapter doc is clear and the confidentiality boundaries are exactly what this tool needs. Three things to fix before this lands, two nits, and one note.

**Blocking**

- `security_cross_ref.tool` / `.ecosystem` are documented as `<project-config>/project.md` keys but `projects/_template/project.md` does not define them ([AGENTS.md § Placeholder convention](https://github.com/apache/magpie/blob/main/AGENTS.md#placeholder-convention-used-in-skill-files): values are threaded through the project manifest). Please add them to the template, or drop the configuration section until a skill reads them.
- `tool.md § Consuming skills` names four skills; none of them references `tools/osv` today, and the PR touches no skill or eval. Please reword as "intended consumers, not yet wired" — or wire one skill together with its eval suite in a follow-up, per [AGENTS.md § Reusable skills](https://github.com/apache/magpie/blob/main/AGENTS.md#reusable-skills).
- `docs/adapters/registry.md` lists "NVD direct, GHSA direct" as open extension points without tracking issues; the registry defines an extension point as "a documented, labelled slot with a tracking issue". Open the issues, or leave the column as `—`.

**Nits**

- The PR description ticks "Python package (`tools/*/` with `pyproject.toml`)"; this is a docs-only adapter.
- Every other `tools/*/README.md` starts with the SPDX comment before the doctoc block.

**Note on the score:** the vendor-neutrality block regenerates correctly — I re-ran `vendor-neutrality-score` against this branch and it is in sync — and the drop to 10/11 is the right reading of a one-vendor contract, not something to paper over.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Magpie maintainer. After you've addressed the points above and pushed an update, an Apache Magpie maintainer — a real person — will take the next look at the PR. The findings cite the project's review criteria; if you think one of them is mis-applied, please reply on the PR and a maintainer will weigh in.*
>
> *More on how Apache Magpie handles maintainer review:* [CONTRIBUTING.md § Opening a pull request](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md#opening-a-pull-request).

Kaap10:
Thanks for the review, @potiuk! I've addressed all the feedback in the latest commit (added template config & vocabulary, reframed consuming skills as intended/not yet wired, cleaned registry extension points, fixed SPDX header, and updated the PR checklist).

onlyarnav (commented):
Reviewed the current changes. The adapter registration, project-template configuration, and vendor-neutrality updates look consistent, and the earlier scope concerns appear to be addressed.

One small documentation point: the commit query uses OSV’s `commit` field, which establishes an association with a published vulnerability but doesn’t necessarily mean the commit is the fixing commit. I’d suggest wording this as “vulnerabilities associated with a public upstream commit hash” for precision.

Other than that, I don’t see any blocking issues in the current scope.

tools/osv/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+# `tools/osv/`
+
+**Capability:** contract:security-cross-ref
+
+**Kind:** implementation
+
+**Vendor:** OSV.dev
+
+OSV.dev vulnerability cross-reference client. Queries the [OSV.dev REST API v1](https://api.osv.dev/v1/) for vulnerability records, aliases (CVE ↔ GHSA ↔ OSV IDs), affected package versions, and commit fix references. Complements [`tools/cve-org/`](../cve-org/) and [`tools/cve-tool-vulnogram/`](../cve-tool-vulnogram/) by providing cross-ecosystem vulnerability records for intended triage consumers (`security-issue-triage`, `security-issue-deduplicate`, `security-cve-allocate`, and `dependency-audit` — not yet wired). See [`tool.md`](tool.md) for endpoint recipes, payload structures, and confidentiality boundaries.
```
potiuk: Semantic line breaks, please — this line is 627 characters and holds four sentences.

[`AGENTS.md` § Writing and editing documentation](https://github.com/apache/magpie/blob/main/AGENTS.md#writing-and-editing-documentation) asks for one sentence per line, so that a later single-sentence edit produces a single-line diff rather than a paragraph re-wrap. For calibration, the longest line in any existing `tools/*/README.md` is 318 characters.

The same applies more mildly further down (the two-sentence lines in the `## Configuration` block, and a few in `tool.md`).

tools/osv/README.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Prerequisites
+
+- **Runtime:** None of its own — this directory documents a read-only adapter; queries are `curl` + `jq` recipes (see `tool.md`).
+- **CLIs:** `curl` and `jq`.
```
potiuk, line 30: These recipes cannot run under the setup the framework itself recommends: `Bash(curl *)` is denied in [`docs/setup/secure-agent-setup.md`](https://github.com/apache/magpie/blob/main/docs/setup/secure-agent-setup.md) — *"network egress via Bash bypasses the sandbox proxy"* — and that deny is part of the baseline [`tools/sandbox-lint`](https://github.com/apache/magpie/tree/main/tools/sandbox-lint) enforces. An adopter following this adapter hits a refused command on the first call.

Rather than documenting the exception, please route the four operations through a closed catalogue the way [`tools/vetted-ops`](https://github.com/apache/magpie/tree/main/tools/vetted-ops) does for forge calls. The fit is good: all four are read-only, unauthenticated, and take a closed parameter set (an advisory ID; a package name plus ecosystem plus version; a commit hash; a batch of those). That is exactly the property vetted-ops exists to exploit — *"a closed catalogue whose parameters can never become commands or flags"* — so one allow entry replaces a wildcard, and no `curl` is involved.

One honest wrinkle: `vetted-ops` today shells out to `gh` and nothing else, so this means either teaching the dispatcher a non-`gh` HTTP read backend, or shipping a small stdlib-only helper under `tools/osv/` in the same shape. Either is fine; the first is the better home if you want the single-allow-entry property.

If you take this on, please fold [`tools/cve-org/`](https://github.com/apache/magpie/tree/main/tools/cve-org) into the same change — its `tool.md` has the same `curl` recipes and the same problem. Doing both together is a much better use of the effort than either alone, and it closes the gap for the whole security-cross-reference surface. A follow-up issue is fine if you would rather not grow this PR; say which you prefer and we will label it accordingly.

docs/vendor-neutrality.md
```diff
@@ -168,11 +168,10 @@ with pluggable backends already include:
 The security-team surface follows the same pattern: CNA backends live
 behind [`tools/cve-tool`](../tools/cve-tool/) (the ASF Vulnogram adapter
 [`tools/cve-tool-vulnogram`](../tools/cve-tool-vulnogram/) is the landed
-reference), inbound report relays behind
-[`tools/forwarder-relay`](../tools/forwarder-relay/), scanner formats
+reference), inbound report relays behind [`tools/forwarder-relay`](../tools/forwarder-relay/), scanner formats
```
potiuk: Unrelated reflow — this line was re-joined although only the sentence below it changed, which adds churn to the diff for no benefit.

Please restore the original wrapping and keep the edit to the OSV sentence itself. Same [semantic-line-breaks](https://github.com/apache/magpie/blob/main/AGENTS.md#writing-and-editing-documentation) rationale as the README comment.

docs/labels-and-capabilities.md
```diff
@@ -321,6 +322,7 @@ or a contract-free mix of substrates (e.g. `tools/spec-inventory` is
...
 | [`tools/mail-source`](../tools/mail-source/) | `contract:mail-source` | Mail-source backend abstraction (mbox / IMAP / Mailman 3) feeding a uniform inbound thread/message view to the intake pipeline |
 | [`tools/maildir`](../tools/maildir/) | `contract:mail-source` + `contract:mail-create` | Local **Maildir** backend (Vendor: Maildir) — the offline, credential-free counterpart of `tools/gmail`. Implements `mail-create` by filing editable outbound drafts into a local Maildir for any mail client to send (never sends itself); its `mail-source` side is the local mbox/Maildir archive reader (`tools/mail-source/mbox`). The second, non-Google `mail-create` backend — closes the outbound-mail vendor-neutrality gap |
+| [`tools/osv`](../tools/osv/) | `contract:security-cross-ref` | OSV.dev vulnerability cross-reference client (Vendor: OSV.dev): queries vulnerability records, alias cross-references (CVE ↔ GHSA ↔ OSV IDs), affected package versions, and commit fix references via OSV.dev REST API v1 |
```
potiuk: `commit fix references` overstates what OSV's `commit` field gives you: it associates a commit with a published vulnerability, and that association is not necessarily the fixing commit. Credit to `onlyarnav`, who raised this on the PR at 21:24 UTC — after your last commit, so it is still open.

`tool.md` already words operation 3 correctly (*"associated with or resolved by"*); this row and the lead paragraph of `tools/osv/README.md` should match it.

```suggestion
| [`tools/osv`](../tools/osv/) | `contract:security-cross-ref` | OSV.dev vulnerability cross-reference client (Vendor: OSV.dev): queries vulnerability records, alias cross-references (CVE ↔ GHSA ↔ OSV IDs), affected package versions, and vulnerabilities associated with a public upstream commit via OSV.dev REST API v1 |
```

potiuk (changes requested):
Rebased onto current `main` and re-reviewed. One change requested — the `curl` recipes cannot run under the framework's own recommended permission baseline — plus three nits inline.

**All three blocking items from the previous round are genuinely fixed.** I checked each against the tree rather than the description:

- `security_cross_ref.tool` / `.ecosystem` are now declared in [`projects/_template/project.md`](https://github.com/apache/magpie/blob/main/projects/_template/project.md), with the Tools-enabled row and the TOC entry to match.
- `tool.md § Consuming skills` is reframed as an intended roadmap that is not yet wired, and all four skill links resolve — including `skills/dependency-audit/`, which I confirmed exists.
- The registry row no longer advertises untracked extension points.

Both nits are fixed too: the SPDX header now precedes the doctoc block, and the PR checklist no longer claims a Python package.

**Verified rather than assumed.** I ran the documented example against the live API: `GHSA-7rjr-3q55-vv33` returns exactly the aliases and summary the doc prints, and looking the record up by its CVE alias also returns `200`, so the *"(or alias)"* claim in `get-vuln` holds. The vendor-neutrality block regenerates correctly at 10/11, and a full `prek` run against the rebased branch is green, lychee included.

**Two notes that are not requests.**

- The new contract ships an implementation with no `tools/security-cross-ref/` spec directory. That is not a rule violation — six of the eleven contracts have no directory either — so it is a maintainer call whether the NVD extension point wants something to implement against before a second backend arrives.
- Related: `tool.md § When to replace this tool with another` still offers *"NVD Direct"* as an option, while the registry now tracks no extension point for it. Worth reconciling whichever way you prefer.

The branch has been rebased onto current `main` and force-pushed, so CI is re-running against the post-merge tree.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Magpie maintainer. After you've addressed the points above and pushed an update, an Apache Magpie maintainer — a real person — will take the next look at the PR. The findings cite the project's review criteria; if you think one of them is mis-applied, please reply on the PR and a maintainer will weigh in.*
>
> *More on how Apache Magpie handles maintainer review:* [CONTRIBUTING.md § Opening a pull request](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md#opening-a-pull-request).

Kaap10:
Thanks for the detailed review! All inline comments have been addressed in commit 4b48520:

1. **Semantic line breaks**: Broken down paragraphs and configuration blocks in `tools/osv/README.md` and `tools/osv/tool.md` into 1 sentence per line per AGENTS.md.
2. **Commit query precision**: Updated descriptions in `docs/labels-and-capabilities.md` and `tools/osv/README.md` to *"vulnerabilities associated with a public upstream commit"* per @onlyarnav's suggestion.
3. **Diff hygiene**: Restored original wrapping on line 171 of `docs/vendor-neutrality.md` to keep the diff minimal.
4. **Reconciled NVD Direct**: Removed the NVD Direct bullet from `tool.md` to align with the registry entry (`—`).

Regarding the `curl` sandbox recipes and `vetted-ops` dispatcher: let's track that in a follow-up issue across both `tools/osv/` and `tools/cve-org/` so this PR stays strictly focused.

tools/osv/README.md
```diff
@@ -0,0 +1,38 @@
...
+Queries the [OSV.dev REST API v1](https://api.osv.dev/v1/) for vulnerability records, aliases (CVE ↔ GHSA ↔ OSV IDs), affected package versions, and vulnerabilities associated with a public upstream commit.
+Complements [`tools/cve-org/`](../cve-org/) and [`tools/cve-tool-vulnogram/`](../cve-tool-vulnogram/) by providing cross-ecosystem vulnerability records for intended triage consumers (`security-issue-triage`, `security-issue-deduplicate`, `security-cve-allocate`, and `dependency-audit` — not yet wired).
+See [`tool.md`](tool.md) for endpoint recipes, payload structures, and confidentiality boundaries.
+
+## Prerequisites
+
+- **Runtime:** None of its own — this directory documents a read-only adapter; queries are `curl` + `jq` recipes (see `tool.md`).
+- **CLIs:** `curl` and `jq`.
+- **Credentials / auth:** None — OSV.dev is a public vulnerability database with an open, unauthenticated REST API.
+- **Network:** `api.osv.dev` (OSV.dev REST API v1) and `osv.dev` (public web UI).
```
potiuk, line 32: `tools/osv` egresses to `api.osv.dev`, but that host is not on the egress gateway's allowlist — so every recipe in `tool.md` gets a `403` before a socket opens, for anyone running the framework's recommended setup.

[`allowlist.py`](https://github.com/apache/magpie/blob/main/tools/egress-gateway/src/egress_gateway/allowlist.py) → `ALLOW_EXACT` carries every other security-data host (`cveawg.mitre.org`, `nvd.nist.gov`, `cve.org`), and the gateway is default-deny in `before_upstream_connection`. The container gateway also hands it to every container as its HTTP proxy, so container traffic is bound by the same list.

Fixed in a fixup on this branch: the host added to `ALLOW_EXACT`, an `osv` row added to *Declared egress surfaces*, and a case added to `test_allowlist.py::test_allowed_hosts` beside `cveawg.mitre.org`. The `osv.dev` web UI is deliberately left out — browser destination, not a tool egress.

tools/osv/tool.md
```diff
@@ -0,0 +1,168 @@
...
+```bash
+curl -sSf -X POST https://api.osv.dev/v1/querybatch \
+  -H "Content-Type: application/json" \
+  -d '{
+    "queries": [
+      {"package": {"name": "jinja2", "ecosystem": "PyPI"}, "version": "2.11.2"},
+      {"package": {"name": "urllib3", "ecosystem": "PyPI"}, "version": "1.26.4"}
+    ]
+  }' \
+  | jq -r '.results | to_entries[] | {query: .key, vuln_count: ((.value.vulns // []) | length)}'
```
potiuk, line 134: `.results[]` discarded which query each result belongs to. OSV returns `querybatch` results in request order, so the index is the only thing tying a result back to its input — as written you got a list of counts with no way to know which package each one was for.

Fixed in a fixup (this line): `to_entries` carries the position through, and `// []` makes a result object with no `vulns` key count as zero rather than erroring.

potiuk (approved):
Approving. Everything from the previous two rounds is fixed, and I've taken the two remaining items myself as fixups on this branch rather than send you a fourth round for six lines.

**Verified against the tree, not the description:**

- Semantic line breaks — longest line in `tools/osv/README.md` is now 306 characters, down from 627, one sentence per line.
- The unrelated reflow in `docs/vendor-neutrality.md` is restored to a two-line change.
- `commit fix references` is reworded to *"vulnerabilities associated with a public upstream commit"* in both `docs/labels-and-capabilities.md` and the README lead — which also closes the precision point `onlyarnav` raised.
- *"NVD Direct"* is gone from *When to replace this tool*, reconciling with the registry.

**I withdraw my own blocking finding from round two.** I said the `curl` recipes couldn't run under the framework's recommended permission baseline. The `Bash(curl *)` deny is real, but `tools/cve-org/tool.md` — the precedent this adapter explicitly follows — ships the same bare `curl … | jq` recipes with no caveat, and no documented rule requires a tool doc to route HTTP elsewhere. That was me holding this PR to a bar the repository doesn't apply. The deny-versus-adapter-recipes inconsistency is real but repo-wide, and belongs in its own issue covering both adapters. My calibration in that thread was also wrong: I said 318 characters was the longest line in any `tools/*/README.md`; it's 2114.

**Two fixups pushed**, explained inline. I also rebased the branch onto current `main` and force-pushed, so CI re-runs against the post-merge tree — the same thing I did in round two.

Thanks for the patience through three rounds.

---

> *This review was drafted by an AI-assisted tool and confirmed by an Apache Magpie maintainer. The maintainer approving this PR has read the findings and signed off. If something feels off, please reply on the PR and a maintainer will follow up.*
>
> *More on how Apache Magpie handles maintainer review:* [CONTRIBUTING.md § Opening a pull request](https://github.com/apache/magpie/blob/main/CONTRIBUTING.md#opening-a-pull-request).

Kaap10:
Follow-up tracking issue created as discussed: #1320. Thanks again for the review and merge, @potiuk!

PR #1298, opened 2026-09-20
feat(container-gateway): run podman and docker from inside the sandbox through a policy proxy

potiuk:
## What this adds

A **container gateway**: a per-project policy proxy in front of the podman / docker daemon socket, so shell commands run by a sandboxed agent can drive containers without the daemon being reachable.

Today an adopter has two options, and both hand over the daemon. Excluding `docker` from the sandbox (upstream's advice) or allowing the daemon socket in `sandbox.network.allowUnixSockets` both grant what upstream's own documentation calls "access to the host system through the Docker socket". On macOS the default Podman machine mounts `/Users` read-write, so that is a bind mount away from `~/.ssh`.

The gateway runs outside the sandbox, listens on two unix sockets inside the project tree, and forwards the Docker-compatible API after enforcing policy. It gives three properties:

1. **Containers only.** A create body is an allow-list: unknown fields are refused, and the fields that turn a container into host access (privileged, capabilities, devices, host namespaces, unconfined security options, `rootfs`, overlay volumes, volume-driver configurations, bind mounts outside the project) are denied by name. The build query, exec bodies and act-by-name queries are policed the same way.
2. **This project's containers only.** Every container, pod, volume and network is labelled with the project; every list, act and remove is filtered or checked against that label, resolved by ID so a rename cannot race it.
3. **The same egress policy as the shell.** Containers and builds get the framework's egress gateway as their HTTP proxy. This is a friction layer, not a wall, and says so.

Both backends are optional; the gateway serves the docker CLI from a podman backend when Docker is absent, and exits quietly when neither exists.

## What is in the branch

Tool `tools/container-gateway/` (stdlib only): routing, policy, the HTTP relay with streaming and hijack support, backend discovery, the daemon and its CLI. A `SessionStart` / `SessionEnd` hook in `tools/agent-isolation/`. Reference settings plus a sandbox-lint invariant that rejects a daemon socket in the allow-list. Documentation in the setup guide and the troubleshooting catalog, a row in RFC-AI-0004 and a cross-reference in RFC-AI-0003. Doctor probe, verify check and install step, with eval fixtures. 581 unit tests and an integration suite that runs against a real daemon.

## How it was built and reviewed

Spec first, then fifteen planned tasks, each implemented and reviewed independently; the security-bearing parts took several fix rounds each. A whole-branch review then found four bypasses that per-component review could not see, all of the same shape: the policy enumerated what was forbidden rather than what was allowed. That is why the create body is now an allow-list.

An integration run against a live daemon found a fifth defect that no test against fakes had caught, and the fix round found a sixth.

## Limits

`tool.md` and the README carry a *Limits and residual risks* section. The ones worth repeating: bind sources are checked on the host at decision time and re-resolved by the daemon at mount time, so a symlink swapped in that window is not caught; images are shared across projects by design; container egress that bypasses proxy variables is not intercepted; backend discovery happens at gateway start.

## Follow-ups

Automating the per-project socket entries in `/magpie-setup config` (documented as manual for now), and exercising the libpod volume-label path against podman.

## Generative AI disclosure

Written with Claude Code (Opus 5) under human direction: spec, plan, per-task implementation and review, and the final review passes. Every commit was reviewed by a separate agent and by the author before landing.


potiuk:
The **last** security layer added - container gateway. It allows to safely (and without prompts) run containers inside your agentic CLI. It will keep containers separated "per-project" so that they are not reachable from inside the sandbox.

